Issue #83, opened 2017-06-13
clip.ts decodeURI() call doesn't compile


kdavis-mozilla:
The call
```Javascript
let sentence = decodeURI(info.sentence);
```
does not compile.

mikehenrty:
Fixed by #82

PR #85, opened 2017-06-13
give typescript hints for pulling strings off headers, fixes #82

### LINKED ISSUE #82: ES5 Incompatable Code in clip.ts
The following methods are used but not part of ES5
- `string.startsWith()`
- `string.includes()`

They can easily be replaced with equivalent code

mikehenrty:


kdavis-mozilla:
This fixes the errors from issue #83

PR #90, opened 2017-06-14
Fixed #68 and fixed #89

### LINKED ISSUE #68: [Server] Use s3 to store audio files
Save clip data into s3 buckets.

Note: we will probably want to save temporary on the local filesystem as a buffer for doing audio conversion.

### LINKED ISSUE #89: There is not background process that converts new files to mp3's
The conversion is implemented by this method [convertMissingToMP3s()](https://github.com/mozilla/voice-web/blob/master/server/src/lib/files.ts#L111)

However, this method is only called once here [files.ts:214](https://github.com/mozilla/voice-web/blob/master/server/src/lib/files.ts#L214) in the [init()](https://github.com/mozilla/voice-web/blob/master/server/src/lib/files.ts#L168) method which only gets called when the server starts.

So any files that need to be converted after the server starts are not converted.

kdavis-mozilla:
Should save everything to S3, convert S3 ogg's to mp3's.

Still need to test the "listen" part, but can't as the "listen" functionality is broken in the master.

I don't want a through review, just a quick look over the come to make sure it doesn't suck too much.

FYI: To set up S3 you should create `~/.aws/credentials` and set it up as follows
```bash
kdaviss-MacBook-Pro:src kdavis$ cat ~/.aws/credentials 
[default]
aws_access_key_id = <my aws_access_key_id>
aws_secret_access_key = <my aws_secret_access_key>
```
after that it all works magically

### DIFF
--- a/config.json
+++ b/config.json
@@ -6,6 +6,7 @@
   "MSQLPASS": "voicecommons",
   "MSQLDBNAME": "voiceweb",
   "MSQLHOST": "localhost",
-  "MYSQLPORT":  "3306"
-
+  "MYSQLPORT":  "3306",
+  "BUCKET_NAME": "",
+  "BUCKET_LOCATION": ""
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3466 @@
+{
+  "name": "voice-web",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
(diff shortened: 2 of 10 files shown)


kdavis-mozilla:
@mikehenrty We still need to add AWS credentials to the test.mozvoice machine

PR #104, opened 2017-06-15
Feature/nubis

gozer:


config.json
```diff
@@ -6,7 +6,7 @@
   "MSQLPASS": "voicecommons",
   "MSQLDBNAME": "voiceweb",
   "MSQLHOST": "localhost",
-  "MYSQLPORT":  "3306",
+  "MYSQLPORT":  "3306"
```
mikehenrty: needs a comma here

package.json
```diff
@@ -27,7 +27,7 @@
     "walk": "^2.3.9"
   },
   "devDependencies": {
-    "@types/node": "^7.0.29",
+    "@types/node": "7.0.29",
```
mikehenrty: nice. we'll have to figure out what changed in 7.0.30+ later, but this is a good workaoround

config.json
gozer: Wow, good catch! Fixed!

package.json
gozer: Yeah, and Iv'e done the binary search, 7.0.30 is where things break

PR #109, opened 2017-06-15
Use outbound proxy for HTTP calls when present

### LINKED ISSUE #108: Support outbound HTTP proxies
Nubis accounts require all traffic to transit via our outbound http proxies, convenienelty found via the HTTP_PROXY environment variables.

gozer:
Fixes #108

### DIFF
--- a/server/src/lib/files.ts
+++ b/server/src/lib/files.ts
@@ -1,6 +1,5 @@
 import { map } from '../promisify';
 import { getFileExt } from './utility';
-import S3 = require('aws-sdk/clients/s3');
 
 const MemoryStream = require('memorystream');
 const path = require('path');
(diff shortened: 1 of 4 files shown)


gozer:
Apologies, #107 should have been on its own branch, but was on master, so this PR will show extraneous commits until #107 gets merged.

server/src/lib/files.ts
```diff
@@ -15,9 +14,19 @@ const CONFIG_PATH = path.resolve(__dirname, '../../..', 'config.json');
...
+var AWS = require('aws-sdk');
+
+if(process.env.HTTP_PROXY) {
+  var proxy = require('proxy-agent');
+
+  AWS.config.update({
+    httpOptions: { agent: proxy(process.env.HTTP_PROXY) }
+  });
+}
+
```
mikehenrty: Since we do essentially the same thing twice here, any chance, I could get you to create a new module. Maybe something like:
`server/src/lib/aws.ts`
```
var AWS = require('aws-sdk');

if(process.env.HTTP_PROXY) {
  var proxy = require('proxy-agent');

  AWS.config.update({
    httpOptions: { agent: proxy(process.env.HTTP_PROXY) }
  });
}

module.exports = AWS
```
Then, both here and in clip.ts, you could just:
`const AWS = require('./aws');`

gozer: I was trying to do precisely that and failed at it! Sure thing!

gozer:
Made the recommended good refactoring into a single code instance

mikehenrty:
👍

PR #129, opened 2017-06-17
more descriptive error message when ios app not connected to the inte…

mikehenrty:
…rnet


mikehenrty:
@andrenatal could you take a look at this PR? just a string change, but i want to make sure i'm not breaking anything

PR #252, opened 2017-06-26
preload audio

### LINKED ISSUE #251: buttons like play and delete not working on ffx for android
Subject says it all.

fiji-flo:
This resolves #251


SteveALee:
When I refreshed the controls became enabled (black rather than grey text) but audio was utterly distorted when I played it.

I submitted any way and got back to the sample page but with blank text to say.

PR #253, opened 2017-06-27
add in a bunch of sentences from the public domain book the alchemist

mikehenrty:


mikehenrty:
@kdavis-mozilla what do you think about us adding a bunch of sentences from the public domain book The Alchemist? I was going to add more sentences from War of the Worlds, but they were so dark and depressing and I think this lightens the mood a bit. No need to read every sentence (I did a filtering pass), I just wanted to run this by you before deploying it.

kdavis-mozilla:
@mikehenrty If you did a filtering pass, I'm fine with it

jimregan:
This book is not in the public domain: https://en.wikipedia.org/wiki/The_Alchemist_(novel)

mikehenrty:
True, but it was written by Paulo Coelho, and he general doesn't mind when you use his book:
https://www.techdirt.com/articles/20110429/01144214078/paulo-coelho-explains-why-he-loves-pirates.shtml

In any case, users are no longer reading this material.

PR #257, opened 2017-06-28
Fixed #255

### LINKED ISSUE #255: Better randomize validation sentences

kdavis-mozilla:
@Could you take a look at this?

I found a way to "continue" the S3 getObject's call so that it walks through all the elements in S3 50 at a time and caches these elements in on the server. I make the update more often as it's only getting 50 elements each time.

Maybe in future we need to think about limiting the number of elements in that are cached on the server. But for now, this will get the the job done with respect to better randomization.


mikehenrty:
sorry, this one had completely slipped my radar until someone reminded me about this bug today

PR #261, opened 2017-07-03
Several small improvements to the Android app

Omniscimus:
The Android app contains just a WebView in which the website is displayed. I made the following changes:

- Clicking any link caused a new page to load in the WebView. Users could navigate away from the mozvoice website, for example to mozilla.org. Clicks on external links (with a different host name) now open in the default browser.
- Changed the app's icon to the same image as the iOS version, instead of the default Android logo. (If the app is actually going to be used, then the icon should probably be updated)
- The app automatically opened the voice input page. Changed this to the home page so the user gets some information about the project first.
- The app asked the audio permission when it was opened for the first time. If the user clicked 'Deny', then it wasn't possible to undo that choice. Fixed this by asking each time the web page requests the permission (if it hasn't already been granted).


mikehenrty:
Thanks for you contribution @Omniscimus!!

Can you take a look at this, @andrenatal ?

andrenatal:
Let's merge this!

Issue #259, opened 2017-06-30
Blank sentences on the "Speak" page after the privacy policy is declined


paulsilaghi:
- Chrome 59, Firefox 54
- Win 10, iOS 10.3.3/iphone 6S

[Steps to reproduce]:
1. Open https://voice.allizom.org/
2. Go to the "Speak" page
3. Complete 3 recordings
4. Press the 'Submit' button
5. Decline the privacy policy
6. Delete the recordings
7. Go to the "Speak" page and try to complete 3 recordings again

[Expected result]:
- all 3 sentences are displayed correctly

[Actual result]:
- 1st sentence displayed correctly, 2nd and 3rd are blank

![blank sentences](https://user-images.githubusercontent.com/29535384/27734321-0105eafc-5da3-11e7-9f7f-6a8e07f15c85.png)

ayushbansal07:
Hi, I want to work on this issue. But, I am not able to run the webapp on my local machine. 
I cloned your repo. Ran these 2 commands successfully - 
npm install
npm install -g gulp
But when I run gulp, I get this error:

const APP_NAME = 'common-voice';
^^^^^
SyntaxError: Use of const in strict mode.
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Liftoff.handleArguments (/usr/local/lib/node_modules/gulp/bin/gulp.js:116:3)
    at Liftoff.<anonymous> (/usr/local/lib/node_modules/gulp/node_modules/liftoff/index.js:198:16)
    at module.exports (/usr/local/lib/node_modules/gulp/node_modules/liftoff/node_modules/flagged-respawn/index.js:17:3)

To resolve this, I removed the line 'use strict' from gulpfile.js, but then I got the following error:

let gulp = require('gulp-help')(require('gulp'));
    ^^^^
SyntaxError: Unexpected identifier
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
[...]

PR #283, opened 2017-07-18
Fixed the bug for Issue #259.

ayushbansal07:
In the reset() function, previously we fetched newSentences first and reset the recordings afetr that.
Now I just interchanged the order.

### DIFF
--- a/web/src/lib/components/pages/record.tsx
+++ b/web/src/lib/components/pages/record.tsx
@@ -220,12 +220,8 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
   }
 
   newSentenceSet() {
-    let recordedSentenceCount = this.state.recordings.length;
-    let numberOfSentenceToGet = SET_COUNT - recordedSentenceCount;
-    this.props.api.getRandomSentences(numberOfSentenceToGet).then(newSentences => {
-      let targetSentences = this.state.sentences.slice(0,recordedSentenceCount);
-      targetSentences = targetSentences.concat(newSentences.split('\n'));
-      this.setState({ sentences: targetSentences});
+    this.props.api.getRandomSentences(SET_COUNT).then(newSentences => {
+      this.setState({ sentences: newSentences.split('\n')});
     });
   }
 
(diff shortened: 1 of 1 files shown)


mbebenita:
I'm not sure how things work in Preact, but in React setting the state is an asynchronous operation and there's no guarantee that the new state will be available right after the call to `setState`.

ayushbansal07:
I too am not sure whether setState would be synchronous or asynchronous.  Though, when I tested the app on my local machine it worked fine. 
The main problem here was that when we fetched the new sentences by calling the function newSentenceSet(), the value of recordedSentenceCount remains 3 since the recordings do not get reset. So, the necessary part was just to reset recordings.

mikehenrty:
Yeah, @mbebenita is right, setState is asynchronous function, so even if this fixes the problem locally, the solution would be non-deterministic. Sounds like the problem is be around recordedSentenceCount not getting reset with the sentences? Is there a way we can make those align?

ayushbansal07:
Would initialising state.recordings to empty array before calling the newSentenceSet() function work?

mikehenrty:
> Would initialising state.recordings to empty array before calling the newSentenceSet() function work?

Unfortunately, you are not allowed to set the object state directly, you must use setState to take advantage of all the react rendering magic.

That said, I think you are on the right track with your change here. It makes sense to setState first (even if it is asynchronous). Part 2 is, we need to simplify the `newSentenceSet` function. The logic we have there now was from a fix (https://github.com/mozilla/voice-web/commit/0a5c9dab7788f09ef401c1cd72c3fe560cb73746) to an outdated bug. If, in addition to the changes you have made here, you also make the `newSentenceSet` function always fetch 3 sentences, I think this will fix this for good. Perhaps something like this:
``` 
newSentenceSet() {
    this.api.getRandomSentences(SET_COUNT).then(sentences => {
      this.setState({ sentences: sentences.split('\n') });
    });
}
```

Can you give that a try @ayushbansal07 ?

ayushbansal07:
@mikehenrty , I did and tested the change so that the function newSentenceSet always fetches 3 ( = SET_COUNT) sentences. Everything was working fine locally.

mikehenrty:
Great! I squashed your commit, and merge this here: https://github.com/mozilla/voice-web/commit/c3ac34db783194bffa1980f0444d2666286d19b1

PR #304, opened 2017-07-20
Fix issue #23 (Larger corpus of text)

cherls:
Stanford Natural Language Inference Corpus ([SNLI](https://nlp.stanford.edu/projects/snli/)) is a collection of 570k human-written English sentence pairs licensed under a Creative Commons Attribution-ShareAlike 4.0 International License (Which suffices for commercial usage and modification, though I'm not sure under what license Mozilla is planning to release the final dataset).

I've extracted ~104k unique sentences from the collection after filtering by character length, Flesch–Kincaid readability tests for sentence complexity, and preprocessed for white-spaces, capitalization, and missing periods.


mikehenrty:
This is amazing @cherls! Very timely and helpful. We will release our voice data under CCO, so we are checking with our legal team now how to incorporate this.

Great find!

mikehenrty:
Here's an update @cherls. We had our legal team look at the licensing concerns, and it looks like if we were to use this text corpus, we would have to also license the derived voice data under ShareAlike. This is not a deal breaker, but our current terms of service says all the donated voice data will be licensed under CC0. So to accept this PR, we would have to create a new bucket of data for these SNLI sentences, update our Terms, and message this to the users (both when contributing and downloading). This is not impossible, but will take us some time to figure out the logistics.

Thank you again for the PR, rest assured we are eager to incorporate it, but it might be a while.

MichaelNMaggs:
I concur with the legal team.  Unfortunately, a dataset licensed under Creative Commons Attribution-ShareAlike is simply not compatible with a release of the end result under CC-0.   Unless anyone can persuade the owners of SNLI to release it to the public domain or CC-0,  the only possibility would be to keep the SNLI data separate from the rest and available only under a share alike licence. As that would be both a practical and a legal nightmare, I fear that the SNLI data can't be used within this project.

kdavis-mozilla:
@mikehenrty Sean lectures at Stanford. Perhaps he has some levers he can pull?

PR #308, opened 2017-07-20
Use Bunyan for JSON logs

gozer:


mikehenrty:
Thanks for the contribution, if you don't mind, I'm going to try something that has a smaller footprint but similar output.

mikehenrty:
Actually let's keep this open until I know for sure what I want to do works.

Issue #315, opened 2017-07-21
Apache app deployment


Omniscimus:
To test a fix for an issue, I tried to set up the app on my web server. By following the instructions in the README, I got the website working on localhost:9000. However, I would like it on my Apache server to change some Apache settings, as [this file](https://github.com/mozilla/voice-web/blob/master/nubis/puppet/web.pp) seems to suggest. It looks like it should automatically redirect requests at the Apache server to port 9000, but that's not working for me. Does it need additional configuration to get the Nubis part working?

Files in nubis/ seem to suggest that the Apache server would use the /var/www/voice directory. I can configure Apache to use this, but would I need to place additional files there? Also, is a Ubuntu server a requirement? (since it chowns something as ubuntu:ubuntu)

I was trying to work on #103. If there's no easy fix due to server infrastructures / deployment / whatever then I'll just move on to another issue.

gozer:
The nubis/ contents is specific to our internal Nubis deployment tool.

If all you are trying to do is add a header for CSP, have a look at:

https://github.com/mozilla/voice-web/blob/master/nubis/puppet/web.pp#L40

Where we are already setting up a couple.

Omniscimus:
Yes I found that, but I couldn't test my fix. The CSP header should be as strict as possible but if it's too strict, things may break. I'll submit a PR anyway. Thanks!

PR #318, opened 2017-07-21
Added a Content Security Policy header

### LINKED ISSUE #103: Add CSP header
This will increase our security rating our Nubis.

Omniscimus:
Fixes #103. The CSP header may be important for security (XSS prevention), especially when more user-generated content will be displayed to other users in the future, such as accent information. The browser will ignore content loaded from external websites, as well as inline scripts and CSS.

Couldn't test this properly due to #315. I did test the header by adding it in a meta tag to web/index.html. If the `media-src` directive is omitted, the browser refuses to load recorded audio, which appears to have a URI similar to `blob:https://voice.mozilla.org/ad3c9602-f211-4a28-88f2-f98f61bbe72f`.

Before being merged, this should be tested on a website where the Nubis stuff works. If the header contains mistakes, legitimate content may not load. I'm curious about downloading validation audio and recording audio, which I could not test properly either because I don't have an Amazon Cloud thing. Chrome logs CSP violations in its debug console, which may be useful.

**NOTE**: any inline CSS will be ignored by the browser with this CSP header. This appeared to be no problem, because it's all loaded from external stylesheets.

### DIFF
--- a/nubis/puppet/web.pp
+++ b/nubis/puppet/web.pp
@@ -42,6 +42,8 @@ apache::vhost { $project_name:
       'set X-XSS-Protection "1; mode=block"',
       'set X-Frame-Options "DENY"',
       'set Strict-Transport-Security "max-age=31536000"',
+      # The media-src directive is required for recording audio.
+      'set Content-Security-Policy "default-src \'self\'; object-src \'none\'; media-src blob: \'self\'"',
     ],
     rewrites           => [
       {
--- a/nubis/puppet/web.pp
+++ b/nubis/puppet/web.pp
@@ -42,8 +42,8 @@ apache::vhost { $project_name:
       'set X-XSS-Protection "1; mode=block"',
       'set X-Frame-Options "DENY"',
       'set Strict-Transport-Security "max-age=31536000"',
-      # The media-src directive is required for recording audio.
-      'set Content-Security-Policy "default-src \'self\'; object-src \'none\'; media-src blob: \'self\'"',
+      # media-src blob: is required for recording audio.
+      'set Content-Security-Policy "default-src \'self\'; img-src \'self\' https://www.google-analytics.com; media-src blob: https://*.amazonaws.com; script-src \'self\' \'unsafe-eval\' https://www.google-analytics.com/analytics.js"'
     ],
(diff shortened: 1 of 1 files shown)


mikehenrty:
Thanks @Omniscimus! @gozer could you take a look at this?

mozfreddyb:
Here's an unsolicited drive-by security review from someone who knows CSP very well, but doesn't know your application and is totally *unaffiliated* with our service/website/infra security people:

This CSP is a security improvement 👍

mikehenrty:
@gozer review ping?

Omniscimus:
I just tested the header with the [Laboratory](https://github.com/april/laboratory) plugin and realized that it wasn't allowing Google Analytics to get through. Also, in the `media-src` directive, the AWS server had to be added because recordings are downloaded from there for validation. I believe that the domain of the AWS server is not always the same (please correct me if I'm wrong), so I couldn't make it any stricter than `https://*.amazonaws.com`.

Finally, with this CSP header, the browser console throws errors if `'unsafe-eval'` is not in the `script-src` directive, although everything seems to be working. `'unsafe-eval'` reduces the security benefits of the CSP header. The Web Console is giving me the following error:

`Content Security Policy: The page’s settings blocked the loading of a resource at self (“script-src https://voice.mozilla.org https://www.google-analytics.com/analytics.js”). Source: call to eval() or related function blocked by CSP.`

According to the Browser Console, this is caused by **two** calls to `Function()`:

`Error: call to Function() blocked by CSP`

`bundle.js` does not contain a call to `Function()`. `analytics.js` does contain the word Function, but not a call to it, and even if I block the analytics script, the error remains. Does anyone have any idea where this could come from? This is the CSP I was testing:

`default-src 'self'; img-src 'self' https://www.google-analytics.com; media-src blob: https://*.amazonaws.com; script-src 'self' https://www.google-analytics.com/analytics.js`

mikehenrty:
@Omniscimus thanks again for your help here.

I've been testing and playing around with some CSP settings, all based on your initial recommendation. I'm going to merge this, make some tweaks, and then push it to our staging server (voice.allizom.org) and test it. Hopefully we'll have this working very soon.

Issue #233, opened 2017-06-23
Cancel should not refresh page on microhpone access dialog


mikehenrty:

augustnmonteiro:
I'm working on this one but I'm having a problem, the dialog only shows once so when someone cancels and the page isn't refreshed we cannot show the dialog again, thinking about UX I don't know if it's better doesn't refresh and when try again show the dialog and press retry

mikehenrty:
@augustnmonteiro Why can't we not show the dialog again? Can we remember that the user denied the microphone, and go straight to the dialog if they click record again? Does that add too much complexity?

augustnmonteiro:
@mikehenrty After the browser shows the Allow/Deny dialog and someone denies it the browser doesn't show the dialog again

mikehenrty:
Ah, the microphone permission dialog, I thought you meant the cancel/retry dialog.

If the user clicks the record button again, can we detect that the user denied permission before, and then show the cancel/retry dialog again?

PR #321, opened 2017-07-22
fixed issue #233

ayushbansal07:


web/src/lib/components/pages/record.tsx
```diff
@@ -187,10 +187,12 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
...
     // Don't start a new recording when full.
     } else if (!this.isFull()) {
-      this.audio.init().then(() => {
-        this.startRecording();
-      });
-    }
+      	this.audio.init().then(() => {
+        	this.startRecording();
+      	}).catch((err) =>{
+      		window.location.href = "/";
```
mikehenrty: This will cause the page to reload, which is not what we want (that's why we filed #233 in the first place). Best to just do nothing when the users cancels.

Also please follow code conventions. Looks like you are using tabs, whereas we have chosen spaces.

mikehenrty:
Closing this in favor of #475

Issue #328, opened 2017-07-23
Typo in funny.txt


DeathCamel57:
Line 1 of [server/data/funny.txt](github.com/mozilla/voice-web/blob/master/server/data/funny.txt) is
> 'm all about those invisible cosmic connections that transcend time and distance.

An "I" should be added at the start of the line, so it reads as follows.
> I'm all about those invisible cosmic connections that transcend time and distance.

DeathCamel57:
Will fork and pull.

DeathCamel57:
Pull request sent in.
#329

mikehenrty:
Fixed in https://github.com/mozilla/voice-web/commit/100d604588eaebd19f4eec9558eab737f467d563

PR #329, opened 2017-07-23
Typo fix for #328

DeathCamel57:
Typo fix for #328


mikehenrty:
Thanks for the fix @DeathCamel57!

We don't actually use this file right now, but if we ever do again, it will be good to have this fixed. Thanks!

Issue #306, opened 2017-07-20
Please include the source of the text in the FAQ


Kusaha:
I was wondering what book (or else) are we using for the sentences, it would be nice if you would include the source text in the FAQ or at least the books title.

Omniscimus:
Duplicate of #287

DeathCamel57:
Here's where the text is located.

- [voice-web/server/data/temporary-sentences.txt](https://github.com/mozilla/voice-web/blob/master/server/data/temporary-sentences.txt)
- [voice-web/server/data/temporary-sentences-2.txt](https://github.com/mozilla/voice-web/blob/master/server/data/temporary-sentences-2.txt)
- [voice-web/server/data/funny.txt](https://github.com/mozilla/voice-web/blob/master/server/data/funny.txt)
- [voice-web/server/data/new-sentences.txt](https://github.com/mozilla/voice-web/blob/master/server/data/new-sentences.txt)

DeathCamel57:
Now to see if the pull request is accepted.

DeathCamel57:
New pull (#330) should fix this.

mikehenrty:
We swapped out those books for contributor donated sentences, and some public domain screenplays like _It's a Wonder Life_. We now document this in our faq, thanks to @DeathCamel57's PR. Closing.

Issue #287, opened 2017-07-18
Document sentence source


lucaswerkmeister:
It would be nice to know where the sentences that are being read come from – not necessarily for any one particular sentence, but at least the collected sources for all sentences together. As far as I can tell, this information isn’t currently available – even in git history ([`server/data` commits](https://github.com/mozilla/voice-web/commits/master/server/data)), only the last one (e969b19b3b266976ee184caf2af3a08afed7f60a) records the source in the commit message (and there seem to be several works titled “The Alchemist”).

mikehenrty:
Good idea @lucaswerkmeister! Right now we really only use War of the Worlds and The Alchemist, but we will add more shortly. Any suggestions on where we should track this?

lucaswerkmeister:
Perhaps in the FAQ, as long as the list is short enough?

angelawyman:
Hunh, Where does the text "Can I get woot woo" come from? Are you taking new book suggestions as currently War of the Worlds and the Alchemist text are heavily male pronoun biased.

DeathCamel57:
I added links for the source text files with all of the phrases to the FAQ page.

angelawyman:
Great!! That clears it up to their origin. Glad to see "You don't know shame until you drunkenly confuse Hellen Keller and Anne Frank in a game of charades." was removed. What year is the 'funny' texts from? There are references to "Mrs Butterworth", "I can't believe its not butter", "Hamburger Helper" and "I have to check with the Mrs!". That's the 70's, right?

lucaswerkmeister:
This doesn’t answer the question, though. It’s nice to see the full list of sentences, but the files don’t document the sentence sources.

mikehenrty:
> What year is the 'funny' texts from? There are references to "Mrs Butterworth", "I can't believe its not butter", "Hamburger Helper" and "I have to check with the Mrs!". That's the 70's, right?

Yeah, I agree these are not good. I'll remove them all today.

mikehenrty:
I removed all the problematic ones I could find in https://github.com/mozilla/voice-web/commit/c5fd185c3ec67bd91090ae0ec9770bee65f009f5. If you find more, please file a new issue for that and I will remove them.

DeathCamel57:
#330 Should cover it now.

mikehenrty:
We now document the source text in our faq thanks to #330.

PR #330, opened 2017-07-23
Add text sources to FAQ

DeathCamel57:
In re: #287 and #306

### DIFF
--- a/web/src/lib/components/pages/faq.tsx
+++ b/web/src/lib/components/pages/faq.tsx
@@ -33,7 +33,8 @@ export default class FAQ extends Component<Props, State> {
 
       <h3>Where can I find the sample text?</h3>
       <p>Common Voice is an open source project that is on <a href="https://github.com/mozilla/voice-web/">Github</a></p>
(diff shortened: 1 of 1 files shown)


web/src/lib/components/pages/faq.tsx
```diff
@@ -30,6 +30,10 @@ export default class FAQ extends Component<Props, State> {
 
       <h3>Why is 10,000 hours the goal for capturing audio?</h3>
       <p>This is approximately the number of hours required to train a production STT system.</p>
+
+      <h3>Where can I find the sample text?</h3>
+      <p>Common Voice is an open source project that is on <a href="https://github.com/mozilla/voice-web/">Github</a></p>
+      <p>Our sets of samples are available <a href="https://github.com/mozilla/voice-web/blob/master/server/data/temporary-sentences.txt">here</a>, <a href="https://github.com/mozilla/voice-web/blob/master/server/data/temporary-sentences-2.txt">here</a>, <a href="https://github.com/mozilla/voice-web/blob/master/server/data/funny.txt">here</a>, and <a href="https://github.com/mozilla/voice-web/blob/master/server/data/new-sentences.txt">here</a>.</p>
```
mikehenrty: Right now, we just use `https://github.com/mozilla/voice-web/blob/master/server/data/new-sentences.txt`.

mikehenrty:
To fix bug #330, you will also have to add where these sentences come from, which is War of the Worlds and The Alchemist.

DeathCamel57:
What about this edit @mikehenrty

web/src/lib/components/pages/faq.tsx
DeathCamel57: Fixed!

mikehenrty:
Unfortunately, I just changed our source material and where it lives in the tree. I'll accept your change and then patch it.

PR #339, opened 2017-07-23
make all console operations output json, closes #308

mikehenrty:
@gozer can you tell me if this output fulfills your logging requirements?


mikehenrty:
Alright, got confirmation from @gozer to try this approach.

Issue #178, opened 2017-06-19
Text is not aligned to check box


mbebenita:
![image](https://user-images.githubusercontent.com/311082/27298967-8f0a074c-54df-11e7-85e2-9d02ec9f097b.png)

The placement of the "I'd like .." text doesn't look right.

kdavis-mozilla:
I seem a similar alignment for OS X + Firefox

kdavis-mozilla:
@mbebenita What OS + Browser are you using?

DeathCamel57:
Same issue on:
**Browser:** Vivaldi | 1.8.770.56 (Stable channel) (64-bit) 
**OS:** Ubuntu 17.04 zesty
**Kernel:** x86_64 Linux 4.10.0-28-generic

![Picture](https://user-images.githubusercontent.com/1226646/28507383-67f388f4-7000-11e7-9159-046586ff7969.png)

DeathCamel57:
#343 hopefully fixes it.
The description line had a fixed `1rem` left padding.
However, the checkbox above didn't have a set size. This caused different web browsers to render the size of the button differently.

The checkbox now has a size of `.8rem` wide and tall, with a `.1rem` margin on all sides.

mikehenrty:
Fixed in #343

PR #343, opened 2017-07-24
Specify size for check boxes

DeathCamel57:
Proposed fix for #178

### DIFF
--- a/web/css/index.css
+++ b/web/css/index.css
@@ -1405,6 +1405,12 @@ header.active {
   margin-top: 1rem;
 }
 
+input[type="checkbox"]{
+  width: .8rem;
+  height: .8rem;
+  margin: .1rem .1rem .1rem .1rem;
+}
+
 label[for="send-emails"] {
   font-size: 1rem;
 }


DeathCamel57:
This should line up the text on the checkbox line, with the line below it.

mikehenrty:
Thanks for the contribution @DeathCamel57!

The issue #178 is about text alignment with the below the check "I'd like to stay informed..". Your change only deals with the current line.

DeathCamel57:
@mikehenrty Though the change only affects the first line, the line below it already has a padding of 1 rem. So making the checkbox 1rem wide and tall makes the two lines line up properly.
screenshot from 2017-07-25 14-52-07

DeathCamel57:
Adding a pipe at the start of both lines makes it clear that they now line up. The pipe is not a code change, just a developer tools edit to show that they line up. The pipes ** are not** added to the code
screenshot from 2017-07-25 14-55-14

mikehenrty:
Oh, so sorry! I missed that somehow. Approved :)

PR #350, opened 2017-07-24
rename nubis file to deployment

mikehenrty:
@gozer any chance we could rename the `nubis` folder to `deployment` so that newcomers know what to expect? We could also do something like `deployment/nubis` is that's an easier change from your end.


DeathCamel57:
@mikehenrty I love this idea. Thank you for coming up with this! :+1:

mikehenrty:
Looks like the naming convention is a requirement of the Nubis project.
http://nubis-docs.readthedocs.io/en/latest/

PR #357, opened 2017-07-25
Clarify dataset license in FAQ

Omniscimus:
Tried to eliminate confusion about the dataset's license, as expressed in [a thread on Discourse](https://discourse.mozilla-community.org/t/sharing-the-dataset/17390). I found issue #7 where CC-0 is mentioned.

Is the thing I wrote about the free download true? How are you going to provide a download if you don't mind me asking? I suppose it will be a huge amount of data.


mikehenrty:
Saying something is CC-0 is same thing as saying free. I'll remove that part about the download.

Issue #243, opened 2017-06-24
Add instructions for classification


skorokithakis:
I tried to classify some sentences but wasn't sure what to do when people misread. For example, in one sentence, someone misread "in" for "is". Is that sentence supposed to be correct or not?

Some guidelines on the page would be great.

kdavis-mozilla:
If they don't read exactly the words in the sentence, then it should be counted as a "Nope"

skorokithakis:
Ah, good to know, thanks.

DeathCamel57:
#358 should help everyone with this question.

mikehenrty:
Duplicate of #273.

PR #358, opened 2017-07-26
Fix for #243

DeathCamel57:
Added short validation instructions.


mikehenrty:
Hi there @DeathCamel57, I'm going to put some thought into this to really try and address all the concerns in #243. I appreciate the thought here, but I need to engage my team internally to before we implement a fix.

Issue #341, opened 2017-07-23
Write some nice, short sentences for people to read.


mikehenrty:
Due to bugs #23 and #319 in our current sentence collection, we are trying to diversify and refresh our sentences in #333. We would love your help! If you would like to contribute to Common Voice sentences with your own writings, please put your sentences (one per line) in a publicly linkable document (eg. [Pastebin](https://pastebin.com/)), then add a comment to this bug with a link to those sentences.

Criteria:
* please write them yourselves. don't copy and paste from somewhere else.
* try to make the sentences conversational, ie. easy to read out load.
* you must agree to releasing your sentences to the public domain with a [cc-0 license](https://creativecommons.org/share-your-work/public-domain/cc0/).
* more than 50 sentences per link, but less than 500 please.
* be nice, don't use offensive language. we aren't collecting that kinda material.
* i'll be reading each one, and i may remove some but i will let you know why.

Thanks!

ftrotter:
Given that this is a specific bug... and is therefore addressable as a software item. I have elsewhere offered up already public domain resources as good sources for sentences:

https://en.wikipedia.org/wiki/Wikipedia:Public_domain_resources

As an open healthcare data person, I obviously would love the opportunity to inject your sentence data set with laymans healthcare terms. Assuming no one else is available to incorporate the various resources that I am discussing here, I might be able to devote some resources towards scrapping some of the resources that have solid APIs, and running them through some per-source filters that would serve to ensure that they do not have unusual acronyms or other confusing industry jargon. 

I would recommend that you develop a corpus that details the following information for each source of sentences:

* The sentence itself
* The url source for the sentence
* A reference to the rule (or contribution agreement to the cc-0 license) that provides evidence that the sentence in question is available under the public domain. 

I think I could write some scripts that would generate a few million (at least) sentences that met some basic rules that I could demonstrate were public domain. or someone else could if I fail to deliver (which I should warn is a frequent occurrence, this is only tangentially related to my day job, after all) 

Let me know if this would be helpful.

Regards,
-FT

ajaydee:
I'm back! Here's another hundred:

https://pastebin.com/1CU7GQYs

CC-0 or whatever license you need. There's a good few simple sentences.

akshit13:
Hope this helps.
https://pastebin.com/BRsfLuNt

mikehenrty:
[...]

PR #361, opened 2017-07-26
Create mycroft_btotharye.txt

btotharye:
adding in the mcyroft file for data from crime and punishment

### DIFF
--- a/server/data/mycroft_btotharye.txt
+++ b/server/data/mycroft_btotharye.txt
@@ -1,0 +1,38 @@
+"And what if there has been a search already? 
+What if I find them in my room?"
+But here was his room.
+Nothing and no one in it.
+No one had peeped in.
+Even Nastasya had not touched it.
+But heavens! how could he have left all those things in the hole?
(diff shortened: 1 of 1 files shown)


mikehenrty:
Thanks @btotharye !

We have to be careful about using old texts like this as they can be quite biased. For instance, the text you have here mention `he` 32 times, but never once `she`. This is why I'm trying to get people to submit stuff they wrote in #341, the language is more updated and we can control bias a little better.

That said, I definitely welcome your sentences! Maybe you can write some of your own for the public domain <3? If you want to use old texts, please correct the pronoun use to be more equal. Also I noticed some sentences start with quotes but have no closing quote. My [text tools](https://github.com/mikehenrty/screenplay2sentences/) did similar things. Please try to clean those up.

btotharye:
Ok np wasn't sure what exactly was needed

mikehenrty:
Yeah, we really need something in the README for this. Maybe I can add that now.

PR #367, opened 2017-07-26
Adding OS dependencies on readme

augustnmonteiro:
Adding a section on readme to list the OS dependencies, in some linux distributions the ffmpeg isn't default so the file is sent to s3 without any data

@MycroftAI
https://mycroft.ai/


mikehenrty:
We've since removed the dependency on ffmpeg.

PR #383, opened 2017-07-29
Fix MySQL error if using empty password

Omniscimus:
When an empty string was specified as the password in config.json, then [the createDb.js script](https://github.com/Omniscimus/voice-web/blob/4386d083f4a12a1260e2a29ee6a26030b60492fe/tools/createDb.js) failed with the following error:

`database create error ER_ACCESS_DENIED_ERROR: Access denied for user 'root'@'localhost' (using password: YES)`

The cause of this error was that empty strings are evaluated to `false` by JavaScript and TypeScript. [In the mysql.ts file](https://github.com/Omniscimus/voice-web/blob/4386d083f4a12a1260e2a29ee6a26030b60492fe/server/src/lib/db/mysql.ts#L37), config values were determined using the OR-operand. If one of `options.password` or `config.MYSQLPASS` was an empty string, then it was ignored, and the next one was used.

Fixed this by explicitly checking if those variables are empty strings.

### DIFF
--- a/server/src/lib/db/mysql.ts
+++ b/server/src/lib/db/mysql.ts
@@ -15,7 +15,7 @@ type MysqlOptions = {
 const DEFAULTS = {
   user: 'voiceweb',
   database: 'voiceweb',
-  password: '',
+  password: null,
   host: 'localhost',
   port: 3306,
   max: 10,
@@ -42,7 +42,17 @@ export default class Mysql {
             DEFAULTS.idleTimeoutMillis,
     };
 
-    this.pool  = mysql.createPool({
+    let isEmptyString = function(str: string): boolean {
+      return typeof str !== 'undefined'
+           && !str.length;
+    };
+
+    if (isEmptyString(options.password)
+        || (!options.password && isEmptyString(config.MYSQLPASS))) {
+      myConfig.password = null;
+    }
+
+    this.pool = mysql.createPool({
         connectionLimit : 100,
         host            : myConfig.host,
         user            : myConfig.user,
--- a/server/src/lib/db/mysql.ts
+++ b/server/src/lib/db/mysql.ts
@@ -42,13 +42,12 @@ export default class Mysql {
             DEFAULTS.idleTimeoutMillis,
     };
 
-    let isEmptyString = function(str: string): boolean {
-      return typeof str !== 'undefined'
-           && !str.length;
-    };
-
-    if (isEmptyString(options.password)
-        || (!options.password && isEmptyString(config.MYSQLPASS))) {
+    // Empty strings are evaluated to false, so if the password is an empty
+    // string, the next one in the order of priority would be used. Prevent
+    // this.
+    if (options.password === ''
+        || (!options.password && config.MYSQLPASS === '')) {
+      // An empty password is equivalent to no password.
       myConfig.password = null;
     }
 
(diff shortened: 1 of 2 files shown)


server/src/lib/db/mysql.ts
```diff
@@ -42,7 +42,17 @@ export default class Mysql {
             DEFAULTS.idleTimeoutMillis,
     };
 
-    this.pool  = mysql.createPool({
+    let isEmptyString = function(str: string): boolean {
```
mikehenrty: let's move this function into the server utility module (`serser/src/libe/utility.ts`)

server/src/lib/db/mysql.ts
```diff
@@ -42,7 +42,17 @@ export default class Mysql {
...
+      return typeof str !== 'undefined'
```
mikehenrty: could we just use `str === ''` ?

server/src/lib/db/mysql.ts
```diff
@@ -42,7 +42,17 @@ export default class Mysql {
...
+           && !str.length;
+    };
+
+    if (isEmptyString(options.password)
```
mikehenrty: this is a little hard to understand what you are doing here. can we add a comment?

server/src/lib/db/mysql.ts
Omniscimus: You're right 😆 Might as well just omit the function. Thanks for the feedback!

server/src/lib/db/mysql.ts
```diff
@@ -15,7 +15,7 @@ type MysqlOptions = {
 const DEFAULTS = {
   user: 'voiceweb',
   database: 'voiceweb',
-  password: '',
+  password: null,
```
mikehenrty: Do we still need to change this? Is `null` equivalent to empty string for mysql?

server/src/lib/db/mysql.ts
```diff
@@ -42,7 +42,16 @@ export default class Mysql {
...
 
-    this.pool  = mysql.createPool({
+    // Empty strings are evaluated to false, so if the password is an empty
+    // string, the next one in the order of priority would be used. Prevent
+    // this.
+    if (options.password === ''
+        || (!options.password && config.MYSQLPASS === '')) {
+      // An empty password is equivalent to no password.
+      myConfig.password = null;
+    }
```
mikehenrty: After thinking about this some more, I think what we want to do is remove line 37 above, and just calculate the password here. Perhaps something like:
```
if (options.password !== undefined) {
  myConfig.password = options.password; {
else if (config.MYSQLPASS !== undefined) {
  myConfig.password ......
}
```

This way is more readable I think.

Omniscimus:
@mikehenrty How about this 45a0fab7e2b0672285a66ced5ba557277f67e6ff ? Used a function instead of the OR operator.

server/src/lib/utility.ts
```diff
@@ -8,3 +8,16 @@
 export function getFileExt(path: string): string {
   return path.substr(path.lastIndexOf('.') - path.length);
 }
+
+/**
+ * Returns the first defined argument. Returns null if there are no defined
+ * arguments.
+ */
+export function getFirstDefined(...options) {
+  for (var i = 0; i < arguments.length; i++) {
```
mikehenrty: nit: let's use `options` instead of `arguments` since you nicely used the spread operator above.

mikehenrty:
👍

PR #400, opened 2017-08-03
fix grammar

jf99:


server/data/rupshabagchi.txt
```diff
@@ -7,7 +7,7 @@ This is the moment I have been waiting for.
 Yesterday my wife bought new pillows for herself.
 Her life is a series of plot twists.
 Here's a joke I like.
-While English is the standard language spoken in his country, the United States of America has no official language.
+While English is the standard language spoken in his country, the United States of America have no official language.
```
mikehenrty: Counter-intuitively, "the United States of American" is a singular noun, so has is correct.

http://grammarist.com/usage/united-states/

jf99:
Oh, thanks for pointing that out. In German "die USA" would be plural and I'm always annoyed when people say it wrong...

jf99:
Note that I've added another commit.

mikehenrty:
Thanks @jf99!

Speaking of German, are you aware of any large and up-to-date collection of public domain German text? We are looking to add a second language soon, and German is very interesting to us. That said, we don't really know how to look for the text, so any leads would be much appreciated.

Note: we are in Berlin! You are welcome to stop by our office and say hi if you are ever around this area.

jf99:
@mikehenrty
Hmm, I searched a little and found http://www.opencritics.de/. The site has closed, however the wayback machine still has snapshots of it: https://web.archive.org/web/20161106082553/http://www.opencritics.de/
Not sure whether it's really "gemeinfrei", though.

Another idea: What about subtitles from ARD, ZDF and other public-law TV stations? You would probably have to ask them, but maybe they can grant you the rights at least for their own productions (Tagesschau, heute-journal). I like the intention to add German as second language.

About Berlin: Thanks for the hint. I don't live in Berlin, but in fact I'm going to finish my master studies next year. Mozilla is now on my list of potential future employers :)  Especially now that you're doing interesting STT and machine learning stuff.

jf99:
Asking the authors of https://netzpolitik.org/ could also be worth a try. Their work is under CC BY-NC-SA 4.0, which is obviously more restrictive than public domain, but maybe they're willing to make an exception for some articles if you explain what you need it for.

mikehenrty:
> Edit: Maybe it's better to open an issue for this discussion?

Or could be added to #23 

Thanks for you ideas!

Issue #397, opened 2017-08-02
Errors in Firefox browser console when recording


Omniscimus:
### Browsers
- Firefox 54.0.1 (64-bit)
- Firefox 55.0b13 (64-bit)

Does NOT occur in Chrome 59.0.3071.115 (Official Build) (64-bit)

### Operating systems
- macOS 10.12.6

### Description

While recording in Firefox, some errors pop up in the console. Everything on the pages seems to be working fine. In the Network tab of the developer tools, it looks like the recordings do get uploaded.

When navigating to [https://voice.mozilla.org/record](https://voice.mozilla.org/record), I get the following error 3 times: (link points towards [https://voice.mozilla.org/record](https://voice.mozilla.org/record))
`Invalid URI. Load of media resource  failed. record`

When clicking 'submit' after recording three sentences: (links are: [upload:1:1](https://voice.mozilla.org/upload/), [record](https://voice.mozilla.org/record), [demographic:1:1](https://voice.mozilla.org/upload/demographic/))
```
XML Parsing Error: syntax error
Location: https://voice.mozilla.org/upload/
Line Number 1, Column 1:  upload:1:1

Invalid URI. Load of media resource  failed. record

XML Parsing Error: syntax error
Location: https://voice.mozilla.org/upload/
Line Number 1, Column 1:  upload:1:1

Invalid URI. Load of media resource  failed.  record

XML Parsing Error: syntax error
Location: https://voice.mozilla.org/upload/
Line Number 1, Column 1:  upload:1:1

XML Parsing Error: syntax error
Location: https://voice.mozilla.org/upload/demographic/
Line Number 1, Column 1:  demographic:1:1
```

Is there any way for me to debug this without setting up an Amazon S3 account for testing? If not, how do I set up one? I noticed that they have some [free plans](https://aws.amazon.com/free/), which one would I need? If setting up a local test server is all too much of a hassle then that's fine, I'll work on other issues.

mikehenrty:
I don't think you need an s3 account to test and debug this. The problem is not with s3, but the fact that the response is being interpreted by the browser as HTML, and instead we are sending text. I think we just need to set the right response header in nodejs ([here](https://github.com/mozilla/voice-web/blob/master/server/src/lib/clip.ts#L200) for instance).

If you would like to set up s3 for local testing, I do believe you can use a free plan, and then add aws creds like this:
http://docs.aws.amazon.com/AWSImportExport/latest/DG/SaveCredentials.html

[...]

PR #411, opened 2017-08-08
respond function for server responses

Omniscimus:
Made a class which can be used to send server responses. This fixes the XML errors in the Firefox console (#397). The cause of these errors was that the Content-Type header wasn't being set. The Responder class specifies some defaults. For example, if the Content-Type is not set, it uses `text/plain`. Using this class should prevent similar issues in the future.


mikehenrty:
@Omniscimus First of all, this is really nice code! And thank you for taking this on.

That said, I think we need to rethink this approach. First, we are allocating new objects for every request, which causes work for the garbage collector at the end of the day, and in a high request enviroment I could see this slowing down the app somewhat.

More generally though, I think a good approach would be to use a library to handle our static requests. Initially we were using [node-static](https://www.npmjs.com/package/node-static), but we took that out because we were having issues with static requests. It turns out, this was not related to node static. So i think the best approach would be to use a library rather than building out a fully featured static backend. What do you think @Omniscimus ?

Omniscimus:
@mikehenrty Ah, I hadn't thought of garbage collection.

Todo:
- use node-static instead of static.ts (revert  be2789adfec2310a4487b45927be81bb8bf5c5bd)
- use some kind of function instead of the Responder class to generalize non-static server responses

Omniscimus:
@mikehenrty How about this? Everything seemed to work on my localhost, but maybe I overlooked something.

mikehenrty:
Sorry about the delay here, this looks good! 👍

PR #412, opened 2017-08-08
Fix issue with record button not centered

### LINKED ISSUE #164: Record button is not centered.
image

kenrick95:
Elements with `display: inline-block` could introduces a space between elements: https://stackoverflow.com/q/5078239

Looks better now? Fixes #164

capture-centered

### DIFF
--- a/web/css/index.css
+++ b/web/css/index.css
@@ -1316,13 +1316,12 @@ body:not(.ios) .recording #background-container {
 }
 
 #record-button {
-  display: inline-block;
   border-radius: 50%;
   border: 0.6rem solid black;
   background-color: white;
   height: 2.5rem;
   width: 2.5rem;
-  margin: 1rem 0;
+  margin: 1rem auto;
   cursor: pointer;
 }
 
--- a/web/css/index.css
+++ b/web/css/index.css
@@ -1322,6 +1322,7 @@ body:not(.ios) .recording #background-container {
   height: 2.5rem;
   width: 2.5rem;
   margin: 1rem auto;
+  margin-top: calc(1rem + 4px);
   cursor: pointer;
 }
 
@@ -1356,6 +1357,8 @@ body:not(.ios) .recording #background-container {
 #record-help {
   font-size: 1rem;
   padding: 0 2rem;
+  line-height: 1em;
+  margin-bottom: 2px;
 }
 
 #voice-submit {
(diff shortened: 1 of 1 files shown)


Omniscimus:
`margin: 1rem 0` means that the top and bottom margins are 1rem, and left and right margins are 0. So changing the `0` to `auto` doesn't do much.

As can be seen on the picture, the margin of the button overlaps the bottom of the black shadow of the `record-sentence` div. This div has some margin at the top, but not at the bottom. Maybe adding some margin-bottom to it would help?

The button is also slightly more above the text because the text has some top padding by default. This is so that in a paragraph, there is some space in between the lines. This could be corrected by adding slightly more margin between the record-sentence div and the button (either at the div bottom or at the button top).

kenrick95:
@Omniscimus Thanks for you comments. Let me reply in three sections:

----

Firstly, when `#record-button` is on `display: inline-block`, it makes the whole `#voice-record` having an extra space between record button and help text. (I inspected each of the elements and could not find what accounts for this space, it is neither element height, padding, nor margin)

capture-init

After I removed that `display: inline-block` style, by default, the element will have `display: block` style, in which, by default, does not align to middle. Hence, I added 'auto' to left and right margin of that style. By the way, removing `display: inline-block` style makes the unaccounted space go away.

capture-inline-block

----

Secondly, unfortunately, this won't work because of collapsing margin. :cry: One workaround that I could think of is to use `margin-top: calc(1rem + 4px);` at the record button (this is because, the record button has 1rem of margin-top and the `.record-sentence` element has 4px of box-shadow)

----

Thirdly, I think what we should do is to enforce the `line-height` so it matches the `font-size` (which is set to 1rem), hence making the element height more precise, removing the issue of "text has some top padding by default".

---

Let me know what you think. Thanks :)

Omniscimus:
Looks pretty centered to me now :smile:

On second thoughts, I don't think the line height issue is very important, it can just be left without adjustments imo. I think the default line height is preferable especially on small devices, when the text wraps to the next line:

1em:

How it looks with 1em line height

default:

How it looks by default

And I don't really see the difference in vertical centering between these two anyway 😆 I believe the `margin-bottom: 2px;` adds bottom margin to the entire text box. It could also be corrected by adding one or two more pixels to the top margin of the button.

kenrick95:
Ah, I never thought of how it looks on small screen device. Thanks for pointing it out. Yeah, you're right, it's easier to read without the `line-height: 1em`.

mikehenrty:
thanks @kenrick95 !

PR #414, opened 2017-08-09
Fix typo

### LINKED ISSUE #413: Typo in sentence "We used all your imgination to design the place"
The word _im**a**gination_ misses _**a**_.

Since this is a very minor kind of issue, what's the best way to report them? Is it worth to create new issue for each case? Or maybe rename this one to something like "Typo reports" and report in comments? Or you have already defined another way (sorry, if I couldn't find it)?

mds3dstn71:
This commit resolves #413


kdavis-mozilla:
Thanks @mds3dstn71

PR #415, opened 2017-08-10
change british english to england english, fixes #404

### LINKED ISSUE #404: British, Scottish and Welsh English?
There's a rather odd choice of accent for the resident of Great Britain (which encompasses England, Wales and Scotland).  These are often lumped together as "British English" although there are vast differences in pronunciation between them.

For this project however, Welsh English and Scottish English have been separated out - perfectly sensible since this project is to do with pronunciation - but what remains is then English English, not British English.  It might be necessary to clarify it as "English as spoken in England" or something like that.  As it stands, inhabitants of Wales and Scotland have two logical choices from the list ("British" or "Welsh", and "British" or "Scottish") whilst inhabitants of England have just one which lumps them in with the Welsh and the Scots.

mikehenrty:


mikehenrty:
Wanna take a look @kdavis-mozilla >?

PR #422, opened 2017-08-12
Reset listen box state

### LINKED ISSUE #419: Pause button on submit page stays paused if submitting while playing
Doesn't really affect the usability but thought I am reporting it. 

The last button always shows pause instead of play although it is not playing.

image

kenrick95:
Fixes #419


mikehenrty:
Great solution

Issue #402, opened 2017-08-05
Web code style: Use async/await?


kenrick95:
Would you like to try changing the coding style from using Promise.then into async/await? Thanks

One example of change (on src/lib/api.ts):

```ts
private getRandomSentences(count: number): Promise<string[]> {
  return this.getSentences().then(sentences => {
    let randoms = [];
    for (var i = 0; i < count; i++) {
      let distribution = Random.integer(0, sentences.length - 1);
      let randomIndex = distribution(this.randomEngine);
      randoms.push(sentences[randomIndex]);
    }
    return randoms;
  });
}
```

become:

```ts
private async getRandomSentences(count: number): Promise<string[]> {
  const sentences = await this.getSentences();
  let randoms = [];
  for (var i = 0; i < count; i++) {
    let distribution = Random.integer(0, sentences.length - 1);
    let randomIndex = distribution(this.randomEngine);
    randoms.push(sentences[randomIndex]);
  }
  return randoms;
}
```

mikehenrty:
I definitely want to! Looking at [caniuse](http://caniuse.com/#feat=async-functions), it seems less than 70% of the global audience supports this feature yet, so it might be a bit soon for this. What do you think @kenrick95?

kenrick95:
Since we are using [TypeScript ^2.3.4](https://github.com/mozilla/voice-web/blob/master/package.json#L35) and [targeting transpilation to ES5](https://github.com/mozilla/voice-web/blob/master/web/tsconfig.json#L3), I think we are fine, because TypeScript, since [v2.1](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#downlevel-async-functions) will generate ES5-supported codes for async functions (although we still might need to add a polyfill for Promise)

nevik:
But adding a Promise polyfill or not is independent from using TS async/await.
This TS feature is syntactic sugar only, so it would be great to start using it 👍

nevik:
@mikehenrty, I just noticed while looking through the changes for #455: Should we also start using async/await in the backend code? #430 took care of pretty much everything in the frontend, but there's a bunch of promise-based async stuff in the backend as well

mikehenrty:
@nevik yeah, would love to start doing this on the backend as well. there's still some promised based stuff in the front-end here and there too I believe.

nevik:
@mikehenrty had a go at converting the server code to async/await, but ran into the following problem:

[...]

PR #430, opened 2017-08-15
Change code style to async/await in web folder

kenrick95:
First step for #402

### DIFF
--- a/web/src/lib/api.ts
+++ b/web/src/lib/api.ts
@@ -59,11 +59,9 @@ export default class API {
     });
   }
 
-  private fetchText(path: string): Promise<any> {
-    return this.fetch(path, { responseType: 'text' })
-      .then((req: XMLHttpRequest) => {
-        return req.responseText;
-      });
+  private async fetchText(path: string): Promise<any> {
+    const req = await this.fetch(path, { responseType: 'text' });
+    return req.responseText;
   }
 
   private requestResourceText(resource): Promise<string> {
@@ -78,11 +76,9 @@ export default class API {
     return API.DEFAULT_BASE + resource;
   }
 
-  getRandomSentences(count?: number): Promise<string[]> {
-    return this.requestResourceText('sentence' + (count ? '/' + count : ''))
-      .then(sentencesText => {
-        return sentencesText.split('\n');
-      });
+  async getRandomSentences(count?: number): Promise<string[]> {
+    const sentencesText = await this.requestResourceText('sentence' + (count ? '/' + count : ''))
+    return sentencesText.split('\n');
   }
 
   getTextFromUrl(url): Promise<string> {
@@ -92,25 +88,23 @@ export default class API {
   /**
    * Ask the server for a clip
    */
-  getRandomClip(): Promise<Clip> {
-    return this.fetch('upload/random/', { responseType: 'blob', headers: {'uid': this.user.getId()}})
-      .then((req: XMLHttpRequest) => {
-        let src = window.URL.createObjectURL(req.response);
-        let glob = decodeURIComponent(req.getResponseHeader('glob'));
-        let sentence = decodeURIComponent(req.getResponseHeader('sentence'));
-        return Promise.resolve({ glob: glob, audio: src, sentence: sentence });
-      });
+  async getRandomClip(): Promise<Clip> {
+    const req = await this.fetch('upload/random/', { responseType: 'blob', headers: {'uid': this.user.getId()}})
+    let src = window.URL.createObjectURL(await req.response);
+    let glob = decodeURIComponent(await req.getResponseHeader('glob'));
+    let sentence = decodeURIComponent(await req.getResponseHeader('sentence'));
+    return { glob: glob, audio: src, sentence: sentence };
+    
   }
 
   /**
    * Ask the server for a clip
    */
-  getRandomClipJson(): Promise<ClipJson> {
-    return this.fetch('upload/random.json', { responseType: 'json', headers: {'uid': this.user.getId()}})
-      .then((req: XMLHttpRequest) => {
-        let response = req.response as ClipJson;
-        return Promise.resolve(response);
-      });
+  async getRandomClipJson(): Promise<ClipJson> {
+    const req = await this.fetch('upload/random.json', { responseType: 'json', headers: {'uid': this.user.getId()}})
+
+    let response = req.response as ClipJson;
+    return await response;
   }
 
   castVote(glob: string, vote: boolean): Promise<Event> {
--- a/web/src/lib/api.ts
+++ b/web/src/lib/api.ts
@@ -77,7 +77,7 @@ export default class API {
   }
 
   async getRandomSentences(count?: number): Promise<string[]> {
-    const sentencesText = await this.requestResourceText('sentence' + (count ? '/' + count : ''))
+    const sentencesText = await this.requestResourceText('sentence' + (count ? '/' + count : ''));
     return sentencesText.split('\n');
   }
 
@@ -94,7 +94,6 @@ export default class API {
     let glob = decodeURIComponent(req.getResponseHeader('glob'));
     let sentence = decodeURIComponent(req.getResponseHeader('sentence'));
     return { glob: glob, audio: src, sentence: sentence };
-    
   }
 
   /**
--- a/web/src/lib/api.ts
+++ b/web/src/lib/api.ts
@@ -90,9 +90,9 @@ export default class API {
    */
   async getRandomClip(): Promise<Clip> {
     const req = await this.fetch('upload/random/', { responseType: 'blob', headers: {'uid': this.user.getId()}})
-    let src = window.URL.createObjectURL(await req.response);
-    let glob = decodeURIComponent(await req.getResponseHeader('glob'));
-    let sentence = decodeURIComponent(await req.getResponseHeader('sentence'));
+    let src = window.URL.createObjectURL(req.response);
+    let glob = decodeURIComponent(req.getResponseHeader('glob'));
+    let sentence = decodeURIComponent(req.getResponseHeader('sentence'));
     return { glob: glob, audio: src, sentence: sentence };
     
   }
@@ -101,10 +101,10 @@ export default class API {
    * Ask the server for a clip
    */
   async getRandomClipJson(): Promise<ClipJson> {
-    const req = await this.fetch('upload/random.json', { responseType: 'json', headers: {'uid': this.user.getId()}})
+    const req = await this.fetch('upload/random.json', { responseType: 'json', headers: {'uid': this.user.getId()}});
 
     let response = req.response as ClipJson;
-    return await response;
+    return response;
   }
 
   castVote(glob: string, vote: boolean): Promise<Event> {
--- a/web/src/lib/components/pages.tsx
+++ b/web/src/lib/components/pages.tsx
@@ -249,11 +249,19 @@ export default class Pages extends Component<PagesProps, PagesState> {
       const sentence = sentences.pop();
 
       await this.props.api.uploadAudio(blob, sentence);
-      let percentage = Math.floor((runningTotal / originalTotal) * 100);
-      progressCb && progressCb(percentage);
-      this.props.user.tallyRecording();
+
+      if (recordings.length !== 0) {
+        let percentage = Math.floor((runningTotal / originalTotal) * 100);
+        progressCb && progressCb(percentage);
+        this.props.user.tallyRecording();
+      }
     }
     await this.props.api.uploadDemographicInfo();
+
+    // Reset robot state to initial state
+    this.setState({
+      robot: ''
+    });
   }
 
   componentDidMount() {
--- a/web/src/lib/components/pages.tsx
+++ b/web/src/lib/components/pages.tsx
@@ -237,47 +237,42 @@ export default class Pages extends Component<PagesProps, PagesState> {
     });
   }
 
-  private uploadRecordings(recordings: any[],
+  private async uploadRecordings(recordings: any[],
                            sentences: string[],
                            progressCb: Function): Promise<void> {
+    try {
+      await this.ensurePrivacyAgreement();
+      let runningTotal = 0;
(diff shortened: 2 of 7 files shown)


web/src/lib/components/pages/record.tsx
```diff
@@ -268,11 +264,12 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
     this.props.onRecordStop && this.props.onRecordStop();
   }
 
-  newSentenceSet() {
+  async newSentenceSet() {
     // If we don't have any sentences in our cache, fill it and try again.
     if (this.sentenceCache.length < SET_COUNT) {
       console.error('slow path for getting new sentences');
-      this.refillSentenceCache().then(this.newSentenceSet.bind(this));
```
nevik: 🎉 for one less `.bind`!

web/src/lib/components/validator.tsx
```diff
@@ -29,29 +29,30 @@ export default class Validator extends Component<Props, State> {
...
+    try {
+      const clipJson = await this.props.api.getRandomClipJson();
       this.setState({
         loading: false,
         glob: clipJson.glob,
         sentence: decodeURIComponent(clipJson.text),
         audioSrc: clipJson.sound
       });
-    }, (err) => {
+    } catch (err) {
```
nevik: you have a space between the `catch` and the opening parenthesis here, while you don't have one on a `catch` further up. please make it consistent :)

web/src/lib/components/pages/record.tsx
```diff
@@ -96,11 +96,10 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
...
 
-  private refillSentenceCache() {
-    return this.props.api.getRandomSentences(CACHE_SET_COUNT)
-      .then(newSentences => {
-        this.sentenceCache = this.sentenceCache.concat(newSentences);
-      });
+  private async refillSentenceCache() {
+    const newSentences = await this.props.api.getRandomSentences(CACHE_SET_COUNT)
+    this.sentenceCache = this.sentenceCache.concat(newSentences);
+    return this.sentenceCache;
```
nevik: this return value is not declared and never used, so IMO nothing should be returned here (even though I think technically it did also return this value before)

web/src/lib/components/pages/record.tsx
```diff
@@ -268,11 +264,12 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
     this.props.onRecordStop && this.props.onRecordStop();
   }
 
-  newSentenceSet() {
+  async newSentenceSet() {
```
nevik: minor: this method can be private

web/src/lib/components/pages/record.tsx
```diff
@@ -268,11 +264,12 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
...
     // If we don't have any sentences in our cache, fill it and try again.
     if (this.sentenceCache.length < SET_COUNT) {
       console.error('slow path for getting new sentences');
-      this.refillSentenceCache().then(this.newSentenceSet.bind(this));
+      await this.refillSentenceCache();
+      this.newSentenceSet();
```
nevik: with async/await, we can neatly turn this recursive function into an iterative implementation:

```typescript
    // if we don't have enough sentences in our cache, fill it before continuing
    while (this.sentenceCache.length < SET_COUNT) {
      console.error('slow path for getting new sentences');
      await this.refillSentenceCache();
    }
```
(disclaimer: untested code)

web/src/lib/api.ts
```diff
@@ -92,25 +88,23 @@ export default class API {
...
-    return this.fetch('upload/random/', { responseType: 'blob', headers: {'uid': this.user.getId()}})
-      .then((req: XMLHttpRequest) => {
-        let src = window.URL.createObjectURL(req.response);
-        let glob = decodeURIComponent(req.getResponseHeader('glob'));
-        let sentence = decodeURIComponent(req.getResponseHeader('sentence'));
-        return Promise.resolve({ glob: glob, audio: src, sentence: sentence });
-      });
+  async getRandomClip(): Promise<Clip> {
+    const req = await this.fetch('upload/random/', { responseType: 'blob', headers: {'uid': this.user.getId()}})
+    let src = window.URL.createObjectURL(await req.response);
```
kenrick95: Ah yeah, thanks for noticing. I must have it confused with the new Fetch API (where response & response body are promises)

kenrick95:
@nevik Thanks for your code review. I've fixed all those points you mentioned, could you re-review it again? Thanks.

web/src/lib/components/pages.tsx
```diff
@@ -242,31 +242,21 @@ export default class Pages extends Component<PagesProps, PagesState> {
                            progressCb: Function): Promise<void> {
     try {
       await this.ensurePrivacyAgreement();
-      let runningTotal = 0;
-      let originalTotal = recordings.length;
+      const originalTotal = recordings.length;
 
       // This function calls itself recursively until
       // all recordings are uploaded.
```
nevik: minor: this code comment is no longer accurate ;)

kenrick95: Okay, removed. Thanks for spotting!

web/src/lib/components/pages.tsx
```diff
@@ -237,47 +237,29 @@ export default class Pages extends Component<PagesProps, PagesState> {
...
+        const blob = recording.blob;
+        const sentence = sentences.pop();
+
+        await this.props.api.uploadAudio(blob, sentence);
+        let percentage = Math.floor((runningTotal / originalTotal) * 100);
+        progressCb && progressCb(percentage);
+        this.props.user.tallyRecording();
+      }
+      await this.props.api.uploadDemographicInfo();
+    } finally {
```
mikehenrty: We don't want this `finally`, and probably not even the `try ... catch` at all.

The code I had before is *really* bad. The way I used the Promise is hard to read, so I will just describe in words what the code was doing in the likely case it was unclear.

This function `uploadRecordings` ultimately determines weather or not we display an error message on the Record page. What happens is, the Record page calls this function (as an `onSubmit` callback prop), and if the function fails it assumed the problem came from `ensurePrivacyAgreement` (the old Promise structure made sure of this in a *very* hacky way) and pops up a dialog asking the user what to do with the clips.

The proper thing to do here would be for `ensurePrivacyAgreement` to throw a specific error type, and catch that in the Record page before displaying the dialog. But in the end I think it will be enough to simply remove the `try...catch` here and change the error message in the dialog.

web/src/lib/components/pages.tsx
```diff
@@ -237,47 +237,29 @@ export default class Pages extends Component<PagesProps, PagesState> {
...
+    try {
+      await this.ensurePrivacyAgreement();
+      const originalTotal = recordings.length;
+
+      for (let runningTotal = 1; runningTotal <= originalTotal; runningTotal++) {
+        const recording = recordings.pop();
+        const blob = recording.blob;
+        const sentence = sentences.pop();
+
+        await this.props.api.uploadAudio(blob, sentence);
```
mikehenrty: note: if this fails, i believe it will now forward that error to the caller of this function, which didn't happen in the previous Promise structure. also note, your way is better. we probably want those errors :)

web/src/lib/components/pages/record.tsx
```diff
@@ -174,22 +172,20 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
...
-        })
+    try {
+      await this.props.onSubmit(this.state.recordings, this.state.sentences, this.onProgress);
+      this.reset();
+      this.tracker.trackSubmitRecordings();
+    } catch (e) {
+      this.setState({
+        uploading: false
       });
+      const keep = await confirm('You did not agree to our Terms of Service. Do you want to delete your recordings?', 'Keep the recordings', 'Delete my recordings')
```
mikehenrty: This catch can now happen if a single upload request failed (as well as not agreeing to privacy terms). Let's change this message to something like `Upload aborted. Do you want to delete your recordings?`

web/src/lib/components/pages/record.tsx
```diff
@@ -268,12 +263,11 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
     this.props.onRecordStop && this.props.onRecordStop();
   }
 
-  newSentenceSet() {
-    // If we don't have any sentences in our cache, fill it and try again.
-    if (this.sentenceCache.length < SET_COUNT) {
+  private async newSentenceSet() {
+    // if we don't have enough sentences in our cache, fill it before continuing
```
mikehenrty: I've learned to prefer full proper sentences as comments (ie. capitalize first letter, and period). I picked that up from Firefox devs. Comments are slightly easier to read when a proper sentence.

web/src/lib/components/pages/record.tsx
```diff
@@ -268,12 +263,11 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
...
+    while (this.sentenceCache.length < SET_COUNT) {
```
mikehenrty: wow! so much cleaner!

web/src/lib/components/pages/record/audio-web.ts
```diff
@@ -189,7 +189,7 @@ export default class AudioWeb {
     });
   }
 
-  stop() {
+  stop(): Promise<AudioInfo | {}> {
```
mikehenrty: 👍

web/src/lib/components/validator.tsx
mikehenrty: This reminds me, I really need to add a style linter.

mikehenrty: This reminds me, I really need to add a style linter.

nevik: Yes! #444

web/src/lib/api.ts
```diff
@@ -92,25 +88,23 @@ export default class API {
...
-  getRandomClip(): Promise<Clip> {
-    return this.fetch('upload/random/', { responseType: 'blob', headers: {'uid': this.user.getId()}})
-      .then((req: XMLHttpRequest) => {
-        let src = window.URL.createObjectURL(req.response);
-        let glob = decodeURIComponent(req.getResponseHeader('glob'));
-        let sentence = decodeURIComponent(req.getResponseHeader('sentence'));
-        return Promise.resolve({ glob: glob, audio: src, sentence: sentence });
-      });
+  async getRandomClip(): Promise<Clip> {
+    const req = await this.fetch('upload/random/', { responseType: 'blob', headers: {'uid': this.user.getId()}})
```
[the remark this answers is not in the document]
nevik: This was in response to my request, to use const unless necessary.

My reasoning is showing your intent - if you assume a variable to not be re-assigned, make it const, and only use let if you really need to re-assign. (I also prefer to use immutable structures where possible, so that philosophy goes along the same line.)
IOW, when touching existing code, const tells me I don't need to look around for any re-assignments of the identifier.

web/src/lib/components/pages.tsx
nevik: I think that should be changed in a separate PR though, so we don't mix logic changes with refactoring here

web/src/lib/components/pages/record.tsx
nevik: 👍

web/src/lib/components/pages.tsx
mikehenrty: @nevik we are already changing the logic in this refactor, that's what i was explaining. this change introduces a bug where if the user clicks on "don't accept" on the privacy policy, the robot turns into the wrong image on the background record page.

nevik:
Ah, then I misunderstood.
We should definitely fix it here, then :D

web/src/lib/api.ts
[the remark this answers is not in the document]
kenrick95: Yes, I agree with @nevik about usage of `const` unless `let` is necessary.

One benefit as I see it is so that the static analyzer (in this case, TypeScript) will throw error if we reassign the variable somewhere we don't expect the variable to be reassigned.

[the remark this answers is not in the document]
mikehenrty: Alright, I'm convinced by both your arguments. Thanks for the info!

web/src/lib/components/pages.tsx
kenrick95: Okay, I have more or less changed the part. Please verify if it works.

web/src/lib/components/pages/record.tsx
mikehenrty: nice

web/src/lib/components/pages.tsx
```diff
@@ -237,47 +237,23 @@ export default class Pages extends Component<PagesProps, PagesState> {
...
-            this.props.user.tallyRecording();
-            uploadNext();
-          });
-        };
-
-        // Start the recursive chain to upload the recordings serially.
-        uploadNext();
-      }).catch(rej);
-    }).then(() => {
-      this.setState({
```
mikehenrty: We still need to set the robot state after we complete the uploading.

web/src/lib/components/pages.tsx
```diff
@@ -237,47 +237,23 @@ export default class Pages extends Component<PagesProps, PagesState> {
...
+    await this.ensurePrivacyAgreement();
+    const originalTotal = recordings.length;
+
+    for (let runningTotal = 1; runningTotal <= originalTotal; runningTotal++) {
+      const recording = recordings.pop();
+      const blob = recording.blob;
+      const sentence = sentences.pop();
+
+      await this.props.api.uploadAudio(blob, sentence);
+      let percentage = Math.floor((runningTotal / originalTotal) * 100);
```
mikehenrty: To keep the same functionality as before, we need to not do the progress/tally work when we have uploaded all the clips. For simplicity, let's add a check here for `recordings.length === 0` and if so let's not execute these three lines.

kenrick95:
Alright, I've made the changes. Please verify again if it works. Thanks a lot for reviewing :)

mikehenrty:
Sweet! Thanks @kenrick95 !!

PR #432, opened 2017-08-16
Recording sense checks

### LINKED ISSUE #384: Sense checks on recordings
The app would benefit from some sense checks on recordings. If possible, both server-side (so contributors who validate recordings can focus more on non-garbage recordings) and client-side (so contributors who record can fix their recordings if they accidentally recorded something invalid). If we're not afraid of trolls, we could start with client-side and make it impossible to submit invalid recordings. Or even better, if the user insists on submitting them, pretend that they have been submitted but actually don't. (Obviously if it's client-side, a troll could disable these checks in the JavaScript.)

If we disallow contributors to submit recordings, then we should be sure that they're invalid. I think we could at least filter out the following:

- too short recordings (<1 second?)
- too long recordings (>10 seconds?)
- silent recordings (not sure how hard this is. Maybe detect volume level 0, or no change in pitch?)

Related: #301, #382.

Omniscimus:
Fixes #384.

I added client-side volume and length checks on recordings. If a recording is invalid, the user has to re-record it.

At the moment, the user doesn't get notified about what's wrong with the recording. I'm not great at UI. Can someone help me to [make some kind of alert appear](https://github.com/Omniscimus/voice-web/blob/75117895b4d8983aa1af8b0d099f2b20103431e2/web/src/lib/components/pages/record.tsx#L137) to notify the user?

Also, thoughts about the length bounds (>0.4s and <10s)?

### DIFF
--- a/web/src/lib/components/pages/record.tsx
+++ b/web/src/lib/components/pages/record.tsx
@@ -124,14 +124,6 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
 
     this.props.onRecordStop && this.props.onRecordStop();
 
-    if (!this.props.onRecordingSet) {
-      return;
-    }
-
-    if (this.isFull()) {
-      this.props.onRecordingSet();
-    }
-
     let error = this.checkRecording();
     if (error) {
       // Remove the invalid recording to go back.
@@ -143,6 +135,14 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
 
       console.log(error);
       // TODO display error to user
+    }
+
+    if (!this.props.onRecordingSet) {
+      return;
+    }
+
+    if (this.isFull()) {
+      this.props.onRecordingSet();
     }
   }
 
--- a/web/src/lib/components/pages/record.tsx
+++ b/web/src/lib/components/pages/record.tsx
@@ -19,9 +19,8 @@ const MIN_RECORDING_LENGTH = 300;   // ms
 const MAX_RECORDING_LENGTH = 10000; // ms
 const MIN_VOLUME = 1;
 
-enum RecordingValidity {
-  VALID,
-  TOO_SHORT,
+enum RecordingError {
+  TOO_SHORT = 1,
   TOO_LONG,
   TOO_QUIET
 };
@@ -45,7 +44,6 @@ interface RecordState {
   recording: boolean;
   recordingStartTime: number;
   recordingStopTime: number;
-  maxVolume: number;
   recordings: any[];
   uploading: boolean;
   uploadProgress: number;
@@ -58,13 +56,13 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
   isUnsupportedPlatform: boolean;
   tracker: Tracker;
   sentenceCache: string[];
+  maxVolume: number;
 
   state = {
     sentences: [],
     recording: false,
     recordingStartTime: 0,
     recordingStopTime: 0,
-    maxVolume: 0,
     recordings: [],
     uploading: false,
     uploadProgress: 0,
@@ -102,6 +100,8 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
       return;
     }
 
+    this.maxVolume = 0;
+
     // Bind now, to avoid memory leak when setting handler.
     this.onSubmit = this.onSubmit.bind(this);
     this.onRecordClick = this.onRecordClick.bind(this);
@@ -131,7 +131,7 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
 
     this.props.onRecordStop && this.props.onRecordStop();
 
-    const error = this.checkRecording();
+    const error = this.getRecordingError();
     if (error) {
       // Remove the invalid recording to go back.
       let recordings = this.state.recordings;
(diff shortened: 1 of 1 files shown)


web/src/lib/components/pages/record.tsx
```diff
@@ -117,6 +124,19 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
 
     this.props.onRecordStop && this.props.onRecordStop();
 
+    let error = this.checkRecording();
```
nevik: not sure if there are any stated style rules for this project yet, but if you don't (need to) re-assign a variable, using `const` is nicer :)

(same goes for the other variables you introduced or touched)

web/src/lib/components/pages/record.tsx
```diff
@@ -126,6 +146,19 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
     }
   }
 
+  private checkRecording(): string {
```
nevik: IMO it would be nice to return some kind of status code here (or introduce an enum for the result), so that the checking logic is more separated from the UI (error message)

Omniscimus:
Makes sense @nevik, thanks for the feedback!

web/src/lib/components/pages/record.tsx
```diff
@@ -15,6 +15,16 @@ import confirm from '../confirm';
 const CACHE_SET_COUNT = 9;
 const SET_COUNT = 3;
 const PAGE_NAME = 'record';
+const MIN_RECORDING_LENGTH = 400;   // ms
```
mikehenrty: Nice. Let's change this to 300 milliseconds, as we could have some 2 syllable sentences (eg. "hi there"), and according to this, unstressed syllables are 150ms long.

https://www.uni-bamberg.de/fileadmin/eng-ling/fs/Chapter_13/Index.html?22Theacousticmethod.html

web/src/lib/components/pages/record.tsx
```diff
@@ -117,6 +131,19 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
 
     this.props.onRecordStop && this.props.onRecordStop();
 
+    const error = this.checkRecording();
+    if (error) {
+      // Remove the invalid recording to go back.
```
mikehenrty: For now, let's only console.log the error. We will wait until we a visual display to the user before we disallow recordings based on this validation.

web/src/lib/components/pages/record.tsx
```diff
@@ -15,6 +15,16 @@ import confirm from '../confirm';
 const CACHE_SET_COUNT = 9;
 const SET_COUNT = 3;
 const PAGE_NAME = 'record';
+const MIN_RECORDING_LENGTH = 400;   // ms
+const MAX_RECORDING_LENGTH = 10000; // ms
+const MIN_VOLUME = 1;
+
+enum RecordingValidity {
```
mikehenrty: let's change this to `enum RecordingError`, and not have a VALID value.

web/src/lib/components/pages/record.tsx
```diff
@@ -126,6 +153,20 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
...
+    if (length < MIN_RECORDING_LENGTH) {
+      return RecordingValidity.TOO_SHORT;
+    }
+    if (length > MAX_RECORDING_LENGTH) {
+      return RecordingValidity.TOO_LONG;
+    }
+    if (this.state.maxVolume < MIN_VOLUME) {
+      return RecordingValidity.TOO_QUIET;
+    }
+    return RecordingValidity.VALID;
```
mikehenrty: let's just have this function return `null` when no error is found

web/src/lib/components/pages/record.tsx
```diff
@@ -126,6 +153,20 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
     }
   }
 
+  private checkRecording(): RecordingValidity {
```
mikehenrty: let's rename this function to `getRecordingError`

web/src/lib/components/pages/record.tsx
```diff
@@ -158,11 +199,21 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
...
-    if (!this.state.recording || !this.props.onVolume) {
+    if (!this.state.recording) {
       return;
     }
 
-    this.props.onVolume(volume);
+    // For some reason, volume is always exactly 100 at the end of the
+    // recording, even if it is silent; so ignore that.
+    if (volume !== 100 && volume > this.state.maxVolume) {
+      this.setState({
```
mikehenrty: Let's make `maxVolume` a property of the Record component, and not RecordState. RecordState is only for things that effect how the component is displayed, and since we only use `maxVolume` for validation checks, it doesn't belong in the state. This also has the side effect of triggering a render when we don't need to.

web/src/lib/components/pages/record.tsx
```diff
@@ -247,13 +298,17 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
     this.setState({
       recording: true,
       // TODO: reanble display of recording time at some point.
-      // recordingStartTime: this.audio.audioContext.currentTime
+      recordingStartTime: Date.now(),
```
mikehenrty: let's also reset `recordingStopTime` to null or 0

Omniscimus:
Thanks for the feedback, makes sense.

mikehenrty:
Looks good @Omniscimus ! Just so you know, we are getting a designer to help us with the error messaging part, so we will be able to use your code very soon!

PR #437, opened 2017-08-19
Add repository field to package.json

nevik:
Since I wasn't quick enough to comment on #433 to also add the `repository` field to `package.json`, here's a separate PR 😉


mikehenrty:
Thanks @nevik !

PR #421, opened 2017-08-11
Discourse link moved to blue bottom bar

deepakks1995:
I have moved the discourse link to the blue bottom bar

PR #439, opened 2017-08-19
Show names of external link targets for help links

nevik:
Since the Discourse link was recently moved from the footer links to the "help links" (cyan section above the very bottom footer) in #421, it was also re-labeled to "Community".

I think it's generally nice to tell the user where a link is going to lead them, especially external ones. And while some people may recognize the GitHub logo for the "Contribute" link, the label and icon for "Community" are so generic that I dont think most people can guess it leads to Discourse.

When I checked the [GUI specs](https://mozilla.github.io/voice-web/docs/GUI/#artboard2), I found that there's a suggestion to include names of external targets in the help links section as small text below the main label.

This change implements this naming for the GitHub and Discourse links, with some adjusted spacing.

Some notes:
* the GUI specs use a 20px font-size for the link labels (Help/Contribute/Community), while the page currently uses the page-wide default 15px font-size, and a smaller font-size (14px) for the names (GitHub/Discourse)
  * instead of making the label text bigger, I decided to scale down the name text proportionally (14px/20px = 70% -> I used `0.7rem`)
* since "Help" is an internal link, I don't think it needs a name below the label (and I couldn't think of a good one either)
  * this has the side effect that two of the three help links have a name, while one doesn't, so they have different visual heights
  * in order to properly align all three of them, I had to adjust the size and alignment of these help links
  * if we want to add a name below the "Help" link (and if someone has a good suggestion), I can easily adjust this PR to include that; that would also allow re-simplification of the style rules, since all three links would then be the same height

---

Screenshots:

| Platform | Before | After |
| --- | --- | --- |
| Desktop | screen shot 2017-08-19 at 12 40 19 | screen shot 2017-08-19 at 12 40 54 |
| Mobile | screen shot 2017-08-19 at 12 43 14 | screen shot 2017-08-19 at 12 43 02 |

### DIFF
--- a/web/css/index.css
+++ b/web/css/index.css
@@ -492,6 +492,8 @@ header.active {
   box-sizing: border-box;
   text-align: center;
   width: 10rem;
+  height: 4.8rem;
+  padding: .2rem 0;
 }
 
 #help-links a:first-child {
@@ -500,6 +502,14 @@ header.active {
 
 #help-links a:nth-child(2) {
   border-right: 1px solid black;
+}
+
+#help-links a .help-link-label {
+  margin-top: .2rem;
+}
+
+#help-links a .help-link-name {
+  font-size: .7rem;
 }
 
 #moz-links {
--- a/web/src/lib/components/pages.tsx
+++ b/web/src/lib/components/pages.tsx
@@ -383,17 +383,19 @@ export default class Pages extends Component<PagesProps, PagesState> {
               <a id="help" onClick={this.linkNavigate}
                  href="/faq">
                 <Icon type="help" />
-                <p class="strong">Help</p>
+                <p class="strong help-link-label">Help</p>
               </a>
               <a id="contribute"
                  target="_blank" href="https://github.com/mozilla/voice-web">
                 <Icon type="github" />
-                <p class="strong">Contribute</p>
+                <p class="strong help-link-label">Contribute</p>
+                <p class="help-link-name">on GitHub</p>
               </a>
               <a id="discourse"
                  target="blank" href="https://discourse.mozilla-community.org/c/voice">
                 <Icon type="discourse" />
-                <p class="strong">Community</p>
+                <p class="strong help-link-label">Community</p>
+                <p class="help-link-name">on Discourse</p>
               </a>
             </div>
           </div>


web/css/index.css
```diff
@@ -492,6 +492,8 @@ header.active {
   box-sizing: border-box;
   text-align: center;
   width: 10rem;
+  height: 4.8rem;
```
nevik: Height specification of the main link element is now necessary because the three links' contents are no longer the same height

mikehenrty:
I'm not a huge fan of the look of this change. What if we changed the titles to just Github and Discourse, rather than had sub headings?

nevik:
I can certainly do that

Most people are familiar enough with the GitHub name to recognize that is where code contributions go.

But from my own experience I can say: I was not familiar with Discourse before I looked into this project. So I'm a bit worried people might be a bit confused, especially since "Discourse" is a _very_ generic name for a specific website.

On the other hand, i'm certainly not a designer. @mikehenrty, could you counter-check this with a design/UX person? if they have any input or alternative suggestions, i'd love to hear them!

mikehenrty:
> @mikehenrty, could you counter-check this with a design/UX person? if they have any input or alternative suggestions, i'd love to hear them!

Good idea! @m-branson, could you give us some guidance on the changes @nevik is proposing?

m-branson:
Thanks for looping me in @mikehenrty! Nice usability point @nevik, appreciate you highlighting this. 👍 

My thoughts here are two-fold: 

1. While I can understand the need for further description at the links, UI preference is to keep this simplified (as one word labels). This keeps overall page legibility, responsiveness and information hierarchy in balance. Additionally, the labels serve as a simplified call to action (CTA) that allows us to infer what the link will allow our users to do; e.g. 'contribute'. @mikehenrty is there another word we could use for the Discourse link that better infers the action associated with it, perhaps 'connect'?
2. From a usability perspective, having more information here is ideal and I'd like to suggest implementing a slightly different solution where we utilize the longer, descriptive labels as tool tips. On hover users can gain more information (as well as be provided this information in a screen reader format) before clicking the link.

To summarize: Keep the single word labels, potentially changing 'Community' to 'Connect' and add tool tips with further description as @nevik pointed out. Thoughts on this approach?

nevik:
Thanks for your insights, @m-branson! 

While I see that tool tips are better in terms of information density here, tools tips don’t really work on mobile devices, since there is not hover state.

And I think that usability on mobile is important to this project, right @mikehenrty?

@m-branson, if you have any ideas how to make this nicely accessible on mobile, that would be great!

m-branson:
@nevik tooltips on mobile can indeed by tricky, nice flag. I recommend implementing them with a dotted underline on the text, this helps signal to users that the text itself is tap-able, but different than and separate from, the icon link itself. This can be used to enable tooltips on mobile via either touch or tap. Once the tooltip is revealed, it can be dismissed by tapping on the tooltip message itself.

Here are some links that discuss this implementation and help demo it visually: 

- https://ux.stackexchange.com/questions/35738/what-are-the-best-practices-for-tooltips-on-mobile-websites
- https://ux.stackexchange.com/questions/98628/how-can-i-display-tool-tips-on-a-touchscreen-or-is-there-a-better-way?noredirect=1&lq=1

cc @mikehenrty for thoughts and feedback

nevik:
@m-branson, I agree that those ways of implementing tool tips on mobile are a good way to provide additional information inline, but I don’t think they’re a good fit for our situation. Since the text element is quite small and close to the link icon, I don’t think it provides an adequate touch target for the user.

Unless a change is planned to the overall style of these “outgoing links”, and we don’t want to include target labels on the second line, I would actually agree with @mikehenrty’s initial comment and lease out the labels altogether.

Since these are “real” hyperlinks (i.e. not triggered by JavaScript), browsers always provide the user with a way to view the link target URL before navigating (desktop: shown at bottom when hovering, mobile: shown in context menu when long-pressing). In the iOS app, long-pressing is currently also showing the context menu, so this is covered as well.

m-branson:
Thanks @nevik, agreed that our touch target isn't the largest at this point. And yes, link targeting has us covered in the most basic sense at the moment.

As we're currently in the planning phase for greater design improvements, let's move forward with the simple approach. The icons lend a graphic assist on the context around 'GitHub' and 'Discourse' as labels. Further refinement of these touts and how they're presented to users is definitely on the design task list. 👍 

cc @mikehenrty

nevik:
Alright, then we leave the external links as they are for now :)

I guess that closes this PR for now, @mikehenrty? ;)

PR #442, opened 2017-08-21
Fix Invalid URI errors in Firefox console

Omniscimus:
This is a fix for the `Invalid URI. Load of media resource failed.` error in the Firefox console when loading the app (#397).

If the listen boxes were disabled, the `src` attribute of its `audio` element had the value of an empty string. This is invalid, because the src attribute on media elements must always contain a valid, non-empty URL if it is present ([source](https://html.spec.whatwg.org/#attr-media-src)).

### DIFF
--- a/web/src/lib/components/listen-box.tsx
+++ b/web/src/lib/components/listen-box.tsx
@@ -138,7 +138,9 @@ export default class ListenBox extends Component<Props, State> {
           <Icon type="x"/>Nope.</a>
       </div>
       <audio className="audio-box"
-        src={this.props.src}
+        // Only include the src attribute if the source is defined
+        // (empty src attributes are invalid)
+        {...this.props.src && {src: this.props.src}}
         preload="auto"
         onLoadStart={this.onLoadStart}
         onCanPlayThrough={this.onCanPlayThrough}


web/src/lib/components/listen-box.tsx
```diff
@@ -138,7 +138,9 @@ export default class ListenBox extends Component<Props, State> {
           <Icon type="x"/>Nope.</a>
       </div>
       <audio className="audio-box"
-        src={this.props.src}
+        // Only include the src attribute if the source is defined
+        // (empty src attributes are invalid)
+        {...this.props.src && {src: this.props.src}}
```
nevik: I'm a bit new to (p)react; why do we need the `...` here? Is it the TS spread operator?

Omniscimus: Here is [some documentation](https://facebook.github.io/react/docs/jsx-in-depth.html#spread-attributes).

nevik: Thanks for the link! That explains the syntax :)

TBH, I don't like the compactness of this implementation in terms of readability though.

Have you considered pulling out the entire `<audio>` element into a local `AudioBox` component (name up to you, of course), where you can set a local variable based on the value of `src`, and then use the spread operator with that local variable, instead of combining it inside the interpolation syntax?

Similar to what's showing here: https://facebook.github.io/react/docs/conditional-rendering.html

Omniscimus: That's another possibility. I'd be happy to make an AudioBox component, but I'm not sure if that's the best option, since we only need this once here. Does anyone else have thoughts about this?

mikehenrty: In general, I agree with @nevik that this is hard to read. But given that you have a comment explaining what this does, and the footprint is very small I'm going to allow it.

mikehenrty:
I think this change might have broke listen functionality on the iOS app, but I'm not sure why. I'm going to revert to see if this fixes it, and then we can re-investigate.

mikehenrty:
Well reverting this didn't fix it, the problem must be somewhere else. I'm going to re-add this code and continue to investigate the problem.

Issue #448, opened 2017-09-03
Using S3 buckets in eu-west-2 region gives saving clip error (InvalidRequest)


nmstoker:
Running the server locally, I muddled through the various steps needed (ie MySQL db, ffmpeg, S3 buckets) but I ran into a problem which I think is to do with having chosen to have my S3 buckets in the region **eu-west-2** (_which is nearest to me_)

When I submitted recordings via the localhost:9000 page, node output showed the following error message for each clip upload attempt:

`{"msg":"saving clip error, InvalidRequest: The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256.,`

I dug into it and this seems to be dealt with here: https://stackoverflow.com/a/31636292/142780

The fix for me was simply to update https://github.com/mozilla/voice-web/blob/master/server/src/lib/clip.ts#L32 from:

`this.s3 = new AWS.S3();`

to

`this.s3 = new AWS.S3( { signatureVersion: 'v4' } );`

Has anyone else seen this and/or is able to confirm if this is the cause? (I _think_ I'm right, but it could be some other config mistake I overlooked).

From what I read on links from that SO page, this would affect any buckets located in newer regions (which perhaps explains why it wasn't spotted with users of this repo if they went with the main US regions)

jgmize:
@nmstoker both your hypothesis and your fix look correct to me. Would you like to file a PR with this change?

nmstoker:
Thanks @jgmize - I've just submitted a PR as suggested (my first in a mainstream project :smile: )

mikehenrty:
Fixed by @nmstoker. Thanks!

PR #451, opened 2017-09-16
Fix for issue #448 - include AWS signature version to enable S3 buckets in new regions

nmstoker:
S3 buckets in newer AWS regions require the v4 signature version, although the older ones support v2.  The default region used in voice-web is older and thus does handle v2 but for anyone who switches to use buckets in a newer region will run into errors with the signature, as explained in issue #448.

This fix simply explicitly adds the v4 signature version so it works on all regions.

### DIFF
--- a/server/src/lib/clip.ts
+++ b/server/src/lib/clip.ts
@@ -29,7 +29,7 @@ export default class Clip {
   private files: Files;
 
   constructor() {
-    this.s3 = new AWS.S3();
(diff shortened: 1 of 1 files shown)


mikehenrty:
thanks @nmstoker for looking into this! we definitely want to support all s3 regions, so we want this change. i'm going to merge this and test it on our staging infrastructure before pushing to prod.

PR #455, opened 2017-09-24
Add linter to build process, fixes #444

### LINKED ISSUE #444: Code style: Apply prettier for consistent code formatting
[Prettier](https://github.com/prettier/prettier) is essentially a code formatter (initially for JS, but now also supports TS). It takes care of all code formatting so code contributors could just code in any style and then apply this code formatter so that code style across contributors are (mostly) consistent. Lots of open source projects already use Prettier, notable ones are React, Webpack, Yarn, Babel.

mikehenrty:


mikehenrty:
Hi @nevik and @kenrick95, can you give me some feedback about this?

The important part to review is in 0bb5f8379c7ff2f3aec0b1de7bb37371d9d289cd. The other commits are just the automatic consequences of that change.

This adds two commands to gulp, `lint` and `prettify`, and then adds linting to the build process. We run all lint/prettify commands through ESLint, which in turn uses Prettier as the style rules backend.

I prefer not to run any logic that alters code implicitly, so I didn't add prettifying to the build process (or in a post-commit hook). Instead you have to run prettify manually when linting fails. This is all by design. One thing that could be improved (which we can do in a follow-up) is have a nice error message for the lint step which asks the user to run `gulp prettify` to fix their errors.

Thanks for all your help!

nevik:
> I prefer not to run any logic that alters code implicitly, so I didn't add prettifying to the build process (or in a post-commit hook). Instead you have to run prettify manually when linting fails. This is all by design. One thing that could be improved (which we can do in a follow-up) is have a nice error message for the lint step which asks the user to run gulp prettify to fix their errors.

Sounds reasonable, that should be easy to add afterwards.

I'll start looking the changes now. Don't know if I'll get through it all today (if not, I'll finish up in the next few days).

web/src/lib/user.ts
```diff
@@ -6,84 +6,84 @@ const USER_KEY = 'userdata';
 interface UserAccents {
   [key: string]: string;
```
nevik: unrelated: hmmm... should we remove this additional key type declaration? we generally wouldn't want additional properties to be defined on these objects, right?

web/src/lib/utility.ts
```diff
@@ -6,8 +6,9 @@
  * Get some random string in a certain format.
```
nevik: haha, awfully vague docstring

server/src/lib/api.ts
```diff
@@ -101,41 +105,44 @@ export default class API {
...
-        });
-
-        sentences = sentences.concat.apply(sentences, sentenceArrays);
-        console.log('sentences found', sentences.length);
-        this.sentencesCache = sentences;
-      })
-      .catch((err: any) => {
-        console.error('could not retrieve sentences', err);
-      });
+    return (
```
nevik: another curious case of additional parens

server/src/lib/db.ts
```diff
@@ -18,9 +18,14 @@ export default class DB {
...
-    return this.user.create()
-        .then(() =>{return this.badges.create()})
-        .then(() =>{return this.progress.create()})
+    return this.user
+      .create()
+      .then(() => {
+        return this.badges.create();
+      })
+      .then(() => {
+        return this.progress.create();
```
nevik: minor: using short-form arrow functions (`() => this.progress.create()`, i.e. without block and `return`) would allow these to be more compact

server/src/lib/db/mysql.ts
```diff
@@ -67,8 +71,8 @@ export default class Mysql {
   }
 
   query(text: string, values: any[], callback: Function) {
-    this.pool.query(text, function (error: any, results: any, fields: any) {
-        error ? callback(error.message, null) : callback(null, results);
+    this.pool.query(text, function(error: any, results: any, fields: any) {
```
nevik: point about code style (which apparently prettier doesn't care about): for lambdas like this i think we should prefer arrow-functions over anonymous `function`s, to avoid accidental misuse of `this` inside the lambda

server/src/lib/files.ts
```diff
@@ -61,22 +60,23 @@ export default class Files {
...
-          return;
-        }
-
-        let sentence = s3Data.Body.toString();
-        this.files[glob].sentence = sentence;
-        res(sentence);
-      });
-    });
+    return new Promise(
+      (res: (stentence: string) => void, rej: (error: any) => void) => {
```
nevik: minor: typo in s>t<entence

mikehenrty:
Thanks for your comments @nevik ! I'll go over them in depth shortly.

> @kenrick95 @mikehenrty do you know if there's a (reasonable) way to adjust prettier's behavior (i.e. not maintaining a fork of prettier)? e.g. by supplying additional custom rules (like eslint/tslint support)

The advantage (and disadvantage) of Prettier over traditional linters like ESLint is how few [options](https://prettier.io/docs/en/options.html) it exposes. This broadly enforces a general style across all projects using Prettier, so I think if we really wanted custom stuff we should not use it.

With that said, I don't have any objects to Prettier's overall approach, so in the name of future consistency I've chosen it for this project. Let's try embracing it, and if we find ourselves not liking it we can drop later.

gulpfile.js
```diff
@@ -1,32 +1,35 @@
...
 let gulp = require('gulp-help')(require('gulp'));
 let shell = require('gulp-shell');
 let path = require('path');
 let ts = require('gulp-typescript');
 let insert = require('gulp-insert');
-
-function compile(project) {
-  return project.src().pipe(project()).js;
-}
+let eslint = require('gulp-eslint');
```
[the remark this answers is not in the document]
mikehenrty: agreed, at some point we can make a pass at converting all `let`s to `const`, but we won't be enforcing this in the linter yet.

server/src/lib/api.ts
mikehenrty: yup, this is a Prettier convention that I couldn't find a config for. i wouldn't write code like this, but it's not a deal-breaker either

server/src/lib/db.ts
mikehenrty: it seems Prettier expands all functions. i agree with the compactness, but don't mind the expansion either

web/src/lib/api.ts
```diff
@@ -77,7 +77,9 @@ export default class API {
   }
 
   async getRandomSentences(count?: number): Promise<string[]> {
-    const sentencesText = await this.requestResourceText('sentence' + (count ? '/' + count : ''));
+    const sentencesText = await this.requestResourceText(
+      'sentence' + (count ? '/' + count : '')
+    );
```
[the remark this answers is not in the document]
mikehenrty: 80 is my personal preference, as since it is also the Prettier default, I left it at that.

web/src/lib/components/logo.tsx
```diff
@@ -1,13 +1,18 @@
...
-    onClick={(evt) =>  {
-      evt.preventDefault();
-      evt.stopPropagation();
-      props.navigate('/');
-    }}>
-    <span class="main-title">Common Voice</span><br />
-    <img class="main-mozilla-logo" src="/img/mozilla.svg" />
-  </a>;
-}
+  return (
```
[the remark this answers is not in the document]
mikehenrty: yup, Prettier convention. i'm trying to get used to it too.

web/src/lib/user.ts
mikehenrty: since we import these interfaces in User and Profile components, and then dynamically access their various strings, we need to tell the TS compiler the type to expect when making this access. 

this was done to when we disallowed implicit anys in 087c10fc89d178708f91248d20da872942dd526c

web/src/lib/utility.ts
mikehenrty: yikes, yeah

web/src/lib/components/pages.tsx
```diff
@@ -39,20 +39,20 @@ const URLS: PageUrls = {
...
+  NOTFOUND: '/not-found',
 };
 
 const ROBOT_TALK = {
-  'home': [
+  home: [
     <p>Greetings human!</p>,
     <p>My name is M.A.R.S. and I am a learning robot.</p>,
     <p>Right now, I am learning to speak like a human.</p>,
-    <p>But. . .  it's so hard!</p>,
```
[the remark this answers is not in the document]
mikehenrty: I believe Prettier knows when you're in a jsx context vs string literal, so we *shouldn't* have to worry about this.

web/src/lib/components/pages.tsx
```diff
@@ -324,124 +326,198 @@ export default class Pages extends Component<PagesProps, PagesState> {
...
+                <div class="content">
+                  <a id="help" onClick={this.linkNavigate} href="/faq">
+                    <Icon type="help" />
+                    <p class="strong">Help</p>
+                  </a>
+                  <a
+                    id="contribute"
+                    target="_blank"
+                    href="https://github.com/mozilla/voice-web">
+                    <Icon type="github" />
```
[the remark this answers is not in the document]
mikehenrty: This would be easier to read if we used the Prettier default behavior for `jsxBracketSameLine`, but in general I think the closing bracket on the same line is more readable. Indeed, React core sets jsxBracketSameLine to true.

web/src/lib/components/pages/record.tsx
```diff
@@ -8,30 +8,38 @@ import AudioWeb, { AudioInfo } from './record/audio-web';
...
+  countSyllables,
+  isNativeIOS,
+  generateGUID,
+} from '../../utility';
 import confirm from '../confirm';
 
 const CACHE_SET_COUNT = 9;
 const SET_COUNT = 3;
 const PAGE_NAME = 'record';
-const MIN_RECORDING_LENGTH = 300;   // ms
```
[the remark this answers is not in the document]
mikehenrty: RIP

web/src/lib/components/logo.tsx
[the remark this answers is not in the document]
kenrick95: Yeah sometimes Prettier does weird things like adding `{' '}` at end of line (at JSX). 😅

web/src/lib/components/pages/record.tsx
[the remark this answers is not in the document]
kenrick95: If you want to exclude codes from prettier-ed, [you could have it](https://github.com/prettier/prettier#excluding-code-from-formatting).

web/src/lib/components/listen-box.tsx
```diff
@@ -130,38 +130,54 @@ export default class ListenBox extends Component<Props, State> {
...
-        <a onClick={this.voteNo}>
-          <Icon type="x"/>Nope.</a>
+    return (
+      <div
+        className={
+          'listen-box' +
+          (this.state.loaded ? ' loaded' : '') +
+          (this.state.playing ? ' playing' : '')
+        }>
+        <div className="sentence-box">
```
kenrick95: Just a note: even though Prettier is set to use single quote, somehow at JSX they enforces double quote at element properties.

mikehenrty: yup, i actually appreciate this, since it looks more like html to me

mikehenrty:
Great! Thanks for your reviews @nevik and @kenrick95. We'll use this as a starting point for our linting, and can discuss changes/tweaks as the project becomes more mature.

PR #468, opened 2017-09-30
Feature/fix hover style for ios

### LINKED ISSUE #424: iOS: Elements' hover state gets stuck after tapping them
Unlike desktop browsers, Safari on iOS does not handle the CSS `:hover` state well.

In particular, the hover state is applied when tapping an element, and is left applied until another element gets focus.

One solution is to not use `:hover` on iOS, but `:active` instead. The `:active` style is applied as one would expect - only while the finger is down during a tap, and removed when the tap is released.

Unfortunately, this requires some form of platform detection to distinguish the platforms/browsers and apply the appropriate CSS.
From a quick look, though, it seems a CSS class `safari` is already applied when loading the page on iOS. However, I have not counter-checked if this class is also applied on macOS (in which case, a more fine-grained distinction would be needed).

This is easy to reproduce - just open the page on iOS and tap the "play" button - its hover state is not unapplied until you tap something else. (This also works in Chrome's mobile/responsive device simulator. I've not tried it in Firefox's inspector.)
If desired, I can attach a short screen recording demonstrating the issue (a hint for where to host the video would be appreciated, GitHub doesn't allow attaching video files directly; elsewise, I'd just zip the file).

I'd be happy to help fixing this (or starting to, depending on how complex the stylesheets are), but first I'd like to clarify what approach to platform/browser detection is used or wanted in this project.

Or if this is considered too cosmetic and not something that should be fixed at the moment?

nevik:
Fixes #424.

Based on #466 (and contains the changes from there, for now, due to the way PRs work).

This PR disables hover styles on iOS, since they "get stuck" as described in detail in #424.

For the generic `button`, the `vote-box` buttons and the robot's bubble, I simple disabled the hover style without a new replacement, since there are already other `:active` styles in place (button, vote-box) or tapping the element immediately navigates away (robot's bubble), respectively.

For the `listen-box` elements (play/pause button and delete button), I adjusted the CSS to apply the style that was used for `:hover` until now to the `:active` state for iOS. This preserves visual feedback while the user is tapping the element, but does not get stuck like `:hover` does.

These changes apply to both Safari on iOS and the native iOS app. However, I could only test this in the iOS simulator on a Mac, since I don't currently have any iOS test devices around.  
Therefore it would be great if someone could double-test this.

### DIFF
--- a/web/css/index.css
+++ b/web/css/index.css
@@ -139,10 +139,8 @@ button {
   }
 }
 
-@media (min-width: 600px) {
-  button:hover:not(.ios) {
-    background-color: var(--color-speak);
-  }
+body:not(.mobile-safari) button:hover {
+  background-color: var(--color-speak);
 }
 
 button.highlight {
@@ -319,8 +317,10 @@ input:invalid {
   transition: 100ms;
 }
 
-.delete-box:hover,
-.play-box:hover {
+body:not(.mobile-safari) .delete-box:hover,
+body.mobile-safari .delete-box:active,
+body:not(.mobile-safari) .play-box:hover,
+body.mobile-safari .play-box:active {
   background-color: var(--color-listen-dark);
 }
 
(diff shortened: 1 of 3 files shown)


mikehenrty:
Tested on an iPhone and everything works great!

Issue #454, opened 2017-09-24
Front end architecture: Use preact-router instead of implementing own router


kenrick95:
Right now, I noticed that the page router is implemented in app.tsx and pages.tsx. This made pages.tsx quite long and contains lots of logic that are used only in one page (e.g. uploadRecordings are only used at Record page). My suggestion is to integrate this project with [preact-router](https://github.com/developit/preact-router) so we could abstract away the page routing logic. I think this will take big effort to do but I think it will worth the effort as it made implementing new pages much easier. Another advantage of using a routing library is that, we could benefit from lazy loading/code splitting (i.e. at first load, only the codes related to initial route are loaded)

mikehenrty:
I like this idea in theory, but a couple of things I would want to know:
1. how big is the routing library?
2. what benefits to code readability do we get?

Maybe a small proof of concept PR would help me understand the advantages.

PR #472, opened 2017-10-01
Try preact router

kenrick95:
This is how it looks like if we decided to use preact-router. (#454)

Take note of the removal of routing code at pages.tsx.

Another thing to take note is that I'm not sure how to handle page transition on this library.

Note: This PR, if compiled, produces a bundle which emits this error: `Error: See almond README: incorrect module build, no module name`


mikehenrty:
I haven't looked in depth yet, but overall I'm liking the approach of this. The include is only a couple of KB, and it removes a bunch of logic on our end. Plus we do things in a more "standard preact" way, which is useful for an open source project.

In the short term, I won't be able to give you much help here with the code @kenrick95. But if you want to keep going, 👍 from me, and I'll commit to reviewing and getting it merged in when everything is ready.

mikehenrty:
@kenrick95 thanks for this! let me know when you want myself or @nevik to take a look when it's ready, even if just for feedback.

kenrick95:
This is actually not ready yet because I'm stuck at the bundling part. More specifically, at this change, I have an import `Match` from `preact-router/match`, but there are actually no `preact-router/match` package (almond seems to be not able to pick `match` up from `preact-router`).

So I'm thinking of continuing this PR after #508 (which I believe is kind of fixing #494 issue)

kenrick95:
Since #494 is fixed, I think I'll start a new PR to integrate with preact-router as the parent of this branch is way out of date already

Gregoor:
@kenrick95 we're very likely going to move to react, so you might want to postpone that. Unless you think preact-router is so similar to react router that the move will be trivial?

Gregoor:
@kenrick95 I'm actually looking into `react-router` now, as React makes more assumptions about setState calls during the lifecycle than preact and I'm thinking that migrating our little frankenstein's monster that is `app.tsx` to `react-router` is simpler right now. Do you still want to help out with that or maybe review it when I'm done with it?

kenrick95:
@Gregoor Yeah, I could still help out with implementation of migration to react-router.

Issue #185, opened 2017-06-19
Long empty space on speak page before the footer


andrenatal:

kdavis-mozilla:
@andrenatal What browser? Or was this on the app? Could you add a picture?

andrenatal:
iOs

andrenatal:
this is gone.

PR #473, opened 2017-10-01
Adjust 'speak' page to avoid empty space

nevik:
On the "speak" page, the review box was taking up space in the DOM even in the recording stage. This pushed the footer bar further down that it should be (basically there was a big block of unused space).

This PR hides the review box and the record box whenever they're not in use, using `display: none`.

Edit: This seems to be the same issue that was reported in #185.

### DIFF
--- a/web/css/index.css
+++ b/web/css/index.css
@@ -1369,11 +1369,13 @@ body:not(.ios) .recording #background-container {
 }
 
 #voice-submit {
+  display: none;
   opacity: 0;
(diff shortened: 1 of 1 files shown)


nevik:
Whew, yeah I definitely overlooked that.

I'm going to re-examine this, and update this PR accordingly :)

nevik:
Since I didn't have time to re-visit this in time, and the re-design is coming up, i'm abandoning this PR

PR #474, opened 2017-10-01
Use async/await in AudioWeb class

nevik:
This replaces `Promise.then()` usage with async/await in the `AudioWeb` class, as part of #402.

@mikehenrty, please take a look at the remaining usages of `then()` in the web part of the code. I would actually leave them as they are, because they intentionally spawn off asynchronous actions and don't directly wait for them.
Introducing async/await in their place would pull up the Promise another call, and require changes in more places. In contrast to our goal with #402, this would actually degrade readability IMO.

### DIFF
--- a/web/src/lib/components/pages/record/audio-web.ts
+++ b/web/src/lib/components/pages/record/audio-web.ts
@@ -111,68 +111,71 @@ export default class AudioWeb {
     this.volumeCallback = cb;
   }
 
-  init() {
-    if (this.initPromise) {
-      return this.initPromise;
+  init(): Promise<void> {
+    if (!this.initPromise) {
+      this.initPromise = this.doInit();
     }
 
-    return (this.initPromise = this.getMicrophone()
-      .then((microphone: MediaStream) => {
-        this.microphone = microphone;
-        var audioContext = new AudioContext();
-        var sourceNode = audioContext.createMediaStreamSource(microphone);
-        var volumeNode = audioContext.createGain();
-        var analyzerNode = audioContext.createAnalyser();
-        var outputNode = audioContext.createMediaStreamDestination();
+    return this.initPromise;
+  }
 
-        // Make sure we're doing mono everywhere.
-        sourceNode.channelCount = 1;
-        volumeNode.channelCount = 1;
-        analyzerNode.channelCount = 1;
-        outputNode.channelCount = 1;
+  private async doInit(): Promise<void> {
+    try {
+      const microphone = await this.getMicrophone();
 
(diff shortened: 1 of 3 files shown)


nevik:
@mikehenrty I removed the cached Promise from `AudioWeb` now (and made `init` in `AudioIOS` also async).

As we had discussed on IRC, this leaves certain edge cases where the user can confuse the system. An obvious one is quickly tapping/clicking the record button twice, since that can cause a race condition between two separate instances trying to initialize the microphone.

However, unlike what I had previously said to you, I think this cached Promise is not the right way to combat this problem. Even with it in place, there are other race conditions possible in `record.tsx` when it comes to starting a recording.

We should tackle this separately, and prevent multiple starting of a recording in `record.tsx` directly. This kind of mutexing/locking is IMO not part of the responsibility of `AudioWeb` and `AudioIOS`, since they're not intended to be shared among multiple consumers.

web/src/lib/components/pages/record/audio-web.ts
```diff
@@ -111,68 +110,63 @@ export default class AudioWeb {
     this.volumeCallback = cb;
   }
 
-  init() {
-    if (this.initPromise) {
-      return this.initPromise;
+  async init(): Promise<void> {
```
mikehenrty: We still need a check here to see if we are already initialized. I agree we will still have an underlying race if the user clicks record twice really fast, and we should solve that in a different bug. But with this change, we initialize a new audio context every time the user presses record, which is unnecessary.

nevik:
@mikehenrty alright, cached Promise is now replaced with a boolean flag as you suggested :)

web/src/lib/components/pages/record/audio-web.ts
```diff
@@ -111,68 +111,67 @@ export default class AudioWeb {
...
+    if (this.isInitialized) {
```
mikehenrty: Where do we set `this.isInitialized` to true?

nevik: Ugh. Good thing you're paying attention!

I had overlooked that we already have that `ready` flag you mentioned previously, and then additionally failed to correctly implement the lifecycle for this new flag!

I've updated it to use the `ready` flag instead now, whose lifecycle should be fine.

mikehenrty:
Great!

nevik:
FINALLY 🎉

mikehenrty:
it only took 1 month! 📆

nevik:
The very image of fast and lean development!

PR #475, opened 2017-10-01
Do not reload page if user denies prompt to retry when mic permission is not granted

### LINKED ISSUE #233: Cancel should not refresh page on microhpone access dialog

nevik:
Based on #474.

Fixes #233.

When microphone access is not granted to the website, a dialog is shown when the user tries to start a recording. The user is asked for permission again, presented with the option to "Retry" or "Cancel".

"Retry" reloads the page, since newly granted permissions only apply after reloading the page.

Previously, "Cancel" would also reload the page, since the result of the dialog was not checked.

This change adjusts the logic involved in the "Record" button click handler and the initialization of the recorder (for web) to reload only if the user chooses so, and do nothing if the user cancels the dialog.


web/src/lib/components/pages/record/audio-web.ts
```diff
@@ -164,14 +173,19 @@ export default class AudioWeb {
...
+      return true;
     } catch (err) {
       if (err === ERROR_MSG.ERR_NO_MIC) {
         const shouldRetry = await confirm(
           'You must allow microphone access.',
           'Retry',
           'Cancel'
         );
-        window.location.reload();
+        this.initPromise = null;
```
nevik: Since the resulting Promise is cached in the AudioWeb instance, and the dialog is only shown when initialization fails, the Promise must be reset here.

Otherwise, the dialog would only show up once if the user clicks record, cancels the dialog, clicks record again (at this point nothing would happen, since the cached Promise is already resolved to false).

nevik:
Note: this also obsoletes #321

mikehenrty:
This looks great, but let's way until we finish #474 before moving forward here. I would love to get rid of this cached promise now that we have async/await coming to our audio classes.

mikehenrty:
I spoke with @nevik about this on IRC. Looks like the cached promise has it's usefulness, and we'll want to keep it. We'll still wait on the refactor in 474 before moving forward in this bug.

nevik:
Just rebased this onto updated #474, and adjusted for the changes made there.

nevik:
Oh, didn’t get around to updating this after the latest async await changes. Thanks for picking up the slack @mikehenrty!

PR #479, opened 2017-10-02
Use TS 'import' instead of 'require' where currently possible

nevik:
Since the Node typings are already part of package.json (though currently in version 7, we need to remember to change this if we decide to make Node 8 the minimum supported version), parts of the Node standard library can imported with `import` already, adding type support.

I'm planning to add more typings for other libraries used, but that will go in a separate PR :)

### DIFF
--- a/server/src/lib/api.ts
+++ b/server/src/lib/api.ts
@@ -1,9 +1,10 @@
+import * as fs from 'fs';
 import * as http from 'http';
+import * as path from 'path';
+
 import WebHook from './webhook';
 import respond from './responder';
 
-const path = require('path');
-const fs = require('fs');
 const Promise = require('bluebird');
 const Random = require('random-js');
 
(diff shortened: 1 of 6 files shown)


server/src/lib/api.ts
```diff
@@ -55,20 +56,14 @@ export default class API {
   private getFileContents(filepath: string) {
     return new Promise(
       (res: (contents: string) => void, rej: (error: any) => void) => {
-        fs.readFile(
-          filepath,
-          {
-            contents: 'utf8',
-          },
```
nevik: Once the typings were imported, TS complained about this parameter. Looking at the available signatures of `readFile`, I guessed that you currently want the one without specifying an encoding (which returns a Buffer).

When you do specify an encoding, `readFile` already gives you a string. We could also change that here.

Since I'm not sure about the effects of this change, and I can't properly test the backend, @mikehenrty please test this thoroughly :)

mikehenrty:
Overall, I like the direction of this, and I am going to make a small change to the readFile function and then merge.

@nevik can you tell me more about what we would need to change if we made node8 required?

in general, my feelings towards requirements is that we should support backwards compatibility as long as we can, so long as there's no security flaw or feature that makes upgrading necessary

nevik:
@mikehenrty from what I can tell, we don't need to change any code to switch to node 8.

At least, I don't see any obvious errors, and there are no compilation problems with the typings for node 8.

Since there are no automated tests (yet) and I don't have a fully functional setup (in particular, no AWS bucket) here, I can't do any more detailed functional testing.

PR #484, opened 2017-10-04
Improvements to the 'Open In App' bar on iOS

### LINKED ISSUE #467: "Open in app" on iOS should launch app if installed
_Noticed while implementing #466._

When viewing the CV website in Safari on iOS, an "Open in app" banner is shown at the bottom of the screen.

However, unlike its label, tapping the banner always opens the AppStore page of the CV app, even if the app is already installed:

https://github.com/mozilla/voice-web/blob/6e646414efe984f98cb39c398ed92bb56dc7ba35/web/src/lib/components/pages.tsx#L128-L130
https://github.com/mozilla/voice-web/blob/6e646414efe984f98cb39c398ed92bb56dc7ba35/web/src/lib/components/pages.tsx#L347-L349

This behavior should be adapted to directly launch the app if it is already installed.

This is usually done using a URL callback scheme (see http://x-callback-url.com).

Omniscimus:
Fixes #467.

iOS Safari users were redirected to the Common Voice app in the App Store with a green banner at the bottom of the screen. I replaced this with a so-called [Smart App Banner](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/PromotingAppswithAppBanners/PromotingAppswithAppBanners.html). This is a bar at the top of the screen which redirects users to the app's page in the App Store if the app is not installed yet, and it opens the app if it is.

Browsers on iOS other than Safari do not support Smart App Banners. The old bar is still in place for them. Clicking this banner previously always opened the App Store, regardless of whether the app was installed. Now, the app's page in the App Store will be opened if the app isn't installed, and the app will be opened if it is installed.

To open the app if it's installed, I added a custom link to the iOS app: `commonvoice://`. iOS will open any links starting with that prefix in the Common Voice app. Side-effect: people who have already installed the app but haven't updated it yet will be redirected to the app's page in the App Store if they click this link in their browser (where I suppose they can update it).

### DIFF
--- a/web/src/lib/components/pages.tsx
+++ b/web/src/lib/components/pages.tsx
@@ -114,8 +114,6 @@ export default class Pages extends Component<PagesProps, PagesState> {
     this.renderUser = this.renderUser.bind(this);
     this.linkNavigate = this.linkNavigate.bind(this);
     this.clearRobot = this.clearRobot.bind(this);
-    this.openInApp = this.openInApp.bind(this);
-    this.closeOpenInApp = this.closeOpenInApp.bind(this);
     this.onVolume = this.onVolume.bind(this);
   }
 
@@ -123,16 +121,6 @@ export default class Pages extends Component<PagesProps, PagesState> {
     if (!this.state.transitioning && this.state.recording) {
       this.setState({ recorderVolume: volume });
     }
-  }
-
-  private openInApp() {
-    window.location.href = getItunesURL();
-  }
-
-  private closeOpenInApp(evt: Event) {
-    evt.stopPropagation();
-    evt.preventDefault();
-    document.getElementById('install-app').classList.add('hide');
   }
 
   private getCurrentPageName() {
@@ -344,10 +332,6 @@ export default class Pages extends Component<PagesProps, PagesState> {
 
     return (
       <div id="main" className={className}>
-        <div onClick={this.openInApp} id="install-app">
-          Open in App
-          <a onClick={this.closeOpenInApp}>X</a>
-        </div>
         <header
           className={
             this.state.isMenuVisible || this.state.scrolled ? 'active' : ''
--- a/web/src/lib/components/pages.tsx
+++ b/web/src/lib/components/pages.tsx
@@ -125,8 +125,15 @@ export default class Pages extends Component<PagesProps, PagesState> {
     }
   }
 
+  /**
+   * If the iOS app is installed, open it. Otherwise, open the App Store.
+   */
   private openInApp() {
-    window.location.href = getItunesURL();
+    window.location.href = 'commonvoice://';
+
+    setTimeout(function() {
+      window.location.href = getItunesURL();
+    }, 500);
   }
 
   private closeOpenInApp(evt: Event) {
--- a/web/src/lib/components/pages.tsx
+++ b/web/src/lib/components/pages.tsx
@@ -114,6 +114,8 @@ export default class Pages extends Component<PagesProps, PagesState> {
     this.renderUser = this.renderUser.bind(this);
     this.linkNavigate = this.linkNavigate.bind(this);
     this.clearRobot = this.clearRobot.bind(this);
+    this.openInApp = this.openInApp.bind(this);
+    this.closeOpenInApp = this.closeOpenInApp.bind(this);
     this.onVolume = this.onVolume.bind(this);
   }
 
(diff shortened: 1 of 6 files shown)


web/src/lib/utility.ts
```diff
@@ -47,17 +47,35 @@ export function isFocus(): boolean {
...
+  return isWebkit && pretendsSafari && isSafari;
+}
+
+/**
  * Check whether the browser is mobile Safari (i.e. on iOS).
- * 
+ *
  * The logic is collected from answers to this SO question: https://stackoverflow.com/q/3007480
  */
 export function isMobileSafari(): boolean {
```
mikehenrty: This changes the meaning of this function slightly. Hindsight 20/20, this function should be called `isMobileWebkit.` We also use it to make buttons behave better on iOS browsers (thanks @nevik !). We still want that functionality for FF and Chrome on iOS, but this disables it. Probably best to leave this function mostly untouched, except for maybe swapping the `isIOS` function with the `isWebkit` function, since we would use the latter in two places.

web/src/lib/components/pages.tsx
```diff
@@ -125,8 +125,15 @@ export default class Pages extends Component<PagesProps, PagesState> {
     }
   }
 
+  /**
+   * If the iOS app is installed, open it. Otherwise, open the App Store.
+   */
   private openInApp() {
-    window.location.href = getItunesURL();
+    window.location.href = 'commonvoice://';
```
mikehenrty: This pops up an error message on iOS Firefox saying it doesn't understand the url. Is there a way we can feature detect for custom protocols? I googled it but didn't find anything. A backup would be to add a `Utility.isSafari` check here before setting this url.

web/src/lib/components/pages.tsx
```diff
@@ -344,10 +351,14 @@ export default class Pages extends Component<PagesProps, PagesState> {
 
     return (
       <div id="main" className={className}>
-        <div onClick={this.openInApp} id="install-app">
-          Open in App
-          <a onClick={this.closeOpenInApp}>X</a>
-        </div>
+        {isIOS() &&
+          !isSafari() && (
+            // In iOS Safari, we display a Smart App Banner instead.
```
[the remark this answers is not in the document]
mikehenrty: sorry about the duplicate comment, I thought the first one was erased. i'm deleting this one.

mikehenrty:
@andrenatal could you take a look at the iOS part of this one? it's about handling custom app protocols.

web/src/lib/components/pages.tsx
```diff
@@ -125,8 +125,15 @@ export default class Pages extends Component<PagesProps, PagesState> {
...
+  /**
+   * If the iOS app is installed, open it. Otherwise, open the App Store.
+   */
   private openInApp() {
-    window.location.href = getItunesURL();
+    window.location.href = 'commonvoice://';
+
+    setTimeout(function() {
+      window.location.href = getItunesURL();
+    }, 500);
```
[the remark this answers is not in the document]
Omniscimus: Most browsers suspend script execution when the page is not focused, so the timeout will stop running and continue if the user returns to the page. The custom link will just open the CV app. I'm not sure if it will always execute in time, but it would be surprising if it took more than 500 ms.

web/src/lib/components/pages.tsx
Omniscimus: You can't feature detect custom protocols, because you can't know in the browser which apps the user has installed. We need this to work in all browsers except Safari. Those browsers are also based on Safari on iOS. Apple doesn't like this kind of custom links because they want us to use [Universal Links](https://developer.apple.com/ios/universal-links/). Previously, a workaround for the popup was to put the custom link in an iframe, but they fixed that 'bug' as well, so now there's always such a popup if you use your own links.

We could maybe use Universal Links, but it involves generating one in an App Store developer account so I can't really do much about that. When using Universal Links, it's impossible for other websites to open your app with a similar link.

mikehenrty: Ah ok, good to know and thanks for the background. Overall the experience is nicer with your change than without it, since they should only see that error message once, and then once updated always go right to the app. Let's keep it.

web/src/lib/utility.ts
```diff
@@ -47,14 +47,41 @@ export function isFocus(): boolean {
...
+/**
  * Check whether the browser is mobile Safari (i.e. on iOS).
- * 
+ *
  * The logic is collected from answers to this SO question: https://stackoverflow.com/q/3007480
  */
-export function isMobileSafari(): boolean {
+export function isMobileWebkit(): boolean {
   const userAgent = window.navigator.userAgent;
   const isIOS = /(iPod|iPhone|iPad)/i.test(userAgent);
```
mikehenrty: let's use the `iOS()` function from above here too.

andrenatal:
LGTM with the suggestions @mikehenrty pointed

web/src/lib/components/pages.tsx
mikehenrty: For the time being, we're going to disable the custom protocol functionality on the site, but still register the custom protocol in the app. In the future, if we are assured more users have updated the app, we can add back in this improved redirect flow.

PR #490, opened 2017-10-07
Switch from npm to yarn

### LINKED ISSUE #477: Switch from npm to yarn to allow cross-platform sharing of lockfile
I had noticed a while back that on my Mac, doing an `npm install` (note: npm v5 and later) always modifies the `package-lock.json` that is committed to the repo.

When I looked into this more today, I found that this is an issue with how npm handles optional dependencies (in our case, that's [`fsevents`](https://www.npmjs.com/package/fsevents), a macOS-only dependency of things that watch filesystem changes):

Since this is an optional dependency which can only be installed on macOS, npm will install it and add it to package-lock.json whenever `npm install` is used on a Mac, but will remove it (if present) from package-lock.json whenever `npm install` is run on another platform.

From what I could find, there are currently at least two issues in the npm repo describing/discussing this: npm/npm#17722 and npm/npm#18135.

Since a fix for this doesn't seem to be in sight, and [there's a report that this isn't an issue with yarn](https://github.com/npm/npm/issues/17722#issuecomment-332450218), I suggest that we switch to [yarn](https://yarnpkg.com/) for package management.

@mikehenrty, thoughts?

nevik:
This switches the repo from npm v5 (using `package-lock.json`) to [yarn](https://yarnpkg.com/en/) (using `yarn.lock`).

Good news: [Travis supports yarn automatically](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Travis-CI-supports-yarn), so the `.travis.yml` doesn't even need to be updated.

When you switch an existing local clone from npm to yarn, it's generally advisable to delete the `node_modules` folder before installing from yarn (using `yarn` or `yarn install`).

Fixes #477.


Omniscimus:
I tested this on a Linux (Xubuntu) virtual machine. `yarn.lock` wasn't updated after running `yarn install`.

nevik:
Note: As of yarn v1.1.0 (and from what i can tell, since yarn v0.28), running `yarn [install]` on macOS with our `package.json` prints a large number of warnings like:

> The case-insensitive file .../common-voice/node_modules/fsevents/node_modules/ajv/lib/dot/v5/switch.jst shouldn't be copied twice in one bulk copy

This is a known bug in yarn: yarnpkg/yarn#3751.

**This does not break anything and can the warnings be entirely ignored.**

yarn maintainers have already submitted a "fix" that suppresses these messages, since they're not useful to the user. From what I can tell, this fix should be part of the next yarn release, which will likely happen soon.

mikehenrty:
Woot, this installed packages in half the time on my arch linux machine.

I just need to check with our deployment team on how we get yarn added to our infra. We'll merge this sometime next week I think. Thank you @nevik !

nevik:
haha oh right, I forgot to mention that yarn is also somewhat faster.

mikehenrty:
Closing this one in favor of #525.

PR #491, opened 2017-10-07
Remove bluebird dependency (use native ES6 Promise instead)

nevik:
As shortly discussed with @mikehenrty in #402, this removes the dependency on `bluebird`.

The server part now uses native ES6 Promises just like the web part does.

@mikehenrty, since I have relatively little time (in particular to make changes you request), I'm creating this as a separate small PR and didn't attack introducing async/await yet. I'll follow up with that in a new PR when I can, but if someone else gets around to starting it first, I don't want to block that :)

Also, feel free to make minor adjustments yourself as you see fit :)

### DIFF
--- a/server/src/lib/api.ts
+++ b/server/src/lib/api.ts
@@ -5,13 +5,12 @@ import * as path from 'path';
 import WebHook from './webhook';
 import respond from './responder';
 
-const Promise = require('bluebird');
 const Random = require('random-js');
 
 const SENTENCE_FOLDER = '../../data/';
 
 export default class API {
-  sentencesCache: String[];
+  sentencesCache: string[];
   webhook: WebHook;
   randomEngine: any;
 
@@ -95,7 +94,7 @@ export default class API {
     }
   }
 
-  getSentences() {
+  getSentences(): Promise<string[]> {
     if (this.sentencesCache) {
       return Promise.resolve(this.sentencesCache);
     }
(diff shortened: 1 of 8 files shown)


server/src/lib/files.ts
```diff
@@ -279,7 +278,7 @@ export default class Files {
   /**
    * Grab a random sentence and associated sound file path.
    */
-  getRandomClip(uid: string): Promise<string[2]> {
```
nevik: stating the length of arrays is not actually valid typescript AFAIK.

at the very least, it breaks type checking/type inference.

That said, this method would be better served returning an object (and the signature specified with an inline object type).

I didn't change that here because I didn't have time to look around and properly assess the usages of this method and an appropriate structure and naming of the returned information.

server/src/lib/api.ts
```diff
@@ -133,9 +132,11 @@ export default class API {
           sentences = sentences.concat.apply(sentences, sentenceArrays);
           console.log('sentences found', sentences.length);
           this.sentencesCache = sentences;
+          return this.sentencesCache;
         })
         .catch((err: any) => {
           console.error('could not retrieve sentences', err);
+          return [];
```
nevik: we now get better typechecking (`bluebird` was used without typings). this method actually returned a `Promise<void>` when the "actually do work" case happened.

I guess the consuming code dealt with this in some way, else we would have noticed problems. but now it is actually correct :)

note: returning the empty array in from the catch clause could be replaced by returning nothing and widening the type signature to `Promise<string[] | void>` (the consuming code then needs to check for `undefined`, but that's probably already the case)

server/src/lib/files.ts
mikehenrty: agreed, we need an interface definition for this, but no need to worry about that for now.

Issue #498, opened 2017-10-11
Update footer link color


m-branson:
Update footer links to match primary green `#b7d43f` UI color.

<img width="498" alt="screen shot 2017-10-10 at 5 02 34 pm" src="https://user-images.githubusercontent.com/18686986/31416302-04b3caae-addd-11e7-8f64-41a145ca5d0e.png">

heltmm:
I can take this.

mikehenrty:
Fixed by #505 .

PR #505, opened 2017-10-12
#498 Update footer link color 

heltmm:
Update footer links to match primary green #b7d43f UI color.

### DIFF
--- a/web/css/index.css
+++ b/web/css/index.css
@@ -530,7 +530,7 @@ header.active {
 }
 
 #moz-links .links a {
-  color: #b7d43f;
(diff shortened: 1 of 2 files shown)


web/css/index.css
```diff
@@ -530,7 +530,7 @@ header.active {
 }
 
 #moz-links .links a {
-  color: var(--color-help);
+  color: #b7d43f;
```
mikehenrty: We have a css variable for this called `--color-listen`, let's use that here instead of hard-coding the value here.

heltmm:
Thanks, changes have been implemented.

Issue #494, opened 2017-10-09
Build system should automatically import external libraries from node_modules


kenrick95:
Right now, I think the external libraries are manually added at `vendor/` folder and prepended into the final JS bundle during gulp build.

Because of this, I had quite some time troubleshooting how to integrate a new library into this project. (still not successful btw) Could the build system be improved such that external libraries are automatically imported from node_modules during building process?

I think something could be modified at the Typescript gulp build process to support this (because I happen to have a side project in TypeScript that could do this, e.g. https://github.com/kenrick95/c4/commit/84d717a4067b26d60d361718ee572bdc48dac25a is a small commit to add a Polyfill library to the app)

mikehenrty:
great idea! this way, we could also include things like minified or unminified sources based on a build flag, source maps, etc.

right now I think we have google-analytics, preact, and almond.js as our vendor libraries. if we could copy them over from node_modules rather than check them directly in, that would be ideal

nevik:
For the web part, webpack can do this (and is a currently very popular solution for it). I’m not sure if it works for our backend too, but it might.

I’ll see if I can investigate a bit this weekend, und less someone else already knows more or gets around to it before me

mikehenrty:
See proof-of-concept for switching to Webpack in #508.

PR #508, opened 2017-10-13
Webpack build PoC

nevik:
This is a proof of concept of how a [webpack](https://webpack.js.org)-based build would look (for both frontend and backend parts).

As such, the broad strokes will be done, and fine-adjustments of config options are left out.

----

**Notes:**
* documentation/explanatory comments are being kept relatively sparse. should be done more extensively
* this PoC shows how build scripting could be done exclusively with NPM scripts
  * this is somewhat orthogonal to the webpack build; 
  * gulp could still be used as a "dumb" scripting engine
  * gulp and npm-scripts have the advantage of being usable across platforms (vs. plain shell scripts)
* (web) uglification is enabled using default settings
* (web) to demonstrate usage of dev/prod environments with (partly) different settings, i created such environments for the web part
  * a common webpack config defines common properties
  * for dev builds, source maps are turned on (note: this isn't completely enabled; for proper source maps, TS compiler settings also need to be adapted)
  * for prod builds, code is uglified
* (server) server code is now also bundled into a single file (server.js)
  * as a side effect of this, all runtime-constructed paths (e.g. everything that uses `__dirname`) from within the `server/src/lib` folder now needs to use one less `../`, since it's running from `server/js`, not `server/js/lib` anymore


nevik:
@mikehenrty, alright, here's a proof of concept for a webpack build for frontend and backend, as we had shortly discussed.

As I said, the changes aren't that major / the build config isn't that complicated.

If you have any questions, I'll be happy to answer them :)

As i mention above, I set up the build to run mostly using npm-scripts now instead of gulp (except for the DB setup script used by the Travis build, because i was lazy), and this is just one way to go.
An alternative is to keep gulp for dumb scripting, and/or use a mix of npm scripts to kick things off and gulp to nicely manage things.

One disadvantage of the npm-scripts i see is that there are no variables, so certain folder paths appear several times (-> more effort & higher error probability if this ever needs to be changed).

mikehenrty:
Thank you for doing this investigation @nevik.

In the next 2 months, we will be making a lot of changes on the front-end (releasing data, adding stats pages, and profiles) and I want to hold off on considering changing our build system to after we ship those. It adds risk and overhead.

But I am interested in learning more. With the current system we've already solved a lot of things that webpack does automagically, but I know it's quite powerful on top of that. What are the biggest features you are looking for in advocating for the switch?

nevik:
I think webpack's biggest features are its focus on building & bundling, and its continued active development. The build speed of webpack and the bundle sizes it produces are being improved on all the time.

That said, however, I think in the grand scheme of thinks, neither gulp nor webpack are entirely better than the other. And since they use different paradigms (imperative for gulp vs. declarative for webpack), I think the choice often comes down to taste, as well.

One advantage of webpack's (more) declarative approach to build configuration is that you don't have to take as much care with the order of things. Task order/intra-build dependency management are handled by webpack itself. (Of course, this can sometimes make it more difficult to debug, or to enforce a certain order if it really matters.)

I think that in the end, a webpack-based build and a gulp-based build have about the same complexity for this project.
So it kind of comes down to how familiar the (current and expected future) maintainers and contributors are with the respective tool.

It's my impression that webpack is comparatively more popular than gulp with web project right now (but I have no data to judge how objective this impression is).
So it might be worth considering that a webpack-based build might (in some small way) make it easier for new contributors to help with build-related tasks. (On the other hand, those tasks are going to be comparatively rare, so the impact might be negligible either way.)

Gregoor:
Alrighty @nevik, I'd like this to get merged because:
- we could throw out the vendors folder and have dependencies on the web (also called #494)
- @mikehenrty: this fixes #592 
  - @nevik though we have to make some changes for that, e.g.:
    - use `cheap-module-eval-source-map` (which makes the build a tad slower)
    - disable the CSP header during development
- require CSS inside of JS, to reduce "distance" between declaration and usage of styles (discussed here #582)

Some things I'd like to improve about this PR though are:
- don't rerun server on web changes
- fewer npm scripts

I tried to address those myself, lmkwyt:
```
{
    "build": "npm-run-all --race --parallel build:server build:web",
    "build:server": "cd server/ && webpack --config webpack.config.ts",
    "build:web": "cd web/ && webpack --config webpack.prod.config.ts",
    "lint": "eslint **/*.ts",
    "prettier": "npm run lint -- --fix",
    "serve": "nodemon --watch server/js/server.js server/js/server.js",
    "start": "npm run lint && npm-run-all --race --parallel watch:server watch:web serve",
    "watch:server": "npm run build:server -- --watch",
    "watch:web": "cd web/ && webpack --config webpack.dev.config.ts --watch"
  }
```

nevik:
I’d love to move this forward, but unfortunately i won’t have any time for code work in the next ~2 weeks.

Since I set up this webpack config, some more tasks and scripts have been added and changed for the gulp script too, so those need to be added as well.

mikehenrty:
@nevik if you don't mind @Gregoor taking this over, I won't stand in the way 👍

nevik:
No that’s totally cool!

This should be rebased onto latest master anyway

Gregoor:
Closing this in favor of #615. Thankslots @nevik!

Issue #499, opened 2017-10-11
Update voice-submit headline/subhead


m-branson:
For `div id=voice-submit` update:

- [ ] 'Thank you!' headline to 'Review & Submit'
- [ ] 'Want to review your recording?' subhead to 'Thank you for recording! Now review and submit your clips below.'
- [ ] For now maintain styling for both.

<img width="336" alt="screen shot 2017-10-10 at 5 08 57 pm" src="https://user-images.githubusercontent.com/18686986/31416408-bb5a60ec-addd-11e7-93cc-68f38e248490.png">

MikeAlphaBravo:
Will be working on this.  Submit by Sunday October 15th.

margaret85:
Finished fix Friday 10/13 at 7pm PST.

PR #509, opened 2017-10-14
Michael Brooks and Margaret Berry fixed issue 499 

margaret85:
Update voice-submit headline/subhead.  Let us know if there were any problems with our pull.


nevik:
For reference: refers to #499

PR #515, opened 2017-10-14
Display error/waiting message while sentences are not/being loaded

### LINKED ISSUE #435: Show an error message when no sentences could be loaded from the server
Otherwise the user will see blank lines instead of sentences:
no text console

This issue has been forked from #360.

AlaaSarhan:
Closes #435 
Paired on it with @irmela 

We thought that it makes more sense to replace the record button while/if sentences are being/not loaded.

We faked the error by throwing an exception inside `getRandomSentences` in `src/lib/api.ts`.

### DIFF
--- a/web/src/lib/components/pages/record.tsx
+++ b/web/src/lib/components/pages/record.tsx
@@ -351,7 +351,7 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
 
   private async newSentenceSet() {
     // If we don't have enough sentences in our cache, fill it before continuing.
-    while (!this.areSentencesLoaded()) {
+    while (!this.areSentencesLoaded) {
       console.error('slow path for getting new sentences');
       await sleep(RETRY_TIMEOUT);
       await this.refillSentenceCache();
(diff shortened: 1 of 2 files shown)


web/src/lib/components/pages/record.tsx
```diff
@@ -364,6 +366,10 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
     }
   }
 
+  areSentencesLoaded() {
+    return this.sentenceCache.length >= SET_COUNT;
+  }
```
nevik: for derived properties like these, [typescript property accessors](https://www.typescriptlang.org/docs/handbook/classes.html#accessors) are nice IMHO.

```typescript
private get areSentencesLoaded() {
  return this.sentenceCache.length >= SET_COUNT;
}
```

usage:
```typescript
  while (!this.areSentencesLoaded) {
    ...
  }
```

To be clear: this has no advantage in terms of performance, security, type safety, or the likes. It's just a syntactical variant, which presents the derived property like a field when accessing it.
Therefore, using an accessor vs. a method here is mostly a matter of taste :)

Issue #285, opened 2017-07-18
Play / Pause button should not Pause


mbebenita:
While validating sentences I often get half way through the sentence and I want to restart it. The only option available now is to pause (and then resume) the playback which is not very useful. I would suggest simply changing the button to restart playback instead of pausing it.

kdavis-mozilla:
There is another option you did not mention, to wait a few seconds until the clip ends, then play it again.

mbebenita:
Waiting is what I'm trying to avoid. And even if I were to wait for playback to end, why would I ever need to pause the clip?

kdavis-mozilla:
If someone interrupts your listening, you would need to pause, then resume.

mikehenrty:
Yeah I can see your point @mbebenita, but I also see the point around pausing and restarting. I'm ambivalent as to which is the better approach. I'd love to hear other weigh in here.

kdavis-mozilla:
Yeah I don't know which is better either. However, there are arguments on both sides.

Omniscimus:
> If someone interrupts your listening, you would need to pause, then resume.

This is not really a use case for me. If this occurs then it's usually paused in the middle of a word so I have to listen to the recording again anyway, which is not a real pain because they are usually only a few seconds in length.

If the pause button would be changed to a :repeat: repeat icon then it might not be clear when the recording is playing.

How about just a :arrow_forward: play and a :white_square_button: stop icon? So it starts playing when you click play, it goes back to the start when you click stop, and the icon reverts to the play icon.

mikehenrty:
Ok, let's turn the pause button into a stop button, and make it restart the recording.

https://github.com/mozilla/voice-web/blob/7d6bb745801d31cb1ae5673caf3c7928cf44c4fc/web/src/lib/components/listen-box.tsx#L68

elahmo:
Working on this, opening PR shortly! :)

PR #516, opened 2017-10-14
Changed pause function and icon to stop in listen mode

elahmo:
As per discussion in #285, I have changed the pause button to have a stop function, and it looks like a stop button.


web/src/lib/components/icon.tsx
```diff
@@ -23,6 +24,7 @@ const ICONS: FontIcons = {
   redo: '',
   play: '',
   pause: '',
+  stop: '',
```
mikehenrty: I'm going to change this to the stop icon with a circle around it.

PR #517, opened 2017-10-14
Add TS typings for (most) server dependencies

nevik:
This further improves the server code by installing and using typings for third-party libraries, where available.

Some notes:
* no typings are available (directly or as separate packages) for the following dependencies:
  * `ff`
  * `proxy-agent`
  * `stream-transcoder`
  * `mediaserver`
* I wrote custom typings for the one function used from `mediaserver` (this need a minor adjustment of the gulp script too)
* `memorystream` was unused and therefore removed
* i replaced default/module imports with direct imports of the needed symbols, where possible and appropriate
* the `Postgres` class (in `server/src/lib/db`) seems to be unused. should it be removed?
* `SimpleGit` was previously initialized/instantiated using `new SimpleGit(PROJECT_PATH)`, even though it should be invoked as a function (not a constructor) according to its docs (and typings); i fixed this

As always, I can't fully test the server code, so please do some reasonably thorough testing of existing/needed functionality, @mikehenrty :)

### DIFF
--- a/server/src/lib/bucket.ts
+++ b/server/src/lib/bucket.ts
@@ -1,11 +1,12 @@
 import * as path from 'path';
 import * as Random from 'random-js';
+import { S3 } from 'aws-sdk';
 
 import { map } from '../promisify';
 import { getFileExt } from './utility';
+import './aws';
 
 const MemoryStream = require('memorystream');
-const AWS = require('./aws');
 
 const KEYS_PER_REQUEST = 1000; // Max is 1000.
 const LOAD_DELAY = 200;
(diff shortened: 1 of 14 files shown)


server/src/lib/bucket.ts
```diff
@@ -1,11 +1,10 @@
...
+import * as Random from 'random-js';
+import { S3 } from 'aws-sdk';
 
 import { map } from '../promisify';
 import { getFileExt } from './utility';
-
-const MemoryStream = require('memorystream');
-const Random = require('random-js');
-const AWS = require('./aws');
+import './aws';
```
mikehenrty: This is a little too bad. We have an anti-pattern where importing a file for the first time modifies the state of another global module. Making this adhere to TS standards makes the problem even more glaring.

I think the right thing to do is make our `awt.ts` module act as a proxy to the global AWS module (rather that just modifying it), and only expose s3 though that. But I'm not sure if we'll ever get to this.

PR #521, opened 2017-10-15
bundle server typescript compile

mikehenrty:


mikehenrty:
@nevik or @kenrick95, I'm having trouble bundling server code using the TS compiler. Can either of you take a look at this and tell me what I'm doing wrong?

kenrick95:
Hmm the bigger question should be: why bundle into one file?

mikehenrty:
> Hmm the bigger question should be: why bundle into one file?

Slightly faster, slightly cleaner, and I don't have to worry about duplicates js files for every ts file when doing fuzzy search working locally.

mikehenrty:
Let's close this for now as it's not worth the effort investigating right now.

PR #525, opened 2017-10-16
Nevik feature/switch to yarn, add yarn to deployment

mikehenrty:


mikehenrty:
@gozer could you take a look at my nubis changes? trying to use `yarn install` on the server.

mikehenrty:
Thinking about this more, I don't think we need to use yarn in deployment just yet. NPM should cover what we need there, and we'll use (and recommend) yarn for local development (for the lock file issue).

Issue #476, opened 2017-10-02
Introduce automated tests?


nevik:
There currently appear to be no automated tests in the backend nor the frontend part of CV.

Should we work to introduce them?

Or is this a deliberate decision?

mikehenrty:
Yes please! This was not a decision, just lack of bandwidth. Now is definitely the time to start discussing these things.

nevik:
that's great to hear!

do you have any experience/preference in terms of testing frameworks/libraries and assertion libraries? (personally, I have some experience with [Jasmine](https://jasmine.github.io/), [karma](https://karma-runner.github.io/1.0/index.html), [mocha](https://mochajs.org), and [chai](http://chaijs.com))

kenrick95:
To chime in, I have experience at work to write unit test using [Jest ](https://facebook.github.io/jest/) (+ [enzyme ](http://airbnb.io/enzyme/index.html) for rendering React components for testing)

mikehenrty:
Most of my testing experience is with mocha, chai, and sinon. I'm not married to any toolkit though.

Here are some test suites we need:
1. Test main server api calls (run the server, load `web/src/lib/api.ts` in a unit testing framework)
1. Test that `index.html` can be loaded
1. Test all the gulp commands

itsravenous:
I've got a lot of experience with building React components TDD, in both Mocha and Jest, and would love to help. There doesn't seem to have been any UI tests written yet - I'd be happy to help if this is still a priority?

Gregoor:
Happy to hear that, it is still something we very much want. I personally prefer Jest, as it imo has a better ouf of the box experience. Do you have a preference @itsravenous? And is there already a/some components you have in mind for testing?

mikehenrty:
We now have a place to start with the tests using Jest. I'm going to close this for now, and we can open individual bugs for new tests and enhancements to the framework.

PR #537, opened 2017-10-22
add tests to make sure server can connect to mysql, helps #476

mikehenrty:


mikehenrty:
@Gregoor could you take a look at this let us know if you have any concerns?

Gregoor:
@mikehenrty didn't really find anything to complain about. Also runs like a charme for me locally. One semi-related question though: Why not go with typeorm or a similar solution? Minimizing dependencies?

mikehenrty:
> Why not go with typeorm or a similar solution? Minimizing dependencies?

Good question. For now, I wanted a very thin DB layer that offered me prepared statements. Mysql2 seemed to be the most robust and active library I could find for this.

That said, we will need to do schema migrations several times over the coming months. If there are any ORMS which handle this nicely, I am open to exploring them. Here's what we would need:

* Ability to specify schemas for each version of the database
* Ability to specify SQL scripts to run for each upgrade.
* Ability for ORM to alter apprioriate tables in upgrade process give the previous schema and new schema.

I'm going to merge this for now, and we can continue the discussion about ORMs.

Issue #478, opened 2017-10-02
Split stylesheet into multiple files


nevik:
The CV web frontend's stylesheet, [web/css/index.css](https://github.com/mozilla/voice-web/commit/1650efa5be2660c8e4f44a329a0bd3e71fda8aa5), has grown rather large and unwieldy.

We should break it down into smaller files (e.g. component-related styles should go in files next to component TSX files), which are then joined during compilation.

This ticket is a discussion result from #470.

Edit: @mikehenrty, you can label this help-wanted, if you want ;)

mikehenrty:
One idea here would be to move the page component css into separate files.
https://github.com/mozilla/voice-web/tree/master/web/src/lib/components/pages

echonox:
Wouldn't be also nice if we used pre-processor like LESS?

nevik:
@echonox, the matter of preprocessors is discussed in the linked issue #470. 
To summarize: for this project, we prefer to use modern (plain) CSS to promote adoption of open standards

Gregoor:
Created a PR for it. While splitting up the CSS I discovered a couple of unused selectors. Haven't yet found a good CSS methodology that prevents that kind of problem (which'll turn into bloat and unmaintainable CSS when the project gets large). The best solution I know is [Styled Component](https://www.styled-components.com/), though that is radically different and not quite standard CSS (but standard JS) and would require a switch to Webpack.

nevik:
@mikehenrty i guess this is done with #542? ;)

mxstbr:
Just to note this here (styled-components co-creator), this isn't correct:

> Styled Components [...] would require a switch to Webpack.

We specifically built styled-components to work without any specific build tool out of the box, so it would not require a switch to webpack. (there is a [Babel plugin](https://github.com/styled-components/babel-plugin-styled-components) which improves the debugging experience, but even that is mostly optional) 

If y'all got any questions about styled-components just ping me! :blush: Would be happy to see Mozilla adopt it.

Gregoor:
Heya @mxstbr, nice to hear from you! I'm still interested in adopting it (ty for creating this great lib btw :) ), but atm there are enough other things on my plate that I don't think it would happen any time soon.

[...]

PR #542, opened 2017-10-23
Split up CSS into multiple files

Gregoor:
Issue #478 

- adds `gulp-cssimport` as a dev dependency
- moves CSS into `src/` and splits it up into multiple files, by page or component
- as most page-specific rules were already grouped, the imports are placed in the `index.css` at the location these rule-groups were before, to not change how it cascades


mikehenrty:
Great work @Gregoor! I love that we cascade building our stylesheet using imports rather than simply slurping up all the css files. This is definitely the way to go.

I wonder about moving all the css into the TypeScript source folders. To me, I like having the CSS in separate locations, even if the CSS structures mimics the TS structure somewhat. For instance, I'm not sure how much sense it makes to have the general CSS file located in the components sub folder. That said, I'm not familiar with idiomatic React, so maybe this is what people prefer? What are you thoughts on this decision?

@nevik or @kenrick95, if either of you have thoughts about this, I would love to hear them! We'll probably merge this in some form in the next day or 2.

Thanks!

Gregoor:
Thanks @mikehenrty!

Personally I prefer having files that I touch at similar times near to each other, which is why I went with that way of structuring it. But if all of you prefer it to be more separated, I could also move the styles back to its own dir with pages & components subdir.

I think the React community's position on this is even more extreme (which I'm also very much on board with, as I hinted at in #478), namely CSS-in-JS. Imo the key reasons for going down that route are:
- every dependency is managed in JS
  - no more dead selectors
  - explicit reference to styles that are used, where I'd consider selectors to be more indirect
- no cascade, unless you explictly want it

In the case of styled-components (and some other CSS-in-JS) libraries it still outputs regular ol' stylesheets. So there is not really a downside to it (apart that you give up on some CSS features).
If that route is also interesting to you, I could create a PR where I apply styled-component to a single small page, to hopefully convince you of the benefits.

mikehenrty:
Sounds interesting! Let's not go down the route of CSS directly in JSX components just yet, but I'm fine with heading that direction by moving CSS next to components. It looks like @kenrick95 agrees with you. Let's give @nevik a day or so to weigh in, and if not we'll merge this. In the meantime, let's keep the conversation open about CSS in JS.

kenrick95:
Personally, I like related things to be grouped together, i.e. if a style is being used in a component, I like this style and component to be located together. One nice thing about this is that at code editors, usually there will be a tree-style navigation and when editing a component, those related files will be directly seen together instead of having to traverse to somewhere far away to look for related files. I think for a small project, navigation wasn't a big issue, but if a project became big and has a deep hierarchical component, it could be even more troublesome to find a CSS file related to a component.

----

Anyway, from my experience, there are many ways of importing style in React, but from what I observe, there isn't any clear "winner" of which CSS import style out there as each import style have its own advantages and disadvantages over the others.

Gregoor:
@mikehenrty Fair enough, I'm fine with not rushing into it :)

@kenrick95 I agree, also on the winner-part. styled-components is probably more of a personal favorite. I'd be curious to hear what your prefered approach is.

nevik:
LGTM ;)

(I had a longer reply, but the network swallowed it and GitHub doesn't keep it in the form until when it submits. So the above is a short summary :D )

mikehenrty:
> (I had a longer reply, but the network swallowed it and GitHub doesn't keep it in the form until when it submits. So the above is a short summary :D )

I hate when that happens!

mikehenrty:
Thanks everyone for the great discussion. Common Voice is very lucky to have such thoughtful contributors.

Onwards and upwards!

Issue #483, opened 2017-10-03
Debug experience: Several Files, React Dev tools


jdittrich:
**Usecase:** I would like to step through the code and use the react dev tools. 

**Problem:** The code seems to be in one big file of 3000+ lines (at least with a source map, but still) and the react dev tools are not activated (though [they can work with preact, too](https://github.com/developit/preact/pull/339#issuecomment-252769393))

mikehenrty:
One thing we could do here is add a "Developer Mode" which would do things like:
* disable js bundling
* output source maps
* not minify css

Perhaps we could use a command like `gulp develop` or maybe even switch the gulp default to work this way (although I usually prefer defaulting to production).

jdittrich:
> Perhaps we could use a command like gulp develop…

Yes, that would be awesome and would have helped me to understand the project and where I could start contributing!

jdittrich:
hi @mikehenrty any progress with this or any way I can help? I tried to configure this with gulp on my own, but I was unsuccessful.

jdittrich:
Is that possibly solved already? I am not very experienced in TypeScript and Preact, but it seems I could now use the source maps for debugging the TypeScript of the components in the dev tools. If that is the case, it seems that `yarn start` already does what is needed.

mikehenrty:
Yeah, I think this is mostly working. We still need to figure out how to get good stacks #592, but that is tracked in a different bug. @jdittrich if you don't mind we could close this bug.

jdittrich:
Yes, please close, if you think it moves the general make-dev-easier

PR #545, opened 2017-10-24
Sourcemaps

Gregoor:
Issue #483

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -15,7 +15,7 @@ const DIR_UPLOAD = DIR_SERVER + 'upload/';
 const DIR_SERVER_JS = DIR_SERVER + 'js/';
 const DIR_DIST = DIR_CLIENT + 'dist/';
 const DIR_TOOLS = CWD + 'tools/';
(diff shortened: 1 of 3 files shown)


Gregoor:
@mikehenrty this should be done now. Needed to add a few more gulp tasks, but didn't really see a way around it.

nevik:
@Gregoor other than switching to webpack ;) (#508)

Gregoor:
@nevik I'm glad someone else said it :grin:

gulpfile.js
```diff
@@ -244,6 +262,10 @@ gulp.task('default', 'Running just `gulp`.', ['build'], () => {
   watchAndListen();
 });
 
+gulp.task('develop', 'Run everything in dev (with sourcemaps).', ['build-dev'], () => {
+  watchAndListen(true);
+});
+
```
mikehenrty: let's just make this the default for when we run `gulp`

Gregoor:
@mikehenrty I just tested it on both and for Chrome it worked without a hick. In FF it worked after a few reloads and now I can't get it to not work.

mikehenrty:
Great, let's do this!

Issue #543, opened 2017-10-24
Split up config and only access it one place


Gregoor:
- split up config into
  - user specific config
    - simple JSON file
    - gitignore it
    - add description in CONTRIBUTING.md
  - main config imports user config and merges it with defaults
- only access main config in one place and pass it through
  - https://github.com/mozilla/voice-web/search?utf8=%E2%9C%93&q=config.json&type=
  - facilitates testability

mikehenrty:
I'm going to take a stab at this one.

PR #547, opened 2017-10-24
remove local config file from tree, make server pass config options t…

### LINKED ISSUE #544: Only access `config.json` in one place, and pass it into the server
Right now, separate files access the config.json file, and that makes things like automated testing harder. Let's fix this!

This blocks #543

mikehenrty:
…o all sub objects, fixes #544


mikehenrty:
@Gregoor can you take a look? I meant to break down #543 into multiple parts, but this might actually address all of it. Thanks!

server/src/lib/model/db/mysql.ts
```diff
@@ -22,55 +21,45 @@ export type MysqlOptions = {
...
-        DEFAULTS.password
-      ),
-      host: getFirstDefined(options.host, config.MYSQLHOST, DEFAULTS.host),
-      port: getFirstDefined(options.port, config.MYSQLPORT, DEFAULTS.port),
-      max: getFirstDefined(options.max, DEFAULTS.max),
-      idleTimeoutMillis: getFirstDefined(
-        options.idleTimeoutMillis,
-        DEFAULTS.idleTimeoutMillis
-      ),
+      user: getFirstDefined(config.MYSQLUSER, DEFAULTS.user),
```
Gregoor: While I know JS truthiness is not to be trusted, for cases where I know that something falsy wouldn't be passed (like here), I'd personally go with: `config.MYSQLUSER || DEFAULTS.user`

Gregoor:
looks good to me 👍

server/src/lib/model/db/mysql.ts
mikehenrty: I agree with you in principal, but to me consistency is nice here since it's easier to maintain down the road. This particular piece of code has already been touched too many times due to JS truthiness problems, and each time I have to touch it I have to *think*, which nobody likes doing.

nevik: if only we could write some program to automatically _test_ if it's working correctly!

mikehenrty: @nevik *poof* now we do!

https://github.com/mozilla/voice-web/blob/2797248a1a4eac737cbdd9fbe8242b5049bc248a/server/src/test/schema-migration.ts#L11

#magic :)

PR #551, opened 2017-10-24
More dropshadow

erolaliyev:


Gregoor:
Thanks for the contribution! While I like your changes to the shadow, we're atm in the process of implementing a redesign. So I'll close this in favor of #553

PR #552, opened 2017-10-24
change recording display immediately after hitting record button, sin…

alemc28:
…ce the audio callback is slow in firefox


nevik:
@mikehenrty seems like this is an old change of yours -- is this still valid?

mikehenrty:
Hi @alemc28, thank you for your contribution! I believe we have already fixed this issue in the latest master. If you are still seeing a slowdown in firefox when hitting record, please open a new bug. Thank you!

Issue #546, opened 2017-10-24
Redesign text-box card


Gregoor:
- change dropshadow
- more space around the text

![record-text-box](https://user-images.githubusercontent.com/4051932/31959693-abf16bd2-b8f5-11e7-959d-e70d7f95350f.png)

iain8:
hi @Gregoor I have some time this afternoon to look at some of the record page things.

PR #553, opened 2017-10-24
updated box shadow and added space in box

iain8:
was limited in possible horizontal space by the page structure and didn't want to start fiddling with the center column size, if more space is needed we would have to look into that #546


Gregoor:
Awesome, thanks for the contribution! Looks right to me :)

Issue #548, opened 2017-10-24
Redesign record button


Gregoor:
- reorder button and text
- put index of recording underneath the button (and color it)
- recolor the button

![record](https://user-images.githubusercontent.com/4051932/31960613-9d3e36b2-b8f8-11e7-90cd-5daf12d93feb.jpg)

ReneVolution:
Is this still up for grab?

PR #554, opened 2017-10-24
Redesign record button

iain8:
added a little more vertical space as per the mockup #548


Gregoor:
Thanks again :) Changed the colors a bit, but rest was spot on 👌

iain8:
Glad to help! Yeah I noticed the color for the red had changed but it was the red used in a couple of places so I figured that was a separate change.

PR #568, opened 2017-10-28
Replace getElementById or queryselectors inside preact components by refs during rendering

### LINKED ISSUE #541: Replace getElementById/querySelector with React refs
There are a couple of components that use `document.getElementById` and `document.querySelector`:
https://github.com/mozilla/voice-web/search?utf8=%E2%9C%93&q=document.getelementbyid&type=

Replace those with [React's ref](https://reactjs.org/docs/refs-and-the-dom.html) so it's more React-idiomatic.

jcfausto:
References to document.getElementById and document.queryselector inside components were replaced by attributes that are set when rendering the components by using refs.

Only preact components were reviewed. The HTML files containing pure HTML elements not rendered by a preact component were not modified (i.e. voice-web/ios/voicebank-ios-wrapper/index-cv.html)

resolves #541

### DIFF
--- a/web/src/lib/components/pages.tsx
+++ b/web/src/lib/components/pages.tsx
@@ -79,6 +79,7 @@ export default class Pages extends Component<PagesProps, PagesState> {
   private scroller: HTMLElement;
   private content: HTMLElement;
   private bg: HTMLElement;
+  private installApp: HTMLElement;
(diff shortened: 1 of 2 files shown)


Gregoor:
Great work @jcfausto, thank you!

Issue #550, opened 2017-10-24
Profile Form underneath Record


Gregoor:
Depends on #549 

When "Create a Profile" is clicked, Profile Form should be shown:
![record-with-profile](https://user-images.githubusercontent.com/4051932/31960931-89e5a4c8-b8f9-11e7-87ae-c5f7753ebb5d.jpg)

Gregoor:
I'm going to take a stab at it 🍴

PR #570, opened 2017-10-28
Profile Refactor

Gregoor:
In preparation of #550, I went ahead and refactored profile a bit. It's 40 lines shorter now, just want to make sure I didn't sacrifice readability (imo I didn't but there's bound to be some subjectivity here). 

Possible points of contention:
- prefer object literal shorthands
- prefer `const` over `let`
- extracting state into local vars in render

Uncontroversial parts of this PR:
- directly access select values
- remove `Component.render()`-call
- extract options renderin into its own method

### DIFF
--- a/web/src/lib/components/pages/profile/profile.tsx
+++ b/web/src/lib/components/pages/profile/profile.tsx
@@ -64,7 +64,7 @@ export default class Profile extends Component<Props, State> {
     });
   }
 
-  private update() {
+  private update(): State {
     const data = {
       email: this.email.value,
       accent: this.profileAccent.value,
--- a/web/src/lib/components/pages/profile/profile.tsx
+++ b/web/src/lib/components/pages/profile/profile.tsx
@@ -22,12 +22,12 @@ export default class Profile extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
 
-    let user = this.props.user.getState();
+    const { email, accent, age, gender } = this.props.user.getState();
     this.state = {
-      email: user.email,
-      accent: user.accent,
-      age: user.age,
-      gender: user.gender,
+      email,
+      accent,
+      age,
+      gender,
     };
 
     this.saveEmail = this.saveEmail.bind(this);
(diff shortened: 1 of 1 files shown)


web/src/lib/components/pages/profile/profile.tsx
```diff
@@ -178,7 +132,7 @@ export default class Profile extends Component<Props, State> {
           ref={select => {
             this.profileAccent = select as HTMLSelectElement;
           }}>
-          {accentOptions}
+          {this.renderOptionsFor(ACCENTS, accent)}
         </select>
```
nevik: i think we should extract these entire `<select>`s (including their labels) into a component, to which we pass the data needed (id, tab index, isModified, options, currentOption).

that component could be local to this file, doesn't necessarily have to be in a separate file IMO

web/src/lib/components/pages/profile/profile.tsx
```diff
@@ -216,4 +170,12 @@ export default class Profile extends Component<Props, State> {
       </div>
     );
   }
+
+  private renderOptionsFor(options: any, selected: string) {
+    return Object.keys(options).map(key => (
```
nevik: :+1: for `map` instead of `forEach` with external accumulator :)

web/src/lib/components/pages/profile/profile.tsx
```diff
@@ -37,99 +37,53 @@ export default class Profile extends Component<Props, State> {
...
+      age: this.profileAge.value,
+      gender: this.profileGender.value,
+    };
+    this.setState(data);
+    return data;
   }
 
   render() {
-    let user = this.props.user.getState();
+    const { email, accent, age, gender } = this.state;
```
nevik: :+1: for the use of destructuring declarations!

web/src/lib/components/pages/profile/profile.tsx
```diff
@@ -37,99 +37,53 @@ export default class Profile extends Component<Props, State> {
...
-      gender: gender,
+      accent,
+      age,
+      gender,
     });
-
-    this.render();
   }
 
   private update() {
```
nevik: i'd love to see a return type declaration on this (as far as i can see, it would be `State`?)

web/src/lib/components/pages/profile/profile.tsx
Gregoor: The problem is that React <16 can't return arrays in render. Otherwise I'd be on your side on moving it to its own component.

nevik: ugh. that's annoying. though couldn't we just wrap them in another div then?

but i see that's not a trivial extraction then, so let's maybe attempt this in another PR

PR #572, opened 2017-10-29
Improve page loading UX by letting content fade in over spinner

nevik:
This improve page load UX by keeping the spinner in the DOM until them `#main` content is considered ready (images are pre-loaded or timeout occurred) and the fade-in transition is done.

It also replaces two `setTimeout` calls used for this with usage of our own Promise-based `sleep(ms)` utility function, improving code readability.

Short preview video: <a href="https://www.dropbox.com/s/bdq5h9zkfkr6arh/load-spinner.mov" target="_blank">on Dropbox</a>

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -17,6 +17,8 @@
     <link rel="stylesheet" href="/dist/index.css">
   </head>
   <body>
+    <div id="spinner"></div>
+    <div id="main"></div>
     <script src="/dist/bundle.js"></script>
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


mikehenrty:
I like the improved readability and UX. The only drawback to this approach is now we mix using react render functionality and modifying the DOM directly in `app.ts`. But, it's not that big a deal, and I have seen a bunch of React apps have special, non-Reactish rendering logic in the top-level component. Let's merge this for now, and rethink this when we have to touch this code again.

nevik:
Yeah this shows i don't have that much React experience, so this is more of a "the only way i knew how" approach.

With more input from @Gregoor i'm sure we can improve this in the future.

PR #574, opened 2017-10-30
Keep sentence in same position after re-recording

### LINKED ISSUE #526: [Bug] When re-recording, the re-recorded sentence moves to the bottom on submit screen
Steps to reproduce:
1. Go to record/speak page
2. Record 3 clips
3. On submit confirmation screen, click the re-record button for the **first** clip
4. Re-record the clip

Expected result:
Submit screen shows the clips in the original order.

Actual result:
The re-recorded clip gets moved to the bottom of the screen.

anbenson:
Fixes #526 

Let me know if I need to rebase this by the time you get a chance to look at this.

### DIFF
--- a/web/src/lib/components/pages/record/record.tsx
+++ b/web/src/lib/components/pages/record/record.tsx
@@ -52,6 +52,7 @@ interface RecordState {
   uploading: boolean;
   uploadProgress: number;
   isReRecord: boolean;
+  reRecordIndex: number;
   whyProfileVisible: boolean;
 }
 
@@ -72,6 +73,7 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
     uploading: false,
     uploadProgress: 0,
     isReRecord: false,
+    reRecordIndex: -1,
     whyProfileVisible: false,
   };
 
@@ -130,12 +132,17 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
 
   private processRecording(info: AudioInfo) {
     let recordings = this.state.recordings;
-    recordings.push(info);
+    if (this.state.isReRecord) {
+      recordings.splice(this.state.reRecordIndex, 0, info);
+    } else {
+      recordings.push(info);
+    }
 
     this.setState({
       recordings: recordings,
       recording: false,
       isReRecord: false,
+      reRecordIndex: -1,
     });
 
     this.tracker.trackRecord();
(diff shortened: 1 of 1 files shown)


web/src/lib/components/pages/record/record.tsx
```diff
@@ -406,11 +407,15 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
     for (let i = 0; i < SET_COUNT; i++) {
       // For the sentences elements, we need to
       // figure out where each item is positioned.
+      // When rerecording, this is determined by reRecordIndex.
+      // Otherwise, this is determined by the number of sentences recorded.
       let className = 'text-box';
       let length = this.state.recordings.length;
-      if (i < length) {
+      let reRecordIndex = this.state.reRecordIndex;
+      let recordIndex = this.state.isReRecord ? reRecordIndex : length;
```
nevik: please use `const` (instead of `let`) where possible :)

nevik:
except for the minor change to prefer `const` where possible, this LGTM.

@mikehenrty i think that `record.tsx` needs some considerable refactoring though, overall. let's talk about this separately and create issues as needed.

web/src/lib/components/pages/record/record.tsx
```diff
@@ -72,6 +73,7 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
     uploading: false,
     uploadProgress: 0,
     isReRecord: false,
+    reRecordIndex: 0,
```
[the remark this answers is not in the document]
anbenson: Sounds like a good idea. JavaScript / TypeScript isn't my main language - which is usually preferred?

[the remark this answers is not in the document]
nevik: in this case, i'd go with `null` or `-1` to show that it has a set value (vs `undefined`, which usually means "this thing hasn't been assigned"). among these, it's really a point of taste, so you can pick what you prefer :)

web/src/lib/components/pages/record/record.tsx
anbenson: Sure thing. When I fix this, should I stack a new commit on top of this one, or force-push an update to this commit?

nevik: whichever you prefer, we don't have a strict policy on this :)

anbenson:
Updated!

web/src/lib/components/pages/record/record.tsx
```diff
@@ -130,7 +132,11 @@ export default class RecordPage extends Component<RecordProps, RecordState> {
...
   private processRecording(info: AudioInfo) {
     let recordings = this.state.recordings;
-    recordings.push(info);
+    if (this.state.isReRecord) {
+      recordings.splice(this.state.reRecordIndex, 0, info);
+    } else {
+      recordings.push(info);
+    }
 
     this.setState({
```
nevik: in this `setState`, where `isReRecord` is reset to `false`, `reRecordIndex` should also be reset to its default (-1).

other than that, LGTM :)

@Gregoor @mikehenrty please check as well ;)

anbenson:
Updated!

Gregoor:
Thanks @anbenson!

LGTMSIMI

PR #577, opened 2017-10-30
Clean up web/api.ts

nevik:
Some clean-up and de-duplication of the API service in the web part.

### DIFF
--- a/web/src/lib/api.ts
+++ b/web/src/lib/api.ts
@@ -7,10 +7,14 @@ export interface Clip {
   sound: string;
 }
 
+interface HeadersMap {
+  [headerName: string]: string;
+}
+
 interface FetchOptions {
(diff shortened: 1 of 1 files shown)


Gregoor:
LGTM :+1: Any idea which browser we're supporting is not supporting fetch? Afaik every browser that supports `getUserMedia` (which we require) also does `fetch`.

nevik:
I don't know, this was already there from the start.

@mikehenrty might know. If he doesn't remember, I think we should double-check this and remove it if it's no longer needed

mikehenrty:
Even if a browser does not support `getUserMedia` and so cannot record sentences, we still want them to be able to listen to other's sentences, which requires an ajax request. So best to for now support browsers like IE for listening, which is why we don't use `fetch` yet.

PR #578, opened 2017-10-30
show profile form underneath record

Gregoor:
Issue #550

### DIFF
--- a/web/src/lib/components/pages/record/profile-actions.tsx
+++ b/web/src/lib/components/pages/record/profile-actions.tsx
@@ -2,15 +2,13 @@ import { h, Component } from 'preact';
 import Profile from '../profile/profile';
 import User from '../../../user';
 
-interface WhyProfileProps {}
-
 interface WhyProfileState {
   expanded: boolean;
 }
 
 const WHY_PROFILE_TITLE = 'Why a profile?';
 
-class WhyProfile extends Component<WhyProfileProps, WhyProfileState> {
+class WhyProfile extends Component<{}, WhyProfileState> {
   state = { expanded: false };
 
   private toggle = () => {
--- a/web/src/lib/components/pages/record/profile-actions.tsx
+++ b/web/src/lib/components/pages/record/profile-actions.tsx
@@ -2,6 +2,27 @@ import { h, Component } from 'preact';
 import Profile from '../profile/profile';
 import User from '../../../user';
 
+interface WhyProfileProps {
+  onClose: () => any;
+}
+
+const WhyProfile = ({ onClose }: WhyProfileProps) => (
+  <div>
+    <span id="why-profile-title">Why a profile?</span>
+    <div id="why-profile">
+      <p id="why-profile-text">
+        Copy explaining value of profile &amp; demographic capture: Lorem ipsum
+        dolor sit amet, consectetur adipiscing elit. Nulla id orci dui.
+      </p>
+      <p>
+        <a name="" onClick={onClose}>
+          Close
+        </a>
+      </p>
+    </div>
+  </div>
+);
+
 interface Props {
   user: User;
 }
@@ -56,29 +77,13 @@ export default class ProfileActions extends Component<Props, State> {
           Create a profile
         </button>
       ),
-      <div>
-        {whyProfileVisible ? (
-          <div>
-            <span id="why-profile-title">Why a profile?</span>
-            <div id="why-profile">
-              <p id="why-profile-text">
-                Copy explaining value of profile &amp; demographic capture:
-                Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla
-                id orci dui.
-              </p>
-              <p>
-                <a name="" onClick={this.toggleWhyProfile}>
-                  Close
-                </a>
-              </p>
-            </div>
-          </div>
-        ) : (
-          <span id="why-profile-title">
-            <a onClick={this.toggleWhyProfile}>Why a profile?</a>
-          </span>
-        )}
-      </div>,
+      whyProfileVisible ? (
+        <WhyProfile onClose={this.toggleWhyProfile} />
+      ) : (
+        <div id="why-profile-title">
+          <a onClick={this.toggleWhyProfile}>Why a profile?</a>
+        </div>
+      ),
     ];
   }
 }
--- a/web/src/lib/components/pages/record/profile-actions.tsx
+++ b/web/src/lib/components/pages/record/profile-actions.tsx
@@ -2,26 +2,45 @@ import { h, Component } from 'preact';
 import Profile from '../profile/profile';
 import User from '../../../user';
 
-interface WhyProfileProps {
-  onClose: () => any;
+interface WhyProfileProps {}
+
+interface WhyProfileState {
+  expanded: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


web/src/lib/components/pages/record/profile-actions.tsx
```diff
@@ -0,0 +1,84 @@
...
+    whyProfileVisible: false,
+  };
+
+  private toggleProfileForm = () => {
+    this.setState({ profileFormVisible: !this.state.profileFormVisible });
+  };
+
+  private toggleWhyProfile = () => {
+    this.setState({ whyProfileVisible: !this.state.whyProfileVisible });
+  };
```
nevik: @mikehenrty has been using the approach of declaring instance methods, then `.bind()`ing them in the constructor, instead of assigning lambdas to fields.

i dont much prefer either variant, but IMO we should try to be consistent here. please talk to @mikehenrty about this :)

web/src/lib/components/pages/record/profile-actions.tsx
```diff
@@ -0,0 +1,84 @@
...
+              </p>
+            </div>
+          </div>
+        ) : (
+          <span id="why-profile-title">
+            <a onClick={this.toggleWhyProfile}>Why a profile?</a>
+          </span>
+        )}
+      </div>,
+    ];
```
nevik: why return an array with one element?

web/src/lib/components/pages/record/profile-actions.tsx
```diff
@@ -0,0 +1,84 @@
...
+      !profileFormVisible && (
+        <button
+          type="button"
+          id="create-profile-button"
+          onClick={this.toggleProfileForm}>
+          Create a profile
+        </button>
+      ),
+      <div>
+        {whyProfileVisible ? (
```
nevik: instead of switching at the top level of the why-profile box, i think we should use a common element structure for it, and switch visibility with CSS based on a class (or so).

but we can do that in a separate PR after this one

web/src/lib/components/pages/record/record.css
```diff
@@ -252,6 +252,17 @@
     border: solid 0.1rem;
 }
 
+#profile-form-container {
```
nevik: the CSS should go to a separate file IMO

web/src/lib/components/pages/record/profile-actions.tsx
Gregoor: Thanks, good point. I prefer class properties, since they're less noisy. But I'll hash it out with @mikehenrty.

web/src/lib/components/pages/record/profile-actions.tsx
Gregoor: It's two elements: the create-button and the why text/toggle

nevik: well, that has very poor readability then :(

can you arrange this in a nicer way?

web/src/lib/components/pages/record/profile-actions.tsx
Gregoor: Right, I've seen that kind of CSS switching in a couple of places. Imo that's not the React way and might also incur some perf cost. I probably shouldn't make perf statements without measuring, but my reasoning would be: By always rendering the whole tree, you make React's diffs more expensive (though I guess your counter would be that it's a tradeoff for less DOM manipulation).

web/src/lib/components/pages/record/profile-actions.tsx
Gregoor: tried to make it more readable by factoring the WhyText into a separate comp. LMKWYT

web/src/lib/components/pages/record/record.css
Gregoor: done

web/src/lib/components/pages/record/profile-actions.tsx
nevik: definitely better.

i'd still prefer to keep the expanded and collapsed variants together, e.g. like this:

```
      <div>
        <span id="why-profile-title">
          {expanded ? (
            <a onClick={this.toggleExpanded}>Why a profile?</a>
          ) : (
            <span>Why a profile?</span>
          )}
        </span>
        {expanded && (
          <div id="why-profile">
            <p id="why-profile-text">
              Copy explaining value of profile &amp; demographic capture:
              Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla
              id orci dui.
            </p>
            <p>
              <a onClick={this.toggleExpanded}>Close</a>
            </p>
          </div>
        )}
      </div>
```

where `expanded` is a different name for `whyProfileVisible`.

Ideally, we'd also deduplicate the "Why a Profile?" text, but i've not tried that yet. The point of this is that if someone later comes and changes the copy, structure or style of the expanded or collapsed variant, they should be able to easily find the other variant as well.

The example above is also a non-CSS-visibility-switching solution, as you said you prefer in another comment (and the reasoning of which i can see).

Gregoor: Makes sense to me 👍 Moved it into its own component (naming is a bit shortened, not out of disrepect but because of me not checking the comment again while making the change).

web/src/lib/components/pages/record/profile-actions.tsx
```diff
@@ -2,55 +2,69 @@ import { h, Component } from 'preact';
 import Profile from '../profile/profile';
 import User from '../../../user';
 
-interface WhyProfileProps {
-  onClose: () => any;
+interface WhyProfileProps {}
```
nevik: minor: since Props are no longer needed for this, you can specify `extends Component<void, WhyProfileState>` instead of using an empty interface :)

other than that: LGTM

Gregoor: While I would like that better as well, that yields the following error for me:

`Type '{}' is not assignable to type 'void & ComponentProps<WhyProfile>'.
  Type '{}' is not assignable to type 'void'.`

Gregoor: aaah, works with `class WhyProfile extends Component<{}, WhyProfileState>`

nevik: whaaat. that's weird D:

i'll try to look into this (void worked for me in a similar situation)

nevik: okay, for my own curiosity and sanity, i looked in to this.

the problem occurs at the usage site of `<WhyProfile />`, because the JSX transform invokes the component creator with an empty object in this case. For `T extends Component<P, S>`, this props object must be of type `P & ComponentProps<T>` (an [intersection type](https://www.typescriptlang.org/docs/handbook/advanced-types.html#intersection-types)). if `P === void`, the resulting `void & ComponentProps === void` (nothing is an instance of `void`, so nothing can be an instance of an even more specific type).

Besides using the empty object type `{}`, as you figured out, using [the types for `null` or `undefined`](https://www.typescriptlang.org/docs/handbook/basic-types.html#null-and-undefined) would also work, since all three of them are neutral in regards to type intersection. (that's actually kinda weird, but has practical no bearing on our situation).

Curiosity satisfied!

Gregoor: I marvel at your curiosity. Thanks for the insight!

Issue #567, opened 2017-10-27
Submit Success "alert"


Gregoor:
Create a closable alert component where one can supply a text as a child. We're going to reuse that one!

And then we need to show it after recordings where submitted (as seen below). Probably best if the alert is also closed, when the user clicks record.

![submit-success](https://user-images.githubusercontent.com/4051932/32110684-6d70e4e2-bb38-11e7-8980-d440634110f6.jpg)

aerohitk:
@Gregoor I would like to take a shot at this, if that's alright?

Gregoor:
@aerohitk that's more than alright :)

nevik:
@Gregoor i'd prefer to name this element a "notification banner" or similar. For me, an "alert" (at least in a web context) refers to something like a popup or modal dialog - from what I can tell, the element we want here is not as aggressive as that.

Gregoor:
@nevik while I agree with you about the literal meaning of an alert, this is the naming bootstrap & foundation went with for such a component. So, it's hard to say what's the web standard here.
Notification is imo a bit off as well, since we do have real notifications on the web now.

nevik:
fair enough, then let's keep the nomenclature that way :)

Issue #579, opened 2017-10-30
Save Profile Success Alert


Gregoor:
When alerts are implemented (#567), we'll also need to add one for profile creation (inside of the record page):

![Save Alert](https://screenshots.firefoxusercontent.com/images/a7221603-87c7-4715-8b73-7eb0e2530f87.png)

Gregoor:
Hi @aerohitk, are you still interested in looking into this? If not, I would take it over tomorrow or beginning of next week.

aerohitk:
@Gregoor sorry, I've been busy with work recently. I'm definitely still interested, but I probably won't have time to work on this until the weekend. Let me know if this can wait that long.

Gregoor:
@aerohitk no worries, just wanted to have the redesign done next week but I've got plenty of other stuff to do, so it can wait this long.

aerohitk:
@Gregoor submitted a PR. Is there anything else I can help with?

Gregoor:
Thanks @aerohitk, will review it first thing tomorrow.

#625 is definitely something we want to get done this week, so if you want to look it in one of the next days, feel free to grab it.

PR #581, opened 2017-10-30
Add alert component to record page

aerohitk:
Issue #567

### DIFF
--- a/web/src/lib/components/alert/alert.tsx
+++ b/web/src/lib/components/alert/alert.tsx
@@ -1,0 +1,49 @@
+import { h, Component } from 'preact';
+import Icon from '../icon';
+
+interface Props {
+  text: string;
+  active: boolean;
+}
+
+interface State {
+  active: boolean;
+}
+
+export default class Alert extends Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    this.onClick = this.onClick.bind(this);
+  }
+
+  state: State = {
+    active: false,
+  };
+
+  componentWillReceiveProps(nextProps: Props) {
+    if (nextProps.active !== this.props.active) {
+      this.setState({
+        active: nextProps.active,
+      });
+    }
+  }
+
+  private onClick() {
+    this.setState({
+      active: false,
+    });
+  }
+
+  render() {
+    const className = 'alert ' + (this.state.active ? 'active' : '');
+
+    return (
+      <div className={className}>
+        {this.props.text}
+
+        <Icon type="x" onClick={this.onClick} className="icon" />
+      </div>
+    );
+  }
+}
--- a/web/src/lib/components/alert/alert.tsx
+++ b/web/src/lib/components/alert/alert.tsx
@@ -18,13 +18,13 @@ export default class Alert extends Component<Props, State> {
   }
 
   state: State = {
-    active: false,
+    active: this.props.active,
   };
 
-  componentWillReceiveProps(nextProps: Props) {
-    if (nextProps.active !== this.props.active) {
+  componentWillReceiveProps({ active }: Props) {
+    if (active !== this.state.active) {
       this.setState({
-        active: nextProps.active,
+        active,
       });
     }
   }
--- a/web/src/lib/components/alert/alert.tsx
+++ b/web/src/lib/components/alert/alert.tsx
@@ -36,10 +36,12 @@ export default class Alert extends Component<Props, State> {
   }
 
   render() {
-    const className = 'alert ' + (this.state.active ? 'active' : '');
+    if (!this.state.active) {
+      return null;
+    }
 
     return (
-      <div className={className}>
+      <div className="alert">
         {this.props.text}
 
         <Icon type="x" onClick={this.onClick} className="icon" />
--- a/web/src/lib/components/alert/alert.tsx
+++ b/web/src/lib/components/alert/alert.tsx
@@ -4,39 +4,39 @@ import Icon from '../icon';
 interface Props {
   text: string;
   active: boolean;
+  autoHide: boolean;
+  hideAlert: Function;
 }
 
-interface State {
-  active: boolean;
-}
+export default class Alert extends Component<Props, void> {
+  timeout: number;
 
-export default class Alert extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
 
     this.onClick = this.onClick.bind(this);
   }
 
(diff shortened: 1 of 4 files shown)


web/src/lib/components/alert/alert.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export default class Alert extends Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    this.onClick = this.onClick.bind(this);
+  }
+
+  state: State = {
+    active: false,
```
nevik: this should be initialized with the given `active` value passed to the constructor :)

web/src/lib/components/alert/alert.tsx
```diff
@@ -0,0 +1,49 @@
...
+    if (nextProps.active !== this.props.active) {
+      this.setState({
+        active: nextProps.active,
+      });
+    }
+  }
+
+  private onClick() {
+    this.setState({
+      active: false,
```
nevik: hmmm... this way, the parent component doesn't get notified when user clicks the banner to hide it.

I guess we don't need that for now (or the immediately foreseeable future), so that's fine for now :)

web/src/lib/components/alert/alert.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+    this.onClick = this.onClick.bind(this);
+  }
+
+  state: State = {
+    active: false,
+  };
+
+  componentWillReceiveProps(nextProps: Props) {
+    if (nextProps.active !== this.props.active) {
```
nevik: by comparing with `this.props` instead of `this.state`, the following behavior can happen:

1. parent component sets banner to visible (`props.active == true`, `state.active == true`)
1. user clicks banner to dismiss it (`props.active == true`, `state.active == false`)
1. parent component re-sets the props of the alert component to `true`
* -> line 25 does not trigger `setState`, banner does not become visible.

this seems like a somewhat unlikely scenario, but if it does happen, it will be difficult to track down.

-> instead, compare `nextProps` to `this.state`, since that reflects the thing you (might) want to change :)

web/src/lib/components/alert/alert.tsx
```diff
@@ -0,0 +1,49 @@
...
+    super(props);
+
+    this.onClick = this.onClick.bind(this);
+  }
+
+  state: State = {
+    active: false,
+  };
+
+  componentWillReceiveProps(nextProps: Props) {
```
nevik: you can destructure `nextProps` like so:

```
componentWillReceiveProps({ active }: Props) {
  // you can use "active" here instead of "nextProps"
}
```

-> slightly less verbose :)

you can look at the changes in #578 for more examples of this style

web/src/lib/components/alert/alert.tsx
```diff
@@ -0,0 +1,49 @@
...
+  render() {
+    const className = 'alert ' + (this.state.active ? 'active' : '');
+
+    return (
+      <div className={className}>
+        {this.props.text}
+
+        <Icon type="x" onClick={this.onClick} className="icon" />
+      </div>
+    );
```
nevik: @Gregoor so here the banner visibility is being switched by CSS, instead of not rendering it. Personally i'm used to doing it this way (i.e. with CSS).

Would you suggest not emitting any DOM nodes at all instead of emitting them and making them invisible? If so, is that a (P)React convention?

aerohitk:
@nevik I made the suggested changes. Let me know how that looks.

web/src/lib/components/alert/alert.tsx
aerohitk: It would be easy enough for me to change that to a conditional render, if that's what you want to do.

Gregoor: Thanks @aerohitk, the component is looking good!

@nevik Regarding not emitting any html nodes: Yes, I'd still stand by that. Though before React 16 it was not possible to `return null` in `render()`, so not emitting any wasn't really possible. But in this case I'd indeed only return a `<div/>`, when it is `!active`.
Do you want to make that change @aerohitk?

Gregoor:
@nevik Yes, I agree. An `autoHide: boolean` prop would be good as well. With a default time of 5 seconds?

aerohitk:
@Gregoor I added the conditional render and the autoHide functionality. I also did a bit of refactoring to remove state from the alert component, because the timeout was creating a conflict between state and props.

Gregoor:
Awesome, thanks for the quick reaction! Got two more nitpicks:
- the `hideAlert` prop should be called `onClose`. Typically in React land a child component doesn't tell the parent what to do, but rather lets the parent listen to events (like close) and handle it however it likes. Though I'll admit that in this case the line is a little blurry.
- clear the timeout on component unmount

aerohitk:
@Gregoor added the fixes. I was thinking of adding some kind of fade in / fade out effect to the alert. Is that something you would want? Also, would you mind if I worked on #579 in this pr?

Gregoor:
I like the idea of having fade animations👍🏻 and #579 is a good for this PR. Thanks!

Gregoor:
@aerohitk I went ahead and merged this one, so we can already use it. But if you're still willing, feel free to open a new PR for #579.

You can see the adjustments I made to it here: https://github.com/mozilla/voice-web/commit/c1627a758db2a6cd60fdb68eef96d5c599a5f337
The one that might benefit from explanation is why I moved the timer logic from `componentWillUpdate` to `constructor` & `componentWillReceiveProps`. For one, CWU is used for preparation of things that directly affect render. Also when optimizing components by minimizing unnecessary renders `shouldComponentUpdate` is implemented and a reasonable implemention for `Alert` might check if only `autoHide` has changed and if so prevent the update. In that case CWU would not be called, even though autoHide has changed. 

Let me know if any other change is unclear or seems wrong to you. Thanks for you help!

aerohitk:
@Gregoor just had time to look at your changes. Only one comment. On this line.

https://github.com/mozilla/voice-web/commit/c1627a758db2a6cd60fdb68eef96d5c599a5f337#diff-d79c443b6e2cac7b1a8cc5ad50042bc9L150

It should be `alertVisible: !isFull` shouldn't it?

Gregoor:
@aerohitk that way the alert was visible after every recording. Instead we want it to be visible when the three recordings were submitted and the user is returned to the sentence record view.

PR #580, opened 2017-10-30
Fix #562: Modify apache config to compress js, css, and svg files.

Jiggmin:
Add settings to Apache's configuration file to compress js, css, and svg files. This fixes #562.

It looks like your site serves javascript with the "application/javascript" content type, but I left the "text/javascript" content type in there just to be on the safe side.

PR #588, opened 2017-11-01
Output all static files to one folder, and serve them from there

### LINKED ISSUE #371: [static] Move all static content under a single url prefix
Something like /s/ or /static/ or /assets/ for instance

That way, managing it outside of node itself becomes a lot easier

nevik:
This should fix #371.

With this, the build outputs all static assets needed for the frontend into the `/web/dist` folder.
The backend is also adjusted to serve the static files from this folder instead.

I also adjusted the puppet file that seemed to be the appropriate place.

The local server backend / dev environment seems to work fine with this change, but I can't test whether the puppet file correctly sets up the apache. So I hope you can check (and, if necessary, fix) that for me, @mikehenrty and @Gregoor.

Edit: I think this shouldn't interfere with #580.

Maybe @gozer can also take a look at this.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -37,6 +37,7 @@ const RELOAD_DELAY = 2500;
 const gulp = require('gulp-help')(require('gulp'));
 const shell = require('gulp-shell');
 const path = require('path');
+const eventStream = require('event-stream');
 const ts = require('gulp-typescript');
 const insert = require('gulp-insert');
 const eslint = require('gulp-eslint');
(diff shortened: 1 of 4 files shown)


mikehenrty:
@nevik @Gregoor, should we wait on this one until after landing #508 ?

nevik:
Yes, IIRC this is already the way it is intended to work in #508, so since this isn’t anything really broken right now, I’d drop this for now

mikehenrty:
Ok, let's close in favor of switching to webpack.

PR #590, opened 2017-11-01
Improved the way S3 is created inside the server

### LINKED ISSUE #569: Unsupported signature version for S3 when using eu-central-1 region
**The problem**
When loading the home page, the requests to GET the audio resources are being rejected by AWS due to unsupported signature version. As a consequence, I cannot play the audio clips in the listen-box component. This component stays disabled.

```sh
GET https://mozz-voice-web-eu-01.s3.eu-central-1.amazonaws.com/...&Signature=8DI5xX66DyqpTMOFR8i4oYFrX0k%3D 400 (Bad Request)
```

```xml 
<Error>
    <Code>InvalidRequest</Code>
      <Message>
        The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256.
      </Message>
    <RequestId>227B51E6364C0B16</RequestId>
    <HostId>
        ...
    </HostId>
</Error>
```
**The cause**
I did my S3 setup to use eu-central-1 region (Frankfurt). This region, and most of the recently created ones, doesn't support the signature version 2. They now require the signature version 4. 

Two S3 instances were created inside the server. One was created with the signature V4 defined, but the other one wasn't. This cause the requests to be created with signature version 2, leading to the error described. 

**Proposed solution**

I did some modifications on the server code to have a single point of creation for the S3 object and the classes that need this object inside the API now receive this object through their constructors. This way we avoid some duplication and also the risk of having different instances of S3 initialized with different parameters. 

Also, I added a step 3 to aws CLI configuration section in the HOWTO_S3.md file just to make sure that the CLI will also use as default the signature version 4. 

Please take a look at the changes here [https://github.com/jcfausto/voice-web/commit/1053d8538be5458dca71b1c86e6cd593b6df1a41](https://github.com/jcfausto/voice-web/commit/1053d8538be5458dca71b1c86e6cd593b6df1a41 ) and let me know if you'd like to have a PR submitted for this issue. 

Kind regards, 

Julio.

jcfausto:
This PR improves the way S3 is created inside the server and solves the issue of getting an error when trying to use the eu-central-1 region for S3 buckets. Details were added to the issue that this PR resolves #569

### DIFF
--- a/docs/HOWTO_S3.md
+++ b/docs/HOWTO_S3.md
@@ -42,6 +42,7 @@ in which to store voice clips.
 
 1. Install the AWS command line UI. https://aws.amazon.com/cli/
 2. Run `aws configure` to configure credentials via the aws command line ui.
+3. Set the default signature version to be version 4: `aws configure set default.s3.signature_version s3v4`
 
 #### Add options to local `config.json`
 
--- a/server/src/lib/api.ts
+++ b/server/src/lib/api.ts
@@ -9,6 +9,8 @@ import Prometheus from './prometheus';
 import WebHook from './webhook';
 import respond from './responder';
 
+import { S3 } from 'aws-sdk';
+
 export default class API {
   config: CommonVoiceConfig;
   model: Model;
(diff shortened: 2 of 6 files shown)


mikehenrty:
great job on spotting and fixing this @jcfausto. I am going to make a few changes and merge it.

PR #601, opened 2017-11-03
Fix whitespace issue on record page

Gregoor:
Switched to conditionally rendering either recordings view or the review submissions view. This should fix the whitespace issues. Also we need less arbitrary offsetting of the submissions view like before.

But a problem that came up is the audio element inside of `ListenBox` not triggering `onCanPlayThrough` after re-recording. I "fixed" this, by setting `loaded` to `true` by default (it gets set to `false` when loading starts, so we should be good?!). Maybe you have an idea about that @mikehenrty?

### DIFF
--- a/web/src/lib/components/listen-box/listen-box.tsx
+++ b/web/src/lib/components/listen-box/listen-box.tsx
@@ -40,7 +40,7 @@ export default class ListenBox extends Component<Props, State> {
   }
 
   state: State = {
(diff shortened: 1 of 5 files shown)


mikehenrty:
Oh, I see now you mentioned this bug in #603. Let me try hiding the "Create Profile" button again and see if the whitespace issue is still fixed.

PR #599, opened 2017-11-03
Adding Havard-sentences #598

iveskins:
Adding Havard-sentences as suggested in #598 cs.cmu.edu.harvsents.txt
Also included a bunch of other sentence from Carnegie Mellon’s School of Computer Science old files directory 
http://www.cs.cmu.edu/afs/cs.cmu.edu/project/fgdata/OldFiles/Recorder.app/utterances/
Maybe contact the school for Usage rights confirmation..

PR #607, opened 2017-11-04
Adding CMU sentences, (this is an update from #599)

mikehenrty:
Included a bunch of other sentence from
Carnegie Mellon’s School of Computer Science old files directory
http://www.cs.cmu.edu/afs/cs.cmu.edu/project/fgdata/OldFiles/Recorder.app/utterances/


mikehenrty:
Thanks for the sentence @iveskins ! We already added the hardvard sentences, but we will stay away from the CMU sentences for now since we are not sure where they came from. Thanks for you contributions!

PR #612, opened 2017-11-06
Store emails and client ids in the database

mikehenrty:


server/src/lib/model/db/table.ts
```diff
@@ -1,9 +1,19 @@
 import Mysql from './mysql';
 
+export type ColumnType = {
+  [key: string]: string;
+};
+
+export type IndexType = {
+  name: string;
+  type: string; // UNIQUE|INDEX
```
Gregoor: Maybe go with an Enum here?

mikehenrty:
When the user changes their email address to one that already exists, that's not an error but a use case. That's a way of saying: "all these recordings I do from this web browser belong to this new email address". There can be multiple browsers (and client ids) all pointing at one email address. And user's may change this association at will.

There is no concept of "claiming" an email address as your own yet. You simply report which email address you want these clips to belong to. When we do a formal login system, then we can worry about validation ownership. For now i think it is sufficient to allow the user to choose and email address for association, even if we can't tell for sure the email is theirs yet.

Gregoor:
I'm with you on how it should work, but I originially thought it would error when a different `uid` (aka different browser) is used with the same email address. But I guess my thining was off, as `upsert` would then just update the `client_id`?

mikehenrty:
If we get a new UID (client id) with an existing email address, we associate that UID with the email address. An email address can have many (or none) UIDs associated with it. A UID can only have 0 or 1 email addresses.

PR #615, opened 2017-11-06
Webpack

### LINKED ISSUE #494: Build system should automatically import external libraries from node_modules

Gregoor:
Fixes #494

Based on @nevik's great work in #508, I made some modifications:

- fewer scripts, outsource heavier scripts into `scripts` folder
- share tsconfig between web and server
- remove (almost) all traces of gulp (@mikehenrty can you look at the remaining gulp stuff in nubis, or give me a hint what to change there?)
- move everything out of `devDependencies` that isn't a dev dependency
- log eslint errors when running the dev server


mikehenrty:
I'm **insulted** by how many problems you found and fixed in my code. I refuse to allow this.

web/src/main.ts
```diff
@@ -1,9 +1,8 @@
-declare var define: any;
+import './lib/components/index.css';
```
mikehenrty: ruh roh, is this the beginning of putting css inside js?

web/src/main.ts
```diff
@@ -1,9 +1,8 @@
...
+
 declare var require: any;
 
-define('preact', () => {
-  return preact;
-});
+require('es6-promise').polyfill();
```
mikehenrty: 💯

package.json
```diff
@@ -6,30 +6,39 @@
...
-    "gulp-clean": "^0.3.2",
-    "gulp-eslint": "^4.0.0",
-    "gulp-help": "^1.6.1",
-    "gulp-insert": "^0.5.0",
-    "gulp-shell": "0.6.3",
-    "gulp-uglify": "^2.1.2",
     "js-md5": "^0.6.1",
     "mediaserver": "0.1.0",
     "mkdirp": "^0.5.1",
+    "mysql": "2.13.0",
```
mikehenrty: did we need to add `mysql`? is this for bindings?

Gregoor: That was just moved to normal dependencies, didn't check that one if it still needed

server/src/config-helper.ts
```diff
@@ -39,17 +37,17 @@ const DEFAULTS: CommonVoiceConfig = {
...
 
 /**
  * Attempt to load a json file, but return null if not found.
  */
-function load(path: string): CommonVoiceConfig {
+function load(): CommonVoiceConfig {
   let config = null;
   try {
-    config = require(path);
+    config = require('../config.json');
```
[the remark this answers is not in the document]
Gregoor: I'm with you on that, I didn't have the patience to figure out why the import wouldn't work with an expression as path. Can look into it again

web/src/main.ts
Gregoor: Well, that's just how Webpack works, we're completely innocent 😁

server/src/config-helper.ts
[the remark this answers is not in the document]
nevik: possibly because if webpack might rewrite this path to refer to the same file while being in a different folder (i.e. where the output bundle is) -- which might only work if it's an inline literal.

pure guesswork though, would have to check this myself

server/webpack.config.js
```diff
@@ -0,0 +1,52 @@
+const fs = require('fs');
+const path = require('path');
+
+const OUTPUT_PATH = path.resolve(__dirname, 'js');
+
+/**
+ * For backend builds, we want to prevent webpack from bundling dependencies
+ * from node_modules into our output file (since we have access to node_modules
+ * at runtime), so this inserts runtime 'require()' calls using modules.
```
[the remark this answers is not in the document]
nevik: yeah, luckily the internet knows these things! it's my first time using webpack for backend code too

package.json
```diff
@@ -6,30 +6,39 @@
...
+    "@types/node": "7.0.29",
     "@types/node-fetch": "^1.6.7",
+    "@types/node-static": "^0.7.0",
+    "@types/random-js": "^1.0.30",
     "@types/tape": "^4.2.30",
+    "@types/webpack": "^3.0.13",
+    "@types/webpack-merge": "^4.1.0",
     "aws-sdk": "^2.67.0",
     "bcrypt": "^1.0.2",
+    "bluebird": "^3.5.0",
```
nevik: mmmmmm we shouldn't be needing bluebird anymore!

web/webpack.common.config.js
```diff
@@ -0,0 +1,52 @@
...
+        exclude: /node_modules/,
+        use: ['ts-loader'],
+      },
+      {
+        /**
+           * By default, Webpack (rather, style-loader) includes stylesheets
+           * into the JS bundle.
+           * 
+           * ExtractTextPlugin emits them into a separate plain file instead.
+           */
```
nevik: @mikehenrty actually, with this plugin we're currently keeping CSS _out_ of the JS (bundle) ;)

the import of the CSS file in the main TS file is just to give webpack an entry point for the CSS (this way, no CSS file needs to be specified separately for the build)

PR #619, opened 2017-11-09
Set bundle size limit to 100kb

Gregoor:
I do want this:
bundlesize
But it looks like they require all-access on the organization for that, which is a no-go.
https://github.com/siddharthkp/bundlesize#2-build-status

So for now it only checks whether the bundle size exceeds 100kb and if so, the test fails.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "webpack-merge": "^4.1.0"
   },
   "devDependencies": {
+    "bundlesize": "^0.15.3",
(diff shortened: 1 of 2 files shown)


Gregoor:
Actually this might still be useful, I just set the bundle size limit to 100kb, so we can easily see when PRs go over that. Though it would sure be nice to have this:
bundlesize

mikehenrty:
+1000

PR #682, opened 2017-11-20
Change review and submit text color. Fixes #671

### LINKED ISSUE #671: Review & Submit text on record screen should be grey
The following text should be grey #4A4A4A:
1. Thank you text
2. Sentence text
3. Review and Re-record text

screen shot 2017-11-19 at 5 22 41 pm

cc/ @m-branson

ukii:


Gregoor:
Thanks for the contribution, @ukii!

Could you change it so it uses the --grey-font-color var instead?

Gregoor:
One more nitpick (sorry that I didn't notice it before): The indentation is a bit off here and there. If you want I can also fix this when merging.

Gregoor:
Oops, looks like this was fixed already in some other commit. Also for some reason this PR touched and added many other non-related (and partly editor specific files). So I'll just go ahead and close it.

PR #686, opened 2017-11-21
Server Side Rendering (SSR)

Gregoor:
# Comparison

## Site rendering
Local requests with network throttled to fast 3G speed:

### Before (Without SSR)
5.5s to fully load the landing page

### After
Text & CSS loaded after 1s.
Landing page with all images after 2.5s **-46%**

## Server Load
There's bound to be a heavier load on the server / less requests can be handled per second. I used [wrk](https://github.com/wg/wrk) to get some numbers on that.

Both before & after were running 30s test @ http://localhost:9000
12 threads and 400 connections

### Before

Thread Stats | Avg | Stdev | Max | +/- Stdev
--- | --- | ---  | --- | ---
**Latency** | 112.64ms | 24.56ms | 416.80ms | 91.38%
**Req/Sec** | 295.56 | 59.61 | 494.00 | 84.04%

105521 requests in 30.09s, 226.03MB read

**Requests/sec:** 3506.83
**Transfer/sec:** 7.51MB 

### After

Thread Stats | Avg | Stdev | Max | +/- Stdev
--- | --- | ---  | --- | ---
**Latency** | 1.08s **+860%**  | 120.60ms **+391%** | 1.63s **+290%** | 91.72%
**Req/Sec** | 48.62 | 44.17 | 220.00 | 72.61%

10796 requests in 30.09s, 108.54MB read

**Requests/sec:** 358.83 **-90%**
**Transfer/sec:** 3.61MB **-52%**

# Future

What is suboptimal is that we don't know about the user at request time, so we can't preload sentences to speak/verify. Also the profile-form doesn't get populated on the initial render (otherwise there'd be a mismatch between what the server vs. the client renders). We'll look at cookie-based user identification in the future, when we have more user data stored in the backend.

### DIFF
--- a/server/src/render-web/index.html
+++ b/server/src/render-web/index.html
@@ -30,7 +30,7 @@
       }</script>
   </head>
   <body>
-    <div id="root"></div>
+    <div id="root"><!-- APP_HERE --></div>
     <script src="/dist/bundle.js"></script>
   </body>
 </html>
--- a/server/src/render-web/render-web.tsx
+++ b/server/src/render-web/render-web.tsx
@@ -1,0 +1,42 @@
+import * as fs from 'fs';
+import * as http from 'http';
+import * as path from 'path';
+import * as React from 'react';
+import { renderToNodeStream } from 'react-dom/server';
+import { StaticRouter } from 'react-router';
+
+import App from '../../../web/src/components/app';
+
+const [startHTML, endHTML] = fs
+  .readFileSync(
+    path.join(
+      __dirname,
+      '..',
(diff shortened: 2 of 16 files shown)


Gregoor:
I'm still interested in landing this, but also afraid to put too much load on our servers. One way to solve that would be only server rendering when CPU usage < ~50% and else we'd fall back to client side rendering.

Gregoor:
While I think we should revisit SSR at some point, we've moved to far apart from where this was branched off.

PR #697, opened 2017-11-24
dont swallow root transaction errors

mikehenrty:


mikehenrty:
@Gregoor can you take a look at this?

PR #703, opened 2017-11-27
make libri speech button open a new window

mikehenrty:


mikehenrty:
quick and dirty. @Gregoor let me know if you have a better way of doing this.

web/src/components/pages/data/data.tsx
```diff
@@ -224,9 +224,20 @@ class DataPage extends React.Component<Props, State> {
                 </div>
 
                 <div className="action">
-                  <CardAction onClick={() => this.showModalFor(dataset)}>
-                    <DownloadIcon />Download
-                  </CardAction>
+                  {dataset.name === 'LibriSpeech' ? (
```
Gregoor: We could replace the LibriSpeech dataset-object's download property with an empty array, and check for that here.

web/src/components/pages/data/data.tsx
```diff
@@ -224,9 +224,20 @@ class DataPage extends React.Component<Props, State> {
...
+                    <Button
+                      outline
+                      className="card-action"
+                      onClick={() => {
+                        window.open('http://www.openslr.org/12/', '_blank');
```
Gregoor: and then use `dataset.url` here

mikehenrty:
@Gregoor ah good call! I will make those changes before submitting. Thanks!

PR #737, opened 2017-11-30
Clips from DB

Gregoor:
Started this with going directly to the DB for every request, which took a sec even with this small dataset. So I built some per-instance caching into it, so it's reasonably fast now.

Also added a flag in the config for whether migrations should be enabled.

### DIFF
--- a/server/src/lib/model.ts
+++ b/server/src/lib/model.ts
@@ -3,6 +3,7 @@ import { UpdatableUserFields } from './model/db/tables/user-table';
 import Users from './model/users';
 import { default as Clips, Clip } from './model/clips';
 import { CommonVoiceConfig } from '../config-helper';
+import { DBClipWithVoters } from './model/db/tables/clip-table';
 
 const MP3_EXT = '.mp3';
 const TEXT_EXT = '.txt';
@@ -23,8 +24,12 @@ export default class Model {
     this.config = config;
     this.db = new DB(this.config);
     this.users = new Users();
-    this.clips = new Clips();
+    this.clips = new Clips(this.db);
     this.loaded = false;
+  }
+
+  set isMigrated(val: boolean) {
+    this.clips.isMigrated = val;
   }
 
   private addClip(userid: string, sentenceid: string, path: string) {
@@ -135,8 +140,8 @@ export default class Model {
   /**
    * Fetch a random clip but make sure it's not the user's.
    */
-  getEllibleClip(userid: string): Clip {
-    return this.clips.getEllibleClip(userid);
+  getEllibleClip(client_id: string): Promise<DBClipWithVoters | Clip> {
+    return this.clips.getEllibleClip(client_id);
   }
 
   /**
@@ -168,7 +173,7 @@ export default class Model {
    * Print the current count of users in db.
    */
   async printUserCount(): Promise<void> {
-    const [ users, clients ] = await Promise.all([
+    const [users, clients] = await Promise.all([
       this.db.getUserCount(),
       this.db.getClientCount(),
     ]);
--- a/server/src/lib/model/clips.ts
+++ b/server/src/lib/model/clips.ts
@@ -2,7 +2,7 @@ import * as Random from 'random-js';
 import DB from './db';
 import { DBClipWithVoters } from './db/tables/clip-table';
 
-const CLIP_CACHE_SIZE = 100;
+const CLIP_CACHE_SIZE = 1000;
 
 export class Clip {
   clipid: string;
@@ -81,7 +81,7 @@ export default class Clips {
   metrics: ClipMetrics;
   loaded: boolean;
   randomEngine: Random.MT19937;
-  isMigrated = false;
+  isMigrated: boolean;
 
   constructor(db: DB) {
     this.allClips = {};
@@ -91,6 +91,7 @@ export default class Clips {
     this.randomEngine = Random.engines.mt19937();
     this.randomEngine.autoSeed();
     this.loaded = false;
+    this.isMigrated = false;
   }
 
   private getClip(userid: string, sentenceid: string): Clip {
@@ -171,7 +172,8 @@ export default class Clips {
    */
   async getEllibleClip(client_id: string): Promise<DBClipWithVoters | Clip> {
     if (this.isMigrated) {
-      if (!this.clipsWithFewVotes.length) await this.refillClipsWithFewVotes();
+      if (this.clipsWithFewVotes.length == 0)
+        await this.refillClipsWithFewVotes();
 
       const i = this.clipsWithFewVotes.findIndex(
         clip => clip.client_id !== client_id && !clip.voters.includes(client_id)
--- a/server/src/lib/model/clips.ts
+++ b/server/src/lib/model/clips.ts
@@ -1,4 +1,8 @@
 import * as Random from 'random-js';
+import DB from './db';
+import { DBClipWithVoters } from './db/tables/clip-table';
+
+const CLIP_CACHE_SIZE = 100;
 
 export class Clip {
   clipid: string;
@@ -70,15 +74,18 @@ interface ClipMetrics {
  * Model for tracking clip information of all files in storage.
  */
 export default class Clips {
+  db: DB;
   allClips: ClipList;
   unverifiedClips: string[];
   verifiedClips: string[];
   metrics: ClipMetrics;
   loaded: boolean;
   randomEngine: Random.MT19937;
+  isMigrated = false;
 
-  constructor() {
+  constructor(db: DB) {
     this.allClips = {};
+    this.db = db;
     this.unverifiedClips = [];
     this.verifiedClips = [];
     this.randomEngine = Random.engines.mt19937();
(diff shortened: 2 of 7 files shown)


server/src/lib/model.ts
```diff
@@ -23,10 +24,14 @@ export default class Model {
     this.config = config;
     this.db = new DB(this.config);
     this.users = new Users();
-    this.clips = new Clips();
+    this.clips = new Clips(this.db);
     this.loaded = false;
   }
 
+  set isMigrated(val: boolean) {
```
mikehenrty: why are we using a setter? seems like a private member would do the same thing, and follow the convention currently set in the code

server/src/lib/model/clips.ts
```diff
@@ -70,15 +74,18 @@ interface ClipMetrics {
...
  */
 export default class Clips {
+  db: DB;
   allClips: ClipList;
   unverifiedClips: string[];
   verifiedClips: string[];
   metrics: ClipMetrics;
   loaded: boolean;
   randomEngine: Random.MT19937;
+  isMigrated = false;
```
mikehenrty: for consistency, would be good to declare this here, and then set the default value in the constructor

server/src/lib/model/clips.ts
```diff
@@ -1,4 +1,8 @@
 import * as Random from 'random-js';
+import DB from './db';
+import { DBClipWithVoters } from './db/tables/clip-table';
+
+const CLIP_CACHE_SIZE = 100;
```
mikehenrty: 100 seems a little on the small side, how about 1000?

server/src/lib/model/clips.ts
```diff
@@ -145,10 +152,36 @@ export default class Clips {
...
+    this.isRefilling = false;
+  }
+
   /**
    * Fetch a random clip but make sure it's not the user's.
    */
-  getEllibleClip(userid: string): Clip {
+  async getEllibleClip(client_id: string): Promise<DBClipWithVoters | Clip> {
+    if (this.isMigrated) {
+      if (!this.clipsWithFewVotes.length) await this.refillClipsWithFewVotes();
```
mikehenrty: nit: it's always better to be more semactic rather than rely on truthiness if possible, ie.:
```
if (this.clipsWithFewVotes.length < 1)

File: server/src/lib/model/clips.ts [hunk 5]
```diff
@@ -145,10 +152,36 @@ export default class Clips {
...
+
+      const i = this.clipsWithFewVotes.findIndex(
+        clip => clip.client_id !== client_id && !clip.voters.includes(client_id)
```
mikehenrty: 👍

server/src/lib/model/clips.ts
```diff
@@ -145,10 +152,36 @@ export default class Clips {
...
+      );
+      if (i == -1) return null;
```
mikehenrty: nit: please use strong equality whenever possible

server/src/lib/model/db.ts
```diff
@@ -130,4 +130,22 @@ export default class DB {
...
   }
+
+  async findClipsWithFewVotes(limit: number): Promise<DBClipWithVoters[]> {
+    const [clips] = await this.mysql.exec(
+      `
+      SELECT clips.*, group_concat(votes.client_id) AS voters
+      FROM clips
+        LEFT JOIN votes ON clips.id = votes.clip_id
+      GROUP BY clips.id
+      ORDER BY COUNT(votes.id)
```
mikehenrty: At some point if would probably be good to add a WHERE clause that eliminated all validate/invalidated clips from this query. For now, it's not a big deal since we have so many invalidated clips

server/src/lib/model/db.ts
```diff
@@ -130,4 +130,22 @@ export default class DB {
   endConnection(): void {
     this.mysql.endConnection();
   }
+
+  async findClipsWithFewVotes(limit: number): Promise<DBClipWithVoters[]> {
```
mikehenrty: Opinion: I think this query should like in the clips-table object (and this db function should query to a similar function there). The weird part is that it's a join, but it is a query based on clips (which just pulls in data from votes), and I think we want these things in the table objects.

server/src/server.ts
```diff
@@ -179,7 +179,10 @@ export default class Server {
       this.print(`${getElapsedSeconds(start)}s to perform maintenance`);
     }
 
-    // await migrate(this.model.db.mysql.conn);
+    if (this.config.ENABLE_MIGRATIONS) {
+      await migrate(this.model.db.mysql.conn);
+      this.model.isMigrated = true;
```
mikehenrty: why are we setting `isMigrated` outside of model? shouldn't model keep this internally, and set this internally?

server/src/lib/model.ts
Gregoor: the Model itself doesn't need the flag, so the setter is just used to forward the flag

Gregoor:
The new clips/votes going to DB part is indeed not part of the PR but it's what I'm currently working on. I agreee with your point, it makes sense to add the PR with the mutative DB interactions, before having this PR. I'll get the other one up EOD.

> Also why are we changing the download-email pop-up with this change? That seems like it belongs in another PR.
Yaiks, how did that get there 🙈  TY for catching that

Appreciate the comments 👍

server/src/lib/model/clips.ts
Gregoor: I'd say when using TS it's fine for comparing literals (because then you can be sure that you're comparing literals). We could add an ESLint rule for that.

server/src/lib/model/db.ts
Gregoor: I thought db would be a good place to have those kind of fns for now because 
- given that a migration library is still on the table, the future of the `*-table.js` files seems uncertain
- fewer middlemen forwarding method calls

server/src/server.ts
Gregoor: Would you prefer `server` calling `model.setIsMigrated(true)`. I'm agnostic about it tbh.

mikehenrty: I know it's too late now, but I'll add some general thoughts.

Right now we are mixing OOP approaches with more functional approach, which is why the above is confusing. We have this function that we import (without being attached to any object) which modifies state, and now we have to manually keep track of the fact that we modified that somewhere. Setting the `isMigrated` flag on the model object explicitly is a symptom of this problem, and it not good OOP practice (more about that in a bit).

Better would be to do something like this:
```
this.mode.migrate();
```
which internally to model will set a private `isMigrated` flag after it does it's work. So then elsewhere in server you could call something like `this.model.isMigrated()` to check if the migration happened. 

I know this involves forwarding calls (ie. `migrate` in this case) through multiple objects, which you may not like. But it is much more readable that mixing functional and OOP approaches, and also setting flags on an object when an object has no idea about that property. As we sit now, we are breaking OOP division of responsibilities by having isMigrated on model, because model has no idea what that is or where it comes from. This will make it harder for other developers to read and figure out what we have done and why we have done it.

I'm not pushing back too hard yet, because we plan to rip out a bunch of this stuff once the DB is in a happy place. But I think this is something to keep in mind going forward.

Gregoor: It's never too late and I'm always happy to talk about clarity of code 👍 

I only partially agree that model having a isMigrated flag violates division of responsibilities as model in its current form has a larger set of (high-level) responsibilities under which `isMigrated` could reasonably fall. Also I don't quite see how changing a setter to a method call would unviolate that principle. And we still need to inform `clips.ts` of the state of the migration as that is where the branching happens.

Would the OOP solution be to have an object with a migration method and a flag-property that indicates whether the migration has completed (and then pass that wherever branching needs to occur)? I can see doing it that way, when the migration would result in multiple properties being set.

mikehenrty: The flag itself is not violating division of responsibilities, and I agree private member var with accessor is same as setter.

What's violating division of responsibilities in this case is that `server` invokes a migrate function it imported from db within the model, and then explicitly sets isMigrate. To adhere to division of responsibilities, migrate should happen within the model, and the model should set it's own `isMigrated` flag. Having `server` play such a conductor role where migrate happens without going directly through the model, and modle is only used as a namespace to store a flag about migration is not a pattern i've seen before, and I'd venture to say it's not good practice.

Speaking of `clips`, since it lives inside the model, the model should inform clips if and when migration happens. As of now, it's unclear how much control/responsibilities the model has over migration, and in my opinion it should be the conductor, not `server`.

Generally OOP is a nice way to say who does what. I think the current setup it's hard, because sometime we take an object/state approach, and sometimes we are taking a more functional approach, and we should aim for consistency. Since OOP is what we have been using from the ground up, I feel like we should adhere to that. In a very specific sense, this means that model should probably do the migration, and set it's own `isMigrated` flag, and also inform sub-components when and if migration has occurred.

mikehenrty: It may seem like I'm speaking a lot about something that is essentially a nitpick. And I can understand that point of view. But in reality, this little thing has important ramifications.

I think the way we are doing migration, ie. having `server` import and invoke a function from deep within the model subcomponents, and then after completion setting a flag on a different subcomponent breaks OOP convention (for reasons explained above). In the short term, it's easy to see what's going on and so it no big deal. But as time goes on if we keep allowing anti-conventions to slip in, the project becomes much harder to reason about. And will be unfriendly to new people on the project.

Gregoor: I think I probably didn't reason too much about who's responsibility it is, because when I think about it now I'd agree with you that it feels more natural to put it in the model (which would probably forward the call to the db). The reason I've done it was probably a pragmatic one: The API needs to know about the successful migration as well. So if we move it into the model (which we probably should do for separation of concerns reasons), we'd need to pass a migration-complete callback into the model. But while that's something worth mentioning and I wouldn't shy away from doing the thing that makes more sense even if it complicates things a bit.

> It may seem like I'm speaking a lot about something that is essentially a nitpick. And I can understand that point of view. But in reality, this little thing has important ramifications.

Not at all. I think it's great we can have these kinds of discussions. Imo as soon as something seems off to someone, it's legitimate to talk about it. The best case for me is that I learn something, which isn't that bad of a thing for me. 😁

PR #741, opened 2017-12-01
Added new sentences

reinhart1010:
These new sentences are created to popularize the rare-occuring words from current Common Voice sample texts, as about 1/3 of the words used in these sentences only occurs **once**.

[Current word statistics](https://gist.github.com/reinhart1010/8717b4d765ee4e5eb0765aa4c9955016)


mikehenrty:
@reinhart1010 wow, thanks for the word statistics and for helping the situation!

PR #742, opened 2017-12-01
Fix some grammar mistakes

kuhsaft:
* Missing apostrophe 's'
* 's of
* it's vs its
* St., Mrs., Mr.
* ATVs
* EMTs
* 20s, 30s, 50s, 60s, 70s, 80s
* -ies
* 's' to 'a'
* 'an' before vowel sounds
* 'a' before non-vowel sound
* Some typos


kuhsaft:
I sent an email concerning this issue and decided to work on it instead of just complain. I think adding a feature to report/correct errors in sentences would be useful in maintaining this project.

kuhsaft:
There were missing commas and other punctuation that I didn't yet fix. There seems to have been a parsing error when creating the data set because all punctuation were replaced with spaces.

mikehenrty:
hi @kuhsaft ! thanks for this. i have a script that parses the flickr30k dataset and removes some punctuation. looks like i got a little too eager with the removal. i'll fix this shortly, and then maybe you can have another look at this

Issue #718, opened 2017-11-28
[Mobile] Top Navigation menu item needs more space on top


ivonnekn:
On mobile screens, when opening the navigation menu, the top item needs more space above it. The space above the (first) item should match the space below the first item (it currently looks smaller).

![screen shot 2017-11-28 at 12 28 38 am](https://user-images.githubusercontent.com/1466558/33303800-4c4aeff2-d3d3-11e7-84a3-e8ffad44cfd5.png)

cc/ @m-branson

PR #750, opened 2017-12-04
[Mobile] Pad menu items from the top.

willstott101:
Menu items had padding at the top and bottom. Therefore the first item was closer to the top, than it was from the second item.

Before:
capture1

After:
capture

Fixes: https://github.com/mozilla/voice-web/issues/718

### DIFF
--- a/web/src/components/nav.css
+++ b/web/src/components/nav.css
@@ -318,6 +318,6 @@ footer a:hover, footer button:hover {
 }
 
 #navigation-modal a {
-    line-height: 3.2;
+    padding-top: 2.5rem;
     font-size: 1.3rem;
 }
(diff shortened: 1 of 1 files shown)


Gregoor:
This PR has the side effect of changing the clickable area of the links, as before there was an uniformly spaced area around the text that was clickable, but with the padding-top this clickable area is only above the text.

Can you revert the change and increase the `#navigation-modal`'s top padding? I think that might achieve the desired effect.

willstott101:
That's a good point, I had done that originally but opted for this approach as then the top padding was defined in the same way as the in-between spacing. Hadn't thought of that side-affect. Will update later on.

willstott101:
I opted to place the padding in `#navigation-modal .nav-list` to keep the offset for the header separate and avoid the `calc()` (even though it'd currently be optimised out). It just felt more right to me.

Happy to change though, if that seems dumb.

Gregoor:
I agree with not going with calc here, looks good and simpler this way 👍 Thanks!

Issue #716, opened 2017-11-28
[Mobile] Move "Delete Profile" in profile form to below heading


ivonnekn:
On mobile screens, in the Create a Profile box, keep the heading and the text link on same column (rather than 2 columns), that way the heading is on 1 line, and the "Delete Profile" text link on another line. That would allow more horizontal space for each item.

cc/ @m-branson

PR #753, opened 2017-12-05
Move "Delete Profile" link to its own line

flyingpimonster:
Fixes issue #716 by changing the wrapping div to flex-direction: column
and centering the "Create a Profile" header when on mobile. Screenshot:
screen shot 2017-12-05 at 12 02 26


Gregoor:
Bueno 👌

Issue #714, opened 2017-11-28
[Mobile] Subfooter needs adjustments to icons


ivonnekn:
On mobile screens, adjust the subfooter with the icon links such that:

1. divider lines are gone
2. size of icon and text are reduced by 20%

![screen shot 2017-11-28 at 12 10 23 am](https://user-images.githubusercontent.com/1466558/33303368-d5061d42-d3d0-11e7-9d0d-865919f4dd14.png)

cc/ @m-branson

jdittrich:
when touching this, we could also fix the strange outline widths of the Discourse icon.

flyingpimonster:
@jdittrich That appears to have been fixed in commit ea761fb.

flyingpimonster:
With the reduced icon size, I think it looks better to also reduce the height of subfooter. Comparison:

![screen shot 2017-12-05 at 13 43 42](https://user-images.githubusercontent.com/24497353/33627176-51558f26-d9c2-11e7-8b4f-54eac7674beb.png) ![screen shot 2017-12-05 at 13 43 28](https://user-images.githubusercontent.com/24497353/33627180-530a668e-d9c2-11e7-9212-4b95f2d3631b.png)


Thoughts?

Gregoor:
@flyingpimonster I agree, looks better with reduced height 👍

PR #757, opened 2017-12-06
Tweaks to help links subfooter

flyingpimonster:
Intends to fix issue #714 by removing the dividers, reducing icon and text size by 20%, and making the subfooter shorter. These changes only apply on mobile devices. Screenshot: 
screen shot 2017-12-06 at 07 43 47

### DIFF
--- a/web/src/components/nav.css
+++ b/web/src/components/nav.css
@@ -72,6 +72,20 @@ footer a:hover, footer button:hover {
     height: 2.5rem;
 }
 
+@media (--mobile) {
+    #help-links .divider {
+        display: none;
+    }
+
+    #help-links img {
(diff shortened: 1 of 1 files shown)


web/src/components/nav.css
```diff
@@ -72,6 +72,24 @@ footer a:hover, footer button:hover {
...
+    #help-links .divider {
+        display: none;
+    }
+
+    #help-links img {
+        width: 21px;
+    }
+
+    #help-links a {
+        font-size: .8em;
```
Gregoor: Would be great if we could use `var(--small-font-size)` here

Gregoor:
Looks great!

PR #774, opened 2017-12-13
Corpus: cleanup used data sets

caugner:
The goal of this PR is to cleanup the used corpus data sets.

- [x] Add missing punctuation.
- [x] Fix typos.
- [x] Replace numbers/symbols with words.
- [x] Replace remaining numbers/symbols.
- [x] Add quotes for names/titles.
- [x] Resolve abbreviations
- [x] Replace dates/years by words.

### DIFF
--- a/server/data/ajaydee.txt
+++ b/server/data/ajaydee.txt
@@ -136,7 +136,7 @@ It's a private club, for only the most respectable people.
 Peter shrunk back, aghast.
 We've been laughing all day about that.
 Monkeys smile when they are scared.
-My dog is an alsatian, we call her 'Poppy'.
+My dog is an Alsatian, we call her 'Poppy'.
 Mozzarella is a creamy cheese that goes well with tomato and basil.
 Jade's cordless drill has become faulty.
 The whole office was struck by a deafening silence.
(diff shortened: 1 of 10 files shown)


caugner:
@mikehenrty Can we remove the `not-used` datasets or are they worth being cleaned up?

Gregoor:
@caugner For now we need to keep them around, to migrate all the sentence to the DB. At some point in the future the sentences might be removed from the repo entirely, but for that we first need to build/use tools to make contributions and modifications of the corpus simpler.

caugner:
This PR is no longer WIP and ready to merge from my side.

server/data/ajaydee.txt
```diff
@@ -544,7 +544,7 @@ The horse galloped uncontrollably.
 The decision was already made.
 They jetted off to the Bahamas.
 The priest collapsed backwards.
-It was a bill for 120 pounds.
```
kscz: Out of curiosity, this interpretation is specific.  Do we want to have a separate version where the interpretation is "one hundred and twenty"?

mikehenrty:
thanks @caugner!

server/data/ajaydee.txt
mikehenrty: hi @kscz, what do you mean by keeping a separate interpretation?

kscz: The sentence could be read both as:
* "It was a bill for one hundred and twenty pounds"
* "It was a bill for a hundred and twenty pounds"

I'm not sure if, in this file, there are similar example like:
* "There were at least one thousand two hundred ducks at the park"
* "There were at least twelve hundred ducks at the park"

I wasn't sure if we were trying to capture the variety of ways people speak a number, or if we're purely trying to train the phonemes for English.

mikehenrty: gotcha. yeah, that is the exact reason that raw numbers are bad in these sentences. ideally, we would have every possible way to read this number as a separate sentence in this file. but for now, @caugner's fix is a huge improvement to the ambiguity we had before.

PR #799, opened 2018-01-11
Record without MediaRecorder

Gregoor:
It adds 60kb and blocks the main thread, I don't like it too much. Looking into a worker solution.

TODO:
- [ ] make sure it runs on Edge
- [ ] ...Safari
- [ ] ...old Android Browser
- [ ] ?add banner informing iOS 11 app-users that they don't need the app anymore?


Gregoor:
=> https://github.com/mozilla/voice-web/pull/800

PR #800, opened 2018-01-12
Record without MediaRecorder

### LINKED ISSUE #53: [iOS] We need a png to be displayed when the user is offline

### LINKED ISSUE #54: [iOS] We need a splash screen

### LINKED ISSUE #184: [iOS] Prevent webapp content from being selected inside the Webview

### LINKED ISSUE #469: Add support for Safari 11 & Edge
Safari 11 on macOS and Safari on iOS 11 (supposedly) add support for WebRTC. This could make the CV native wrapper app obsolete.

We should investigate whether the support for UserMedia in current Safari suffices for CV's needs, and think about when to abandon the app if this is the case.

_Note: I mostly wanted to bring this up so we don't forget. I'm going to look around and add some sources for this in the next few days. I might also try around in the simulator and/or on my iOS devices._

### LINKED ISSUE #605: [iOS] The app on iOS shows the "Open in App" banner.
This check must be failing:

https://github.com/mozilla/voice-web/blob/master/web/src/lib/components/pages.tsx#L351

Gregoor:
TODO:
- [x] make sure it runs on Edge
- [x] ...Safari
- [ ] ...iOS Safari
- [ ] ...old Android Browser
- [ ] only show banner for iOS < 11 users
- [ ] ?add banner informing iOS >= 11 app-users that they don't need the app anymore?

Fixes #469 

And since we'd probably stop work on the iOS app with this:
Closes #605, closes #184, closes #54, closes #53


guest271314:
Is the required resulting audio extension requirement `.mp3`?

Gregoor:
While it runs fine on Mac Safari now, real device iOS Safari is not recording properly. Unfortunately the iPhone I have atm is not tied to the Mac I borrowed and the Simulator is behaving even stranger. Soo I think I'll have to borrow an iOS device again that I can tie to this Mac >.<
Unless someone else wants to look into this?

Gregoor:
I'm just gonna close this. I think the option we'd rather look at in the future is one with WebASM.

PR #822, opened 2018-01-31
Scrolling shouldn't be disabled

### LINKED ISSUE #819: [iOS] can't scroll in iOS native app
iPhone SE
iOS 11.2.5

syilmaz:
Looking through the commit history, I'm not sure why scrolling is only enabled on iPads.

This pull request resolves #819.

### DIFF
--- a/ios/voicebank-ios-wrapper/ViewController.swift
+++ b/ios/voicebank-ios-wrapper/ViewController.swift
@@ -32,7 +32,6 @@ class ViewController: UIViewController, WKScriptMessageHandler, WKNavigationDele
         webView?.isHidden = true
         webView?.navigationDelegate = self
         webView?.configuration.userContentController.add(self, name: "scriptHandler")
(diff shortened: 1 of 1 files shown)


syilmaz:
Travis build failing doesn't seem to be related to the changes:

```
 FAIL  server/src/test/upload-recording.test.ts (6.646s)
      saving clip error { CredentialsError: Missing credentials in config
```

mikehenrty:
thanks for the PR @sylimaz!

Indeed all our PR travis builds are failing, due to a configuration problem. Sorry about that.

The reason we had disabled scrolling for iPhone native is that in an older version of the website, we would handle scrolling with CSS in a nested element on mobile screens. We have since then simplified the code to always use top-level scrolling, but forgot to test the iPhone app.

Thanks for the fix!

mikehenrty:
We fixed this in the web client, and we have no plans to update the iOS client at this time. So closing this.

PR #832, opened 2018-02-02
Handle multiple environments for aws_db_parameter_group

gozer:
- Don't hard-code a resource's name, instaed build it up with environment, region, etc
 - link resources with their internal IDs instead of hard-coding names


gozer:
@Gregoor This should fix production deployments!

Gregoor:
Was wondering where I went wrong. Thanks @gozer, I won't hardcode names again!

Issue #842, opened 2018-02-06
Fluentify Profile-Page


Gregoor:
See meta-issue #838 for a description

Issue #841, opened 2018-02-06
Fluentify Profile-Component


Gregoor:
See meta-issue #838 for a description

MekliCZ:
Not sure how to fix that. :/ @stasm, could you take a look? More info is in #845.

Gregoor:
One option would be to use the [withLocalization-HOC and `getString`](https://github.com/projectfluent/fluent.js/blob/master/fluent-react/examples/higher-order/src/App.js) to explicitly set the label. I have not yet found word on whether it's possible to only set attributes, without replacing children.

PR #845, opened 2018-02-07
Fluentify profile page

MekliCZ:
Not sure if this fixes only #841, #842 or both?

### DIFF
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -39,3 +39,23 @@ status-loading = Loading...
 status-hours = { $hours } validated hours so far!
 status-goal = Next Goal: { $goal }
 status-more-soon = More languages coming soon!
+
+[[ProfileForm]]
+profile-form-email
+    .label = Email
+profile-form-username
+    .label = User Name
+profile-form-emails = Yes, send me emails. I'd like to stay informed about the Common Voice Project.
+profile-form-language
+    .label = Language
+profile-form-more-languages = More languages coming soon!
+profile-form-accent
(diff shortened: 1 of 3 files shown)


Gregoor:
This fixes indeed both. I put them into separate issues, because the Profile-Form is also used in Recording. And I forgot that we don't have that much content on the Profile-Page yet 😋 
Anyway, thanks for the PR. Looks good!

Gregoor:
Don't know how I missed that, but looks like the Localized wrapped-selects didn't render their child options elements anymore. So I reverted that particular change and reopened the issue.

stasm:
Hey, thanks for discovering this bug and for your patience. (I'm back from a week-long vacation.) I published [fluent-react 0.6.1](https://github.com/projectfluent/fluent.js/releases/tag/fluent-react%400.6.1) which has the fix for disappearing `<options>` of `<selects>` wrapped in `<Localized>`. Can you try it out and see if it helps?

Gregoor:
Thanks @stasm, that did the trick. I did have to tweak our build process a bit to support transpiling (i.e. object spread) node_modules, but since that seems to be the way of the future it's probably not a bad idea to start getting used to it. 👍

stasm:
Importing from `fluent-react/compat` shouldn't require transpilation, but I remember you said TS did module resolution differently. I'm happy to hear it wasn't much extra work to get object spread working. Also, hello ES2018 :)

Gregoor:
Uh right, forgot about that 🙈 Well, good to be ready for the new year of ES :)

PR #850, opened 2018-02-07
Pr/fluentify static

### LINKED ISSUE #843: Fluentify static pages
FAQ, Terms & Privacy

See meta-issue #838 for a description

MekliCZ:
fixes #843 

after consultation with @stasm I have updated fluent-react to 0.6.0 (otherwise, I would not be able to insert HTML to FTL) but kept fluent on 0.4.2; sorry for not making separate PR for this

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,9 @@
     "nodemon": "^1.12.1",
     "react-dev-utils": "^4.2.1"
   },
+  "resolutions": {
+    "fluent": "0.4.2"
+  },
   "engines": {
     "node": ">=8.9.0"
   },
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "fluent": "^0.4.2",
     "fluent-intl-polyfill": "^0.1.0",
     "fluent-langneg": "^0.1.0",
-    "fluent-react": "^0.4.1",
+    "fluent-react": "^0.6.0",
     "glob": "7.1.2 ",
     "jest": "^22.1.1",
     "js-md5": "^0.6.1",
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -9,7 +9,7 @@ terms = Terms
 cookies = Cookies
 faq = FAQ
 content-license = Creative Commons license
-content-license-text = Content available under a { $link }
+content-license-text = Content available under a <licenseLink>{ content-license }</licenseLink>
 share-title = Help us find others to donate their voice!
 
 [[Home Page]]
@@ -26,7 +26,7 @@ wall-of-text-more-mobile =
   recognition open to everyone.
 wall-of-text-more-desktop =
   Now you can donate your voice to help us build an open-source voice database that anyone can use
-  to make innovative apps for devices and the web.{ $lineBreak }
+  to make innovative apps for devices and the web.<lineBreak></lineBreak>
   Read a sentence to help machines learn how real people speak. Check the work of other
   contributors to improve the quality. It’s that simple!
 show-wall-of-text = Read More
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -66,7 +66,7 @@ notfound-content = I'm afraid I don't know what you're looking for.
 
 [[Privacy]]
 privacy-title = Common Voice Privacy Notice
-privacy-effective = Effective June 19, 2017
+privacy-effective = Effective { DATETIME($date, month: "long", year: "numeric", day: "numeric") }
 privacy-policy = When Mozilla (that’s us), receives information from you, our <policy>Mozilla Privacy Policy</policy> describes how we handle that information.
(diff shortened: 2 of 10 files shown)


MekliCZ:
Looks like I forgot to fluentify FAQ. Don't merge until I'll push FAQ, please :)

stasm:
Nice work, @MekliCZ !

`fluent-react` 0.6.0 removes the possibility of passing elements as $arguments, so you'll also need to fix these two messages in the PR:

https://github.com/mozilla/voice-web/blob/d36ed1117ba8635f89e0fa489cc56a2363d19b3a/web/locales/en/messages.ftl#L12
https://github.com/mozilla/voice-web/blob/d36ed1117ba8635f89e0fa489cc56a2363d19b3a/web/locales/en/messages.ftl#L29

See the [migration notes](https://github.com/projectfluent/fluent.js/blob/master/fluent-react/CHANGELOG.md#dont-pass-elements-as-arguments) for more information about this.

MekliCZ:
Everything fixed, FAQ fluentified, ready to merge. :)

Gregoor:
Thanks you two!
For some reason I get the regeneratorRuntime error (from fluent 0.6.0), even though our package.json specifies 0.4.0. Does it run fine for you @MekliCZ? Might also make a difference whether you use npm or yarn (we're using yarn).

MekliCZ:
@Gregoor I'm using Yarn and this happens sometimes. What resolves it for me is recompilation.

Gregoor:
@stasm I never really published a library, so take this with a grain of salt, but shouldn't `fluent` be a peerDependency of `fluent-react`?

stasm:
It's a [hard dependency](
https://github.com/projectfluent/fluent.js/blob/22da361f77185012d20ca62ec9f5f45e00dd3ddc/fluent-react/package.json#L35) because it exposes [methods for iterating over a sequence of `MessageContexts`](https://github.com/projectfluent/fluent.js/blob/22da361f77185012d20ca62ec9f5f45e00dd3ddc/fluent-react/src/localization.js#L1).

Gregoor:
Ah dang, that makes sense. But then I don't see a way how we can currently use `fluent-react` 0.6.0, as that'll lead us to having `fluent` in there twice with different versions:
```
├─ fluent-react@0.6.0
│  └─ fluent@0.6.0
└─ fluent@0.4.2
```
Unless there's a way for `fluent-react` to use `fluent` 0.4.x when the root application uses is, but I'm way out of my depth here. Not sure if that's possible or makes sense.

Unfortunately I can't get it to not error for me (and also wouldn't like it much if it error'd sth like 50% of the time).

stasm:
I would think `npm` would be capable of re-using `fluent@0.4.2` in this case... Let me see.

MekliCZ:
@Gregoor Dunno why, but I killed the server, started it again and I'm not able to load the react app without regeneratorRuntime. :/

stasm:
I just did the following:

```bash
mkdir tmp && cd tmp
npm i fluent@0.4.x
npm i fluent-react
```

And this is what I see in `npm ls`:

```
/home/stas/tmp
├─┬ fluent-react@0.6.0
│ ├── fluent@0.4.2
│ └─┬ prop-types@15.6.0
│   ├─┬ fbjs@0.8.16
│   │ ├── core-js@1.2.7
│   │ ├─┬ isomorphic-fetch@2.2.1
│   │ │ ├─┬ node-fetch@1.7.3
│   │ │ │ ├─┬ encoding@0.1.12
│   │ │ │ │ └── iconv-lite@0.4.19
│   │ │ │ └── is-stream@1.1.0
│   │ │ └── whatwg-fetch@2.0.3
│   │ ├── loose-envify@1.3.1 deduped
│   │ ├── object-assign@4.1.1 deduped
│   │ ├─┬ promise@7.3.1
│   │ │ └── asap@2.0.6
│   │ ├── setimmediate@1.0.5
│   │ └── ua-parser-js@0.7.17
│   ├─┬ loose-envify@1.3.1
│   │ └── js-tokens@3.0.2
│   └── object-assign@4.1.1
└── UNMET PEER DEPENDENCY react@^0.14.9 || ^15.0.0 || ^16.0.0
```

Did you try `npm install`? It should remove unused packages. Or nuke `node_modules`? :)

Gregoor:
Interesting, thanks for looking into that @stasm! I'll try to figure out why yarn behaves differently there.

Gregoor:
Completely forgot about yarn `resolutions`, which allows us to specify which version a package resolves to (even for dependencies). That fixes it for me!
Unfortunately I can't push to this branch, so could you add this under `devDependencies` in `package.json` and see if it works for you as well @MekliCZ?
```
"resolutions": {
  "fluent": "0.4.2"
}
```

MekliCZ:
Works for me

Gregoor:
Thank you everyone, great work 👍

PR #851, opened 2018-02-07
Add editorconfig

### LINKED ISSUE #849: Editorconfig
I would add .editorconfig so the code will be formatted the same in every editor that respects editorconfig. What do you think @Gregoor?

MekliCZ:
fixes #849

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.html]
+indent_style = space
+indent_size = 2
+
+[*.{ts,tsx,js}]
+indent_style = space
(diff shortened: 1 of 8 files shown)


Gregoor:
Works 👍 Maybe we should use this as an opportunity to make existing css files conform to this?

MekliCZ:
I can reformat them all, if you want.

Gregoor:
That'd be great! For some reason I can't push to this branch directly, even though GitHub is telling me it should be possible 😒

MekliCZ:
@Gregoor Are you now able to push to my branch? I have added you as an collaborator. This is definetly not my last PR so it will be useful in the future. :)

Gregoor:
That did work, thank you :)

PR #855, opened 2018-02-09
Express backend

### LINKED ISSUE #834: Backend Router
Current routing logic could be more declarative by using a router. I'd just go with express as it is quite solid and well-known (which helps contributors read the code).

There's an error in the way we do routing, e.g. https://voice.mozilla.org/metrics/upload/random.json gives us a random recording as we don't check for exact route match but rather whether certain routes are included in the string. When I first noticed that I didn't want to change it, because I didn't know whether anything relied on that behavior. Now I'm fairly certain that this isn't the case. This would also be fixed by using a router.

@apoorvaagupta expressed interest in tackling this.

apoorvaagupta:
Partially fixes #834 
Hello @Gregoor ,
I have just setup the basic server part!
To run and test, you can use following commands:
`yarn &&
npm run clean-express &&
npm run build-express &&
npm run start-express`

there may come an error if you run the build command but it can be ignored right now. I will fix it later and the start command can run even with that error!

To test atm, you can go to localhost:3000/api/status
this is a dummy route to test if the api is working or not!

Now I will start adding the api routes to it!!


Gregoor:
Thanks @apoorvaagupta, but I wonder if we could start with a more integrated approach, e.g. without a separate folder, changing the tsconfig, separate start scripts etc. That might mean breaking the old server for a few commits but imo that's no problem as we can squash all those commits together once the PR is done (as to not have broken commits).

apoorvaagupta:
Thanks @Gregoor , But I think you can go ahead and setup a little basic work because I am not sure how exactly it is to be started. If you start the basic PR, then i would be able to get the better hang of it and start contributing further!

Gregoor:
Heya @apoorvaagupta, I got around to starting the basic PR #864.

PR #867, opened 2018-02-13
Improve FTL and l10n support

flodolo:
I can drop the change to single unicode character if you prefer, but that's what is usually used across Mozilla projects.

The new FTL syntax also has an equal sign when the value is empty

```properties
foo =
    .title = bar
```

**l10n.toml**
You can use it get the status of locales, find errors, etc.

Install [compare-locales](https://pypi.python.org/pypi/compare-locales), then run `compare-locales l10n.toml . en ab-CD` (where `ab-CD` is a locale code). You can also run it on multiple locales, as part of automation, etc.

CC @pike

### DIFF
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -36,7 +36,11 @@ help-us-explain = Press play, listen & tell us: did they accurately speak the se
 status-title = Overall project status: see how far we’ve come!
 status-contribute = Contribute Your Voice
 status-loading = Loading…
-status-hours = { $hours } validated hours so far!
+status-hours =
+    { $hours ->
+        [one] One validated hours so far!
+       *[other] { $hours } validated hours so far!
+    }
 status-goal = Next Goal: { $goal }
 status-more-soon = More languages coming soon!
 
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -13,7 +13,7 @@ share-title = Help us find others to donate their voice!
 
 ## Home Page
 home-title =
-  The Common Voice project is Mozilla's initiative to help teach machines how real people speak.
+  The Common Voice project is Mozilla’s initiative to help teach machines how real people speak.
 home-cta = Speak up, contribute here!
 wall-of-text-start =
   Voice is natural, voice is human. That’s why we’re fascinated with creating usable voice
@@ -33,7 +33,7 @@ help-us-title = Help us validate sentences!
 help-us-explain = Press play, listen & tell us: did they accurately speak the sentence below?
 
 ## ProjectStatus
-status-title = Overall project status: see how far we've come!
+status-title = Overall project status: see how far we’ve come!
 status-contribute = Contribute Your Voice
 status-loading = Loading…
 status-hours = { $hours } validated hours so far!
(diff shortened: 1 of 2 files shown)


flodolo:
You can also see some usage of compare-locales in the wild, e.g. in [Notes](https://github.com/mozilla/send/tree/master/scripts).

web/locales/en/messages.ftl
```diff
@@ -31,38 +31,43 @@ wall-of-text-more-desktop =
...
 help-us-explain = Press play, listen & tell us: did they accurately speak the sentence below?
+
 ## ProjectStatus
-status-title = Overall project status: see how far we've come!
+status-title = Overall project status: see how far we’ve come!
 status-contribute = Contribute Your Voice
-status-loading = Loading...
-status-hours = { $hours } validated hours so far!
+status-loading = Loading…
+status-hours =
```
flodolo: @pike 
Do you know if this require [code changes](https://github.com/mozilla/voice-web/blob/master/web/src/components/pages/home/project-status.tsx#L70)?

web/locales/en/messages.ftl
```diff
@@ -31,38 +31,45 @@ wall-of-text-more-desktop =
...
+    { $hours ->
+        [one] One validated hours so far!
+       *[other] { $hours } validated hours so far!
+    }
+# Variables:
+# $goal - number of hours representing the next goal
```
flodolo: Is this correct?

flodolo:
Side question: should the liability section be translated from English? I assume Legal has an opinion about that.

web/locales/en/messages.ftl
Pike: nope, yay, fluent

Gregoor:
Thank you for looking into all that!

web/locales/en/messages.ftl
```diff
@@ -31,38 +31,45 @@ wall-of-text-more-desktop =
...
+
 ## ProjectStatus
-status-title = Overall project status: see how far we've come!
+status-title = Overall project status: see how far we’ve come!
 status-contribute = Contribute Your Voice
-status-loading = Loading...
-status-hours = { $hours } validated hours so far!
+status-loading = Loading…
+status-hours =
+    { $hours ->
```
Gregoor: out of curiosity, is that always required? e.g. in this case it's a safe assumptions that hours is > 1

flodolo: That's not how plurals work in a lot of languages ;-)

There are [18 variants in Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals#List_of_Plural_Rules), from 0 to 6 different forms, depending on the number.

Gregoor: fair enough 👍 I guess it would be syntactic sugar for cases where languages with "threshold-plurals" are used in context where a value wouldn't be under that threshold.

flodolo: I confess that I got lost in your reply :)

What I meant is: there are languages where the noun is declined differently if `$hours` is 2, 3, 4 or 6 (Welsh, for example).

en: 2 cat, 3 cats, 4 cats, 6 cats.
cy: 2 gi, 3 chi, 4 ci, 6 chi.

The only way you can support that is by using a proper plural form. The number of forms will be determined depending on the language.

PR #881, opened 2018-02-20
Fix FAQ string

### LINKED ISSUE #876: Localisation string specifies English
```
faq-native-q  I am a non-native English speaker and I speak with an accent, do you still want my voice?
```

Should this be ```I am a non-native { $lang } speaker ...``` ?

MekliCZ:
fixes #876


TonnesM:
It looks like "English" is now always used for localized versions. Should it?

Gregoor:
@TonnesM good catch. Indeed that changed, now we'd want it to be general again.

PR #882, opened 2018-02-27
Split

### LINKED ISSUE #806: Determine train / dev / test split before recording utterances
I’m a speech researcher and I want use common voice speech data for my experiments. Unfortunately there is a big problem with how the corpus (v1) and specifically the train/test/dev split is designed. This issue is also being discussed here: https://github.com/kaldi-asr/kaldi/issues/21419 (Commonvoice results misleading, complete overlap of train/dev/test sentences #2141) and I've posted the issue to the discourse message board, too. 

There is a near complete overlap (>99%) of train/dev/test sentences in Common Voice v1:

unique sentences in train: 6994
unique sentences in dev: 2410
unique sentences in test: 2362
common sentences train/dev (overlap) = 2401
common sentences train/test (overlap) = 2355

The goal should be to reach the optimum of **no sentence overlap** + **no speaker overlap** in the train / dev / test split as close as possible. Ideally, also, the dev and test utterances are all unique utterances (no repetitions), to accurately measure an unbiased WER. The train corpus may have repetitions, but having too many, as right now in the v1 release, might be detrimental and would lead to overfitting in typical speech recognition models. But if dev/test are correctly set up, then WER numbers can atleast be accurately measured and the effect might be taken into account. My guess is that right now utterances to record are selected uniformly at random from a corpus of prespecified text. This random selection process needs to be changed so that some new text inputs are labeled apriori as belonging to dev/test and are recorded just a single time, while the selection process for training utterances remains as is.

Otherwise, any WER numbers reported on this train/dev/test split are pretty much meaningless unfortunately and encourage absolute overfitting on the training data. Also best results are obtained when the language model is trained only on the train sentences, without any other sentences. This is pretty much what the Kaldi recipe is doing now (https://github.com/kaldi-asr/kaldi/blob/master/egs/commonvoice2) and it encourages recognizing only these ~7000 sentences and nothing else really. 

Solving this critical bug is important to get publications and researchers to use the corpus.

Gregoor:
Resolves #806 

Here's how it works:
- when importing sentences
  - look at distribution of (used) sentences
  - based on the distribution & the [ideal split](https://github.com/Gregoor/voice-web/blob/1467ca98366bbdcef8dfa466fa6fb312de1d9824/server/src/lib/model/split.ts#L12-L16), [pick a random bucket](https://github.com/Gregoor/voice-web/blob/1467ca98366bbdcef8dfa466fa6fb312de1d9824/server/src/lib/model/split.ts#L57-L69) for each new sentence
- when a new user is created
  - look at clip distribution
  - based on the distribution & the [ideal split](https://github.com/Gregoor/voice-web/blob/1467ca98366bbdcef8dfa466fa6fb312de1d9824/server/src/lib/model/split.ts#L12-L16),  [pick a random bucket](https://github.com/Gregoor/voice-web/blob/1467ca98366bbdcef8dfa466fa6fb312de1d9824/server/src/lib/model/split.ts#L57-L69) for the user
- when serving a sentence, only serve the ones from the users bucket


server/src/lib/model/db.ts
```diff
@@ -42,9 +42,20 @@ export default class DB {
...
+  ): Promise<void> {
     let { age, accent, email, gender } = fields;
     email = this.formatEmail(email);
+    await this.mysql.query(
+      `
+        INSERT INTO user_clients (client_id, bucket) VALUES (?, ?)
+        ON DUPLICATE KEY UPDATE client_id = client_id
+      `,
+      [client_id, bucket]
+    );
```
mikehenrty: Hrm, this adds an additional upsert every time we want to update the user. This is confusing and also spreads the user_client queries between separate code modules (db, and userClient). Let's make this a part of the userClient update call instead (which will require some more robust logic to determine when to reset the bucket, which I think will also be good for code readability and quality).

server/src/lib/model/db/import-sentences.ts
```diff
@@ -109,13 +114,34 @@ export async function importSentences(pool: any) {
   await pool.query(
     'DELETE FROM sentences WHERE id NOT IN (SELECT original_sentence_id FROM clips)'
   );
+
   await pool.query('UPDATE sentences SET is_used = FALSE');
 
+  const [rows] = await pool.query(
+    'SELECT bucket, COUNT(bucket) AS count FROM sentences GROUP BY bucket'
```
mikehenrty: In the name of DRY, this query string should probably be abstracted somewhere, since we use it in db.getClipBucketCounts

server/src/lib/model/split.ts
```diff
@@ -0,0 +1,85 @@
...
+      return obj;
+    }, {});
+}
+
+/**
+ * Given a real split, in absolute numbers (i.e. count of occurrences) calculate a compensatory
+ * split which would move the real data closer to the split defined as IDEAL_SPLIT.
+ */
+export function calculateCompensatorySplit(realSplit: Split): Split {
+  const compensatorySplit: any = {};
```
mikehenrty: shouldn't this be `Split` type?

server/src/lib/model/split.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  for (const [bucket, count] of Object.entries(normalize(realSplit))) {
```
mikehenrty: We need some comments for each step of this function, since it's not obvious what is going on at first glance.

server/src/lib/model/split.ts
```diff
@@ -0,0 +1,85 @@
...
+    );
+  }
+
+  const totalShares = sumValues(compensatorySplit);
+
+  if (totalShares === 0) return IDEAL_SPLIT;
+
+  for (const [bucket, share] of Object.entries(compensatorySplit)) {
+    compensatorySplit[bucket] = share / totalShares;
+  }
```
mikehenrty: More readable, but slightly less efficient, would be something like:
```
if (sumValues(compensatorySplit) === 0) return IDEAL_SPLIT;
return normalize(compensatorySplit);

File: server/src/lib/model/db/import-sentences.ts [hunk 6]
```diff
@@ -2,6 +2,11 @@ import * as fs from 'fs';
 import * as path from 'path';
 import promisify from '../../../promisify';
 import { hash } from '../../clip';
+import {
+  calculateCompensatorySplit,
```
mikehenrty: Doesn't look like we explicitly need `calculateCompensatorySplit`, as it's buried in `randomBucketFromDistribution`

server/src/test/split.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      pass: !found,
+      message: () =>
+        `Expected ${printReceived(received)} to be equal to ${printExpected(
+          argument
+        )}`,
+    };
+  },
+});
+
+describe('calculate compensatory split', () => {
```
mikehenrty: Love to see these tests! But since you went through all the trouble to create this nice setup, let's test a few more edge cases (all zeros, all 1s, all 0.5s, etc etc). Just want to make sure we're properly handling unexpected input.

server/src/lib/model/db/migrations/20180221154147-add-bucket-to-user_clients-and-sentences.ts
```diff
@@ -0,0 +1,13 @@
...
+  await db.runSql(
+    `
+      ALTER TABLE user_clients ADD COLUMN bucket ENUM ('train', 'dev', 'test') DEFAULT 'train';
+      ALTER TABLE sentences ADD COLUMN bucket ENUM ('train', 'dev', 'test') DEFAULT 'train';
+      ALTER TABLE clips ADD COLUMN bucket ENUM ('train', 'dev', 'test') DEFAULT 'train';
+    `
+  );
+};
+
+export const down = function(): Promise<any> {
```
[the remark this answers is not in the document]
Gregoor: Apparently we can omit it (haven't tried before).

server/src/lib/model.ts
```diff
@@ -1,14 +1,47 @@
...
 
 /**
  * The Model loads all clip and user data into memory for quick access.
  */
 export default class Model {
   db = new DB();
   clipCache = new Cache(count => this.db.findClipsWithFewVotes(count));
-  sentencesCache = new Cache(count => this.db.findSentencesWithFewClips(count));
+  sentencesCaches: { [bucket: string]: Cache<string> } = Object.keys(
+    IDEAL_SPLIT
```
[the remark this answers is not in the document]
Gregoor: Yeah, unfortunately there's no way to get it from the Interface.

server/src/lib/model/db.ts
Gregoor: I put it outside of the `userClient.update` because every other field in there can be changed, while the bucket should be only set on creation. (Maybe I should add a comment explaining that reasoning).
I guess I could also overwrite `userClient.update` and just move the call into it. Would that be what you have in mind?

mikehenrty: yeah, probably better to have both queries there (if two queries are unavoidable)

Issue #828, opened 2018-01-31
When sentences can't be loaded, the error message displays improperly


mikehenrty:
![screenshot from 2018-01-31 23-17-31](https://user-images.githubusercontent.com/635365/35650636-16530604-06dd-11e8-9dc0-d7a6e2efd0f1.png)

himanish-star:
@mikehenrty, can I try and fix this issue?

mikehenrty:
Hi @himanish-star, we would love your help here! Let us know if you have any questions.

himanish-star:
Sure I'll complete the setup and start my work

himanish-star:
Actually @Gregoor I have a number of commitments at other repositories, so I am not quite getting time to look upon this issue. So is it ok if someone else can take over this?

Gregoor:
Sure no worries, I'll look into it. Just wanted to make sure I'm not eating someone else's pie 🙃

flyingpimonster:
I went ahead and submitted PR #883 to fix this. I hope that wasn't someone else's pie--it looked extremely delicious.

PR #883, opened 2018-03-02
Fix issue #828: When sentences can't be loaded, the error message displays improperly.

flyingpimonster:
- Wrap the "sentences are being loaded" error message in a `<div>` and `<p>`
  that give it a solid background, as well as fading the text to give a
  visual indication that something is missing. This is the same as what
  the validation box does when there are no sentences to validate.
- Remove the "1 of 3" text when there are no sentences to read.


Gregoor:
Looks good, thank you! 🙂

Issue #608, opened 2017-11-05
Review button during playback not working 


iveskins:
After recording a set of sentences, when reviewing your recordings, if you press the play button on a sentence and then while the audio is still playing, press the re-record button, the re-record button will be highlighted black, and the audio will stop, but you won't be taken to the redo page. At this point you have to push the re-record button again.
I expected the button to work the same way (one touch) even if audio is playing.

benjamin21st:
Can I take this one?

benjamin21st:
I think i found the culprit, small fix, will send for PR in a bit :)

PR #885, opened 2018-03-03
fix: #608 expect to one touch and do rerecord

benjamin21st:
The `return` statement blocked further code execution, which is why user needs to click twice in order to rerecord. Essentially, the first time for "pausing" the playback, and the second time to actually switch view to rerecord.


Gregoor:
Thank you! I'm positively surprised it was that simple 🙂

benjamin21st:
You are welcome! I’m glad I could help!

PR #895, opened 2018-03-18
Fix focus on listen box

### LINKED ISSUE #894: Outline on voting
On voting element, there is an outline on focus which looks pretty horrible.

MekliCZ:
fixes #894


Gregoor:
Thanks!

PR #897, opened 2018-03-19
Improve local development setup using docker

johngian:


johngian:
Hey @Gregoor! 

This is pretty much the groundwork I've done to setup my environment locally. Let me know if this is helpful for the project.

Gregoor:
Very helpful, thankslots @johngian! I hope this makes setup easier for contributors.

The only part I'm not 100% sure about now, is directly using `getConfig()` in `aws.ts`. The config is passed down in other places so that we can have a separate config for stuff like testing. That might be relevant here when running the upload-recording test, which interacts with S3.
So what I'd prefer doing is having a more global config injection mechanism, so that we don't have to pass the config around for every little dependency. I'll create another PR for it and will merge this one after.

Gregoor:
Okay that config part was more straightforward than expected. Thanks again! :)

PR #909, opened 2018-04-03
Fix #908

### LINKED ISSUE #908: On short, wide screens, the bottom of the mobile navigation menu shows
screenshot_20180403-161253

flyingpimonster:
Fixes issue #908: On short, wide screens, the bottom of the mobile navigation menu shows

On screens that are short, but wide enough not to trigger the mobile
layout, the bottom of mobile navigation modal still showed up. This
fixes the problem by setting the modal to `visibility: hidden` when it
is not active.


Gregoor:
Thanks! 👍

PR #910, opened 2018-04-04
update language tatoeba dataset based on their feedback

mikehenrty:


mikehenrty:
@Gregoor I got some requests to change the copy on our site from the tatoeba folks. Am I doing this the right way?

Gregoor:
LGTM! 👍 Though I just now realized that we're duplicating the descriptions. I'll remove them all in a separate commit.

PR #922, opened 2018-04-17
Update copy for requesting or getting involved with the multi-languag…

mikehenrty:
…e effort


Gregoor:
As `request-language-success-text` and `language-updates-success-text` aren't used anymore, we should also delete them in the non-english language files, so that only strings that are being used are loaded.

PR #925, opened 2018-04-18
Change data-download-no to No (was 'yes')

jimregan:


Gregoor:
Thanks for catching that. Since some translators already used the wrong string I supplied, I went in and removed that string completely (and added a new one called `data-download-deny`).

Issue #773, opened 2017-12-13
user-scalable=no - Zooming and scaling must not be disabled


KBPsystem777:
I found `user-scalable=no` under the `<meta name="viewport">`  at **[voice-web/web/index.html](https://github.com/mozilla/voice-web/blob/master/web/index.html)**
This tag disables zooming in the browser for several devices.

mikehenrty:
Whoops, I broke this. @KBPsystem777 can you tell me why we should not disable zoom?

KBPsystem777:
Thanks for looking on this @mikehenrty . 
This is a very critical app feature, disabling zooming capabilities on mobile browser users may not be a good experience for people that has a low or vision problems who simply relies on screen magnification/zooming to properly see the contents of a web page.
And I know people personally who experienced the same problem after running the site on their mobile browsers. This might help: https://dequeuniversity.com/rules/axe/2.2/meta-viewport

PR #974, opened 2018-04-30
disable user zoom ability, to prevent mobile safari from zooming on s…

### LINKED ISSUE #960: Mobile navigation 'zooms in' when activating translate, doesn't zoom back out.
@Gregoor when testing mobile on iPhone X navigation translate zooms in when activated. After interacting with and closing translation menu screen does not zoom back out. Bug or possible to turn this off?

mikehenrty:
…elect box, fixes #960

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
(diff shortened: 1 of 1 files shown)


Gregoor:
We might get pushback on disabling zooming (#773)

PR #979, opened 2018-05-02
add cc-0 license to text files, fixes #968

### LINKED ISSUE #968: text files (/server/data) license
What is the license of all text files? CC-0, right? (like the recordings)
If that's the case, then there should be a LICENSE file in that folder, imo

https://github.com/mozilla/voice-web/tree/master/server/data

I just wanted to bring this up, so those text files can also be used for all other purposes outside of this, since they can be useful for training several kinds of algorithms.

mikehenrty:


mikehenrty:
@Gregoor can you take a look?

PR #982, opened 2018-05-04
Multi-Language Voice Contribution & Validation

Gregoor:
Fixes #213 

Some things we should figure out before merging:
- [x] #933 Shortcuts across languages
- [ ] what to do with accents across languages?

PR #985, opened 2018-05-06
Irish sentences

jimregan:
Depends on #982


jimregan:
@Gregoor - this now contains more than 2000 sentences. Is anything else required?

Issue #981, opened 2018-05-02
Please add /contribute.json to your project.


floatingatoll:
Hi, the HTTP Observatory noticed that your site is missing a valid `contribute.json` file. While it isn't necessary to fill out every possible field in it, various incident response teams at Mozilla do really appreciate having one with the basics "where to file bugs" "what is this site" filled out. Webops can assist with writing one and/or there's a lot of examples hiding on various Mozilla sites as well.

https://observatory.mozilla.org/analyze.html?host=voice.mozilla.org

Undigon:
I'll start looking at https://www.contributejson.org/ and see what I can make with it.

PR #992, opened 2018-05-10
Add contribute.json

Undigon:
I made this because of issue #981.

I scanned the contribute.json at https://www.contributejson.org/ and it passed.
Please review the keywords that I have chosen.


mikehenrty:
Thanks so much @Undigon !

Why do we need the file in both locations? In the interest of DRY, i'd rather not include the copy. I wonder if we can copy it into the web folder as part of the build process. @Gregoor any thoughts?

Gregoor:
Thankslots for adding that! I removed the duplicated file from `web/` and changed the server to serve this file from the top-level.

Issue #990, opened 2018-05-08
Get involved popup doesn't show completely on the staging server


danielsjf:
I'm using a OnePlus 3T cellphone using Chrome. See #982.

Reprex:
1) Go to voice.allizom.org
2) Open menu and go to LANGUAGES
3) Click on the collaborate button on any language

I can never scroll completely down on this screen. There are always a few lines that are hidden.

![bug](https://user-images.githubusercontent.com/6696142/39768374-704b111c-52e9-11e8-8dad-654bc35b52ab.jpg)

Djfe:
I can confirm this.
Firefox on Android works fine though (I'm a OnePlus 3T user as well ^^)

PR #994, opened 2018-05-10
WIP - Remove display-table to on modal css to resolve issue 990

mrchrisadams:
This seems to resolve a display issue (#990) that I investigated with @mikehenrty at the voice sprint.

The `display-table` property here causes an issue where text flow on the language flows off the bottom of the modal you can see this in the screenshot from #990 (this affected mobile firefox and chrome when we looked into it together):

before-modal-borkage

Removing the `display-table` property resolves it, but I understand this to be used for vertical centering normally, and I'm not sure what layouts this is likely to affect.

Is there chance someone give me an idea of where i might look, to see if fixing this issue breaks things elsewhere?

### DIFF
--- a/web/src/components/modal/modal.css
+++ b/web/src/components/modal/modal.css
@@ -1,5 +1,5 @@
 .ReactModal__Overlay {
-    display: table;
+    /* display: table; */
     z-index: var(--top-z-index);
     width: 100%;
     height: 100%;


mikehenrty:
Here is the fixed version, scrolled all the way down:

photo5451827268893649055

mikehenrty:
@Gregoor would you have a look?

Gregoor:
Looks good to me and I was not able to come up with an edge-case for which we needed display: table. Thanks!

Issue #1010, opened 2018-05-19
The link to the TED-LIUM corpus is outdated


Franck-Dernoncourt:
The link to the TED-LIUM corpus on https://voice.mozilla.org/en/data is outdated. The latest version of the TED-LIUM corpus is v3, which may be found at https://lium.univ-lemans.fr/en/ted-lium3/

missuniverse:
Correction, the link is: http://www.openslr.org/resources/51/TEDLIUM_release-3.tgz
The external url is the one you stated.

lissyx:
@Franck-Dernoncourt Training has been done on v2 of the corpus, should we still update the link ?

mikehenrty:
@lissyx yes we still want to update the link. even though the DeepSpeech models were based off the older version of the data, Common Voice is a separate project, and so we want to promote the most up-to-date dataset regardless of what DeepSpeech is using.

kdavis-mozilla:
@mikehenrty @lissyx No models released by Deep Speech made use of TED data. TED data is not licensed so as to be usable in commercial products.

Gregoor:
fixed through #1015

PR #1015, opened 2018-05-21
Updated link to TED-LIUM.

missuniverse:
Adjusted the link to the third release instead of the first.
Concerns issue #1010


Gregoor:
Thanks!

PR #1019, opened 2018-05-24
Added email validation with a regular expression (Fixes #980)

### LINKED ISSUE #980: [Backend] Email Validation
we've been seeing some weird email address show up in the DB, we need to do some validation before saving it in the db.

igomez10:


server/src/lib/model/db.ts
```diff
@@ -29,13 +29,12 @@ export default class DB {
 
   /**
    * Normalize email address as input.
-   * TODO: add validation here.
    */
-  private formatEmail(email?: string): string {
-    if (!email) {
-      return '';
+   private formatEmail(email?: string): string {
+	let re = new RegExp('/^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/');
```
MichaelKohler: please check the indentation here. Also, would it make sense to use another variable name to describe what the regex does? maybe something like invalidRegex?

MichaelKohler: (as discussed: as the RegEx tests for a valid email, the if below should be inverted ;))

Gregoor:
Thanks you two!
From reading the [StackOverflow comments](https://stackoverflow.com/a/46181/1501916) it sounds like there could be valid email addresses that would be excluded with this regex. Maybe we should use the RFC one instead?

mikehenrty:
@Gregoor what's the RFC one?

Gregoor:
It's the answer underneath the one that was used for this PR. Though it is suboptimal as well apparently (but at least it's right around 99.9% of the time). Then again the solution we have here is probably good enough for our purposes. Thanks for the PR!

PR #1021, opened 2018-05-25
add more validated sentences from jf99

mikehenrty:


mikehenrty:
more reviewed sentences from @jf99 (reviewed by @MichaelKohler and @elioqoshi)

jf99:
The first 400 sentences are not from me. It seems that quotation marks are no longer desired?

MichaelKohler:
I had talked to Michael before reviewing and we decided to remove the quotes as these do not provide additional value for speech recognition itself. This also might make the model better, depending on how the quotes would be treated by the machine learning algorithm (which I have no idea about).

Great work with those sentences though! Thanks!

PR #1062, opened 2018-06-06
add Irish (ga-IE) corpus

jimregan:


mikehenrty:
hi @jimregan , thanks for this! i will need to go over this and cut out any sentences that are more than 14 words (we are aiming for a length of 5 seconds per recording). if you could do that, that would help us move quicker! it but then i think we can launch Irish very soon!

jimregan:
14 words? ouch.

mikehenrty:
@jimregan or whatever you think is 5 or so seconds long. some people have been breaking up their longer sentences into parts (like by comma), if it helps not lose too many sentences.

Here's a guide we've been using about writing sentences:
https://voice-sprint.mozilla.community/contributing/

PR #1071, opened 2018-06-08
Fixed punctuation marks in german and french sentences

Djfe:
@mikehenrty This should fix the punctuation characters that went missing due to the script's bug.
I also fixed around 3 other small issues in German, I stumbled upon.
It would be nice, if we could find someone that speaks french to take a short look at this, I'm not 100% sure, if this many exclamation marks are accurate.
For German the exclamation marks have just been restored from @jf99 's file

### DIFF
--- a/server/data/de/jf99.txt
+++ b/server/data/de/jf99.txt
@@ -29,8 +29,8 @@ Stecker sollten niemals am Kabel aus der Steckdose gezogen werden.
 Planfeststellungsverfahren ist doch ein wunderbares deutsches Wort.
 Die Sache hat allerdings einen Haken.
 Wenn du das Handtuch so aufhängst, wird es morgen immer noch nicht trocken sein.
-Heute gibt es "Asterix und Kleopatra" im Fernsehen.
-Auf Kika läuft gerade "Jim Knopf und Lukas der Lokomotivführer".
+Heute gibt es Asterix und Kleopatra im Fernsehen.
+Auf Kika läuft gerade Jim Knopf und Lukas der Lokomotivführer.
 Geschrieben von Michael Ende und gespielt von der Augsburger Puppenkiste.
 Als Kind hatte ich alle Folgen vom Sams auf Kassette.
 Heutzutage wissen die Kinder wahrscheinlich gar nicht mehr, was eine Kassette ist.
@@ -90,7 +90,7 @@ Bremen und die anderen Stadtstaaten sind mal wieder die Schlusslichter der PISA-
 Otto kommt aus Ostfriesland.
 Heißt du wirklich Uli, oder ist das die Abkürzung für Ulrich?
 Trotz seiner relativ geringen Größe hat Brandenburg an der Havel immerhin eine Straßenbahn und zwei Hochschulen.
-"Meinen Bachelor in Mathematik habe ich zuvor in Jena gemacht", erzählte Anton.
+Meinen Bachelor in Mathematik habe ich zuvor in Jena gemacht, erzählte Anton.
 Berta hat uns zu einer Weinprobe nach Kaiserslautern eingeladen.
 Trier ist die älteste Stadt Deutschlands und wurde ursprünglich von den Römern gegründet.
 Gaius Julius Cäsar kannte sie jedoch noch nicht.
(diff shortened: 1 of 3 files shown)


Djfe:
I can't fix cy, because I couldn't find the source strings in https://github.com/Common-Voice/text-tools

missuniverse:
I'll take a look tomorrow, I'm a native french speaker.

Djfe:
Lots of them were in line.
sentence1 ! sentence2

or
sentence1 !

and the script split them like this (used the ! for a new line but removed the !):
sentence1
sentence2

then users looked at them and might've corrected errors but probably didn't add ! back.

now they are
sentence1 !
sentence2

I added ! after every new line without a punctuation character

if that's wrong in most cases, then we should drop that part of my commit.

missuniverse:
@Djfe I didn't spot any errors, except you might want to remove the space before the exclamation point, otherwise its weird.

Djfe:
ok, I will do that.

The source files (or some of them) had spaces before the exclamation marks.
Therefore I actually looked it up, and the internet was unsure about it as well.

Djfe:
Done.
I also rebased the commits into one.

Djfe:
@missuniverse Thank you for looking at the file!

lissyx:
@Djfe I'm unsure about your fixes (for french), you are adding excalmation marks everywhere while the original data does not have those. And you are adding that in a bogus way since there is no space while it should be here.

Djfe:
I know :/
I don't speak French and couldn't find the original strings.
Wait so there needs to be a space in front of exclamation marks?
I changed my commit after @missuniverse told me that those look weird to her.

Would you be up for the task of readding the exclamation marks to the file where they are missing?

lissyx:
@Djfe So we discussed with @missuniverse on another issue, it turns out that fr_FR and fr_CA do not follow the same typographic rules, and that in canadian's french your syntax is valid. But it is not for french of France :-). Regarding the source data, it's extracted (mostly) from french parliament debates and project Gutenberg file. At least most of the sentences where you added exclamation marks were from the parliaments debates.

It's hard to link to direct source for them, since it's all in one big XML file, but it's available from http://data.assemblee-nationale.fr/travaux-parlementaires/debats

Djfe:
I recently heard of something similar (Belgian Dutch) hehe

the "originals" should be in here:
https://github.com/Common-Voice/text-tools/tree/master/txt

but I couldn't find them.
the csv folder contains the output which the script in that repo accidentally deprived of all exclamation marks.

the csv's were uploaded to Crowdin, proofread, and then uploaded to the voice-web repo

lissyx:
@Djfe Okay. Well my processed source out of the XML that Michael used is at https://github.com/lissyx/commonvoice-syceron/tree/master/data

lissyx:
For example, https://github.com/mozilla/voice-web/pull/1071/files#diff-14f63c6b9d6bf33e26979ec66935813dR16 is likely to come from https://github.com/lissyx/commonvoice-syceron/blob/2c10e475a253108ae309439c8d0e4b59b9e3813c/data/debats-assemblee-nationale/20140611213000000.txt#L94

mikehenrty:
thanks for everyone's work here. please tag me when there is agreement on when this is ready to be merged.

zecakeh:
Hi there! I was actually planning on doing a full review of this file in the next few days because I've seen a lot (and I mean a lot!) of mistakes when trying to record/validate: spelling or grammar errors, missing accents, missing punctuation, missing capitalized letter at the start of the sentence, some sentences are in Breton (which is not french at all).

I also think there are too many exclamation marks in this new version. In my opinion we should just put those where necessary.

About the (unbreakable) space in front of punctuation marks, since the rule is always the same, a simple script could change that according to the region when the sentences are used in production. According to [the Wikipedia page about differences in typographic rules in french speaking countries](https://fr.wikipedia.org/wiki/Comparatif_des_diff%C3%A9rents_codes_typographiques_francophones) it seems canadian french is the odd one for this rule, so I'd vote for putting unbreakable spaces in front of !, ? and ;.

lissyx:
Thanks, I'll have a look tomorrow, but I dont have merge credentials :-)

zecakeh:
I'm not an expert with git but from what I read online, to have my commit included, I have to send a PR on @Djfe 's fork, which I did a couple of days ago: [Fix punctuation #6](https://github.com/Djfe/voice-web/pull/6). So know we have to wait for him to accept it for my commit to appear here.

Djfe:
screenshot_20180706-160127__01
how shall I resolve this merge conflict?

mikehenrty:
>  @mikehenrty we had a few similar cases in german, where two short sentences that belong next to each other where separated by your script.
It might be better to only separate sentences in one line if the number of characters of one line is longer than you want them to be.

Yup, I still need to get back to those text-tools and your PR at some point. For now, I appreciate you raising that again.

Djfe:
@zecakeh 
should I always prefer the new version or are there cases in which I should prefer head? (for the french sentences)
(you can take a look by clicking on resolve conflicts)

mikehenrty:
@Djfe, that conflict comes from #1222. I can fix those :) 

Thanks for your help everyone!

Djfe:
you just reverted at least #1189
Maybe another commit as well, let me check

Djfe:
It was only that one commit #1227

mikehenrty:
thanks for catching that @Djfe !

PR #1091, opened 2018-06-09
[German data] Small punctuation fix

Ordoviz:
Similar to #1077 and #1071


Djfe:
Thx, for correcting those! :)

Gregoor:
Thanks indeed!

Issue #1108, opened 2018-06-14
French corpus and configuration errors


missuniverse:
I discovered many errors in the french corpus relating to the punctuation & duplicates.

There's also a good amount of errors related to the gender of words in the configuration files that use french language.

I will propose a pull request shortly that will address those issues.
@mikehenrty

missuniverse:
My mistake, canadian french is different then french from france.

PR #1109, opened 2018-06-14
Correction of french corpus and config

missuniverse:
Corrections on the french text and sentences in both the corpus and the configs.
@mikehenrty 

Related to #1108

### DIFF
--- a/server/data/fr/ofis_publik_ar_brezhoneg.txt
+++ b/server/data/fr/ofis_publik_ar_brezhoneg.txt
@@ -1,5 +1,4 @@
 10 personnes.
-10 personnes.
 12 % des enfants de la commune de Plouvien sont scolarisés dans la filière bilingue.
 22 octobre.
 300 phrases anglaises de base.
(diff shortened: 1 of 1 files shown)


mikehenrty:
@lissyx, would you mind have a look at this?

lissyx:
I'm not sold on the "de" to "de la", that feels weird when reading. the diff of breton is waaay to big to be manually reviewed on github, it looks like it's completely different file :(

missuniverse:
@lissyx Concerning the use of feminine for the country names, this is a dictionary reference:
http://www.larousse.fr/encyclopedie/departement/Guadeloupe_971/122457
https://www.larousse.fr/dictionnaires/francais-espagnol/Guadeloupe/100852
According to the dictionary, we should be using "de la" instead of "de".

Also, for the second file, there was over 500 duplicates and I had to change the order of the text to remove the duplicates, which is why it looks so different.

lissyx:
Looking at each diff, even the one removing duplicates add a lot of new strings, it's hard to check.

lissyx:
@missuniverse Your links to larousse does not give anything, while for instance I find https://fr.wikipedia.org/wiki/Fran%C3%A7ais_de_C%C3%B4te_d%27Ivoire

lissyx:
@TheoChevalier any opinion there ?

TheoChevalier:
Omitting the pronoun doesn’t make the noun masculine, so I disagree with this change and I suggest we keep the current translations.
We say `Je parle le français de France`, not `Je parle le français de la France`.

missuniverse:
@lissyx 
I'll just say this as a comment, irony:
image

I guess it depends on where your from that it sounds right or not.

lissyx:
Yes, please, and can you also make the other changes more readable ?

missuniverse:
@mikehenrty @lissyx I removed the config file changes, but concerning making the changes more readable, the only way would be if someone sorted the original file, then the diff would be way easier to understand. But otherwise I don't know how I can make it more readable.

lissyx:
@missuniverse you just gave the solution yourself, rebase and sort it before doing your changes :-)

missuniverse:
@lissyx I adjusted the pull request like you said.

server/data/fr/ofis_publik_ar_brezhoneg.txt
```diff
@@ -12,200 +11,180 @@
 Absolument identique.
 Absolument nulle part.
 À cause de cela.
-À cause de quoi ?
+À cause de quoi?
```
lissyx: This is wrong.

server/data/fr/ofis_publik_ar_brezhoneg.txt
```diff
@@ -12,200 +11,180 @@
...
 à ce propos.
 À certains endroits.
 À certains endroits il y a des manques.
 À ce soir.
-à ce sujet.
-à cette époque, les gens n'avaient pas beacoup de temps libre.
+À ce sujet.
```
lissyx: Wrong as well

server/data/fr/ofis_publik_ar_brezhoneg.txt
```diff
@@ -12,200 +11,180 @@
...
 À gauche.
 Ah, bon.
-Ah, c'est bon !
-Ah ! Tu es étais là ?
-à huit heures je suis rentré à la maison.
-aimes.
-aimez.
-aimons.
+Ah, c'est bon!
+Ah! Tu es étais là?
```
lissyx: Wrong and wrong, you remove spaces

missuniverse:
@lissyx To my knowledge there is no accepted use of a space before a punctuation mark at the end of a sentence in french. Same thing for the capital, all sentences have them at the start.
@TheoChevalier Is this a thing in france or only in canada?

TheoChevalier:
Yes, we must use non-breaking spaces in this context.
I believe removing them is specific to fr-CA.

You can learn more in our international French styleguide https://github.com/mozfr/besogne/wiki/Guide-stylistique-pour-la-traduction#Rgles_typographiques

missuniverse:
Closing pull request, seems like it was my mistake.

PR #1122, opened 2018-06-19
other sentences not in French

jeanbaptisteb:


mikehenrty:
good catch @jeanbaptisteb

PR #1135, opened 2018-06-22
Correct typos in the corpas-beag

kscanne:
Correct typos that will lead to problems when users record these sentences.


mikehenrty:
we need to remove the irish corpus for now due to permission issues. let's re-examine when we get the permission (cc @jimregan )

jimregan:
Sorry about this; if anything did get recorded, I have assurances that that's ok, but they need to review and discuss before committing to releasing the thing as a whole. 
Other sentences I had selected came from copyright-free sources; I'll add provenance information to the next pull request.

mikehenrty:
That's good to hear. There were very Irish recordings anyway (40 or so), so I don't think we would have lost much anyhow. Thanks for getting Irish up and running @jimregan !

PR #1136, opened 2018-06-22
Fix more typos in Irish sentences

kscanne:
Important fixes which should improve quality of submitted recordings


mikehenrty:
we need to remove the irish corpus for now due to permission issues. let's re-examine when we get the permission (cc @jimregan )

jimregan:
There are no permission issues with this set of sentences; they are drawn from speeches given by the president of Ireland: per [http://www.president.ie/en/about/terms-and-conditions](president.ie) "The President’s speeches and remarks are a matter of public record and as such are copyright free."

mikehenrty:
Oh, my mistake! Care to review this @jimregan ? We'll merge on your approval

jimregan:
Hmm... it's more that Kevin's changes are the review, so they are more than welcome. I was a bit too trusting of the source (you'd think the president's speech writers would use a spellchecker, right?). So yes, please merge!

jimregan:
@mikehenrty - could you merge this, please?

PR #1139, opened 2018-06-23
Fix a lot of french sentences

Booteille:
Fix a lot of spelling mistakes and removed a ninja sentence duplication.

### DIFF
--- a/server/data/fr/ofis_publik_ar_brezhoneg.txt
+++ b/server/data/fr/ofis_publik_ar_brezhoneg.txt
@@ -37,7 +37,7 @@ il est parti.
 elle est partie.
 j'aime aussi ce livre.
 j'aime ce livre.
-la couleur qui me plait le plus est le jaune.
+la couleur qui me plaît le plus est le jaune.
 je mange souvent des pâtes.
 ces gens-là sont souvent en train de se plaindre.
 j'avais souvent froid durant l'hiver.
@@ -99,7 +99,7 @@ prénoms et nom du futur époux.
 prénoms et nom de la future épouse.
 Audit annuel des fermes pour le lait, des fromageries pour le lactosérum.
 Nous avions eu peur.
-Licencié : Personne autorisée à prendre part aux compétitions des fédérations sportives.
+Licencié : Personne autorisée à prendre part aux compétitions des fédérations sportives.
 des messieurs fort arrogants.
 approuver la proposition de bail des nouveaux bureaux de Rennes.
 la page 44.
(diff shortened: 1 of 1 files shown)


mikehenrty:
@lissyx could you take a look?

server/data/fr/ofis_publik_ar_brezhoneg.txt
```diff
@@ -901,7 +901,7 @@ dans ton bureau.
 dans tes mains.
 dans ta maison.
 dans ta main.
-nous avons besoin de bénévoles prêts à nous aider ou à nous proposer de nouvelles idées pour organiser des évènements de qualité.
+nous avons besoin de bénévoles prêts à nous aider ou à nous proposer de nouvelles idées pour organiser des événements de qualité.
```
lissyx: évènements est également valide

server/data/fr/ofis_publik_ar_brezhoneg.txt
```diff
@@ -1104,7 +1104,7 @@ elle, elle lirait.
 elle, elle lira.
 elle, elle a été aimée.
 elle, elle est aimée.
-elle a l'esprit affuté.
+elle a l'esprit affûté.
```
lissyx: Both are valids

Booteille:
Oh, that looks like something Libreoffice did alone, sorry.

zecakeh:
While we are at it we should fix the missing capitalized first letter of the sentences.
Also there are a lot of duplicates (and I don't mean the sentences where only one word is changed, real duplicates). They are really hard to miss since they are usually grouped together.

lissyx:
@zecakeh I'm not sure it's really a big deal for the lack of capitalization. Duplicates should be easy to find with `sort | uniq` call

Booteille:
I wanted to concentrate on the first draw only on wrong typos but if someone have time to apply another fix, feel free to do it! For my part, i'll not have the time for next days.

lissyx:
I'm also busy right now, I might also be able to have a look at that in the next days though.

drzraf:
mentioning #1191

zecakeh:
@lissyx I'm sure it's not a big deal for the AI, however it feels nicer when you're recording/validating (for me anyway).
I'm looking at the file and am planning to be done by tomorrow or monday.

zecakeh:
According to #1191 (if I got it right), we're not supposed to remove duplicates because they want to keep the same sentences between the french version and the breton version. That's a major setback for me since a the first part of my work was removing those duplicates.
So I won't do anything until I know what is supposed to be corrected.

lissyx:
@zecakeh Maybe, but then the original codebase should be fixed instead of just this output file, otherwise the next import might overwrite that.

mikehenrty:
> @mikehenrty I think I read somewhere that we are supposed to spell the numbers instead of keeping the numeric form. Am I wrong ? Also I guess there's no point in keeping the abbreviations like n° for "numéro" or cf for "confer".

Exactly right, spelling out the numbers is preferred, and best to expand all abbreviations.

mikehenrty:
> According to #1191 (if I got it right), we're not supposed to remove duplicates because they want to keep the same sentences between the french version and the breton version. That's a major setback for me since a the first part of my work was removing those duplicates.

Thanks for flagging this. We have decided that you can remove duplicates. See my comment for more info: https://github.com/mozilla/voice-web/pull/1191#issuecomment-401748267

lissyx:
@zecakeh About spelling numbers and abbreviations, this is stuff I already dealt with in https://github.com/lissyx/commonvoice-syceron/tree/master/, any contirbution is welcome to make that a more generic, more robust tooling.

Anyone interested in helping contribution on french should check https://github.com/mozfr/besogne/wiki/Common-Voice-Fr

zecakeh:
Added a PR on @Booteille 's fork with all the fixes I mentioned before : [Fix and cleanup all sentences](https://github.com/Booteille/voice-web/pull/1).
I took @lissyx 's good advice to keep valid but unusual spellings.
@drzraf 's modifications in PR #1191 are also included in my commit.

From what I see we're gonna have conflicts with some merged PRs.

@lissyx Good to know. It's hard to know where all the informations are since they are not centralized. Also I might have a look at your tool one of these days, looks promising!

lissyx:
@zecakeh Please note that Breton's originated strings are actually from @ftyers, and not my work :-). Can someone send an updated / rebased PR ?

mikehenrty:
fixed in #1254

PR #1140, opened 2018-06-23
Updated spacing for easier editing, added some lines

MrLaaky:


mikehenrty:
@Gregoor are we able to make such changes outside of Pontoon? Is is best practice? Should we ask @MrLaaky to work inside pontoon instead?

Gregoor:
@mikehenrty @MrLaaky yes, those changes should be made inside of [Common Voice's Slovenian Pontoon](https://pontoon.mozilla.org/sl/common-voice/).

PR #1162, opened 2018-06-26
Add Slovenian sentences

MrLaaky:


mikehenrty:
great @MrLaaky, are these reviewed and ready to go?

MrLaaky:
I went through the sentences twice and tried my best to make it mistake free while following the guidelines. However, I am not immune to mistakes, so there might be a mistake here and there, but not to my knowledge.
Kind regards

PR #1175, opened 2018-06-29
add sentences for zh-TW

irvin:
most are came from common voice sprint


mikehenrty:
i'm so happy to see this @irvin !

PR #1177, opened 2018-06-29
add Lower Sorbian to locales

jimregan:


mikehenrty:
hi @jimregan, no need to manually do this. we have tools that import these new locales when they are ready :)

jimregan:
Fair enough

mikehenrty:
if you're interested, it's here: https://github.com/mozilla/voice-web/blob/master/scripts/import-locales.js

Issue #1049, opened 2018-06-02
How to best deal with corrections with regards to train/validation/test split


jf99:
According to #806 the data will be split into disjunct sets, so that
- each reader is in no more than one set and
- each sentence is in no more than one set (focus of this issue).

I don't know how this is currently implemented, but it might be worth to spend some thoughts on it. Essentially, I see two possible approaches:
- 1.) Split the sentences based on line number. E.g. numbers ending on 9 are test set, ending on 8 is validation set and the like.
- 2.) Compute a hash value out of the sentence string and use it to separate the sets.

The choice for either one of the possibilities has consequences when it comes to editing the sentence database. Obviously, despite all proofreading, belated corrections will be necessary here and there. If you choose possibility 1, then it must be prohibited to insert or delete lines before existing ones. Otherwise the following sentences would move to another dataset.

On the other hand, if you choose possibility 2, even a minor correction (e.g. inserting a comma) might assign the corrected sentence to a new dataset. As a result, you would have the (almost) same sentences in two different datasets. To get around this problem, one could delete all recordings based on the uncorrected sentence. However, a drawback of this measure would be that even minor corrections could destroy a hundred of valuable recordings. So at least the corrector should be warned.

Is there a better solution than the proposed?

mikehenrty:
@Gregoor When we do a sentence correction (ie. modify a sentence line), what happens to the old sentence? Is it deleted from the DB?

Gregoor:
@mikehenrty only unused sentences are deleted. Hence, if we do a correction of a sentence for which there is a clip, it'll keep existing and a new sentence will be added (which could indeed put that sentence into a different bucket).

Gregoor:
@jf99 if the sentence doesn't change, it'll stay in the DB (basically untouched).

mikehenrty:
> Hence, if we do a correction of a sentence for which there is a clip, it'll keep existing and a new sentence will be added

@Gregoor, will the existing sentence be marked as "inactive" so that we no longer record new versions of it?

mikehenrty:
> new sentence will be added (which could indeed put that sentence into a different bucket).

This is not ideal, but fine I think. We have lots of sentences that are *close* phonetically, but we still bucket randomly, so unless we implement some sort of bucketing heuristics around phoneneme's, we'll take some "close" sentences in different buckets (so long as there are not too many).

Gregoor:
[...]

PR #1191, opened 2018-06-30
[lang:fr] grammar error

drzraf:
By the way, the formulations for this set of sentences is a bit unusual and vague because speaker may like interpret it as a question "are you sure we liked?" rather than "you're sure we liked".
(@ftyers)

### DIFF
--- a/server/data/fr/ofis_publik_ar_brezhoneg.txt
+++ b/server/data/fr/ofis_publik_ar_brezhoneg.txt
@@ -2509,7 +2509,7 @@ tu es sûr qu'elle aima.
 tu es sûr qu'il aimait.
 tu es sûr qu'elle aimait.
 tu es sûr que vous aimiez.
-tu es sûr qu'on aimaient.
+tu es sûr qu'on aimait.
 tu es sûr que nous aimions.
 tu es sûr que j'aimais.
 tu es sûr qu'ils aimaient.
--- a/server/data/fr/ofis_publik_ar_brezhoneg.txt
+++ b/server/data/fr/ofis_publik_ar_brezhoneg.txt
@@ -4582,8 +4582,8 @@ On l'a frappé.
 Il l'a frappé.
 Il a été frappé.
 J'ai été frappé.
-J'ai été frappé par leurs opinions tranchés.
-Ils ont été frappé par la crise.
+J'ai été frappé par leurs opinions tranchées.
+Ils ont été frappés par la crise.
 La crise les a frappé.
 La crise les a durement frappé.
 Ils ont été durement frappé par la crise.
--- a/server/data/fr/ofis_publik_ar_brezhoneg.txt
+++ b/server/data/fr/ofis_publik_ar_brezhoneg.txt
@@ -29,7 +29,7 @@ c'est les Kurdes qui ont gagné à Kobané.
 les Kurdes ont été victorieux à Kobané.
 c'est le gars de Carhaix qui a gagné.
 le gars de Carhaix a été victorieux.
-elles est partie.
+elle est partie.
 il est parti en mer.
 il est parti.
 elle est partie.


drzraf:
Another possibly confusing one is: "Je suis en train d'écouter. (téléphone)."

drzraf:
That file is **really** of a very bad quality. Grammar, tipo, unknown words, duplicated sentence. WTF?
Can't believe Mozilla (and us the french speakers) could not have been provided with a better set!
@mikehenrty (#1095), why such a set?

drzraf:
Worth noting that any change to a line create a new hash mapping (sha256|b64encode) basically disgarding the work of speakers on the former form the of sentence (!)

ftyers:
Hi @drzraf the sentences come from a collection of parallel sentences in the public domain released by Ofis Publik ar Brezhoneg in Breton and French. The reason we added them was because we had them and they were in public domain. I did not personally check through each of the sentences, but an eyeball glance showed that they looked ok. Personally I expected higher quality, but when you are automatically extracting things there are always likely to be problems with data processing. The most important part was the Breton side of the data, the French part was added as an after thought.

We would be more than happy to make adjustments to the sentences, you are welcome to send a pull request to fix them them. You are also welcome to refer to specific sentences here or in an issue and I will make the edit and pull request. Note that we would like to keep the same order and number of sentences as in the Breton version of the file. This could prove to be a useful dataset for future work on Breton--French speech-to-speech translation.

In addition you are also very welcome to propose your own sentences or collection of sentences. We are always happy to receive more data. In the web interface you may also "skip" a sentence if you do not like it. I expect that this may be used in the future as a method of quality control.

Thank you for your feedback, it has been very helpful.

zecakeh:
@ftyers So from what I got, we're not supposed to remove duplicated sentences from the french version ? After a quick look at the breton version, they also have duplicated sentences, but that would require working on both files at the same time.

I understand the need for sentences to start the project in a language, but I think the validation process should be more thorough.

ftyers:
@zecakeh in an ideal world the sentences would remain aligned. However, given people's strong feelings I don't see much of a problem with removing them. I have the original spreadsheet and can always do a string-for-string lookup to find the translations.

I'd like to apologise to all French contributors for the inadequate proofreading of the sentences before I submitted them. In the future I will endeavour to do a better job.

drzraf:
> Note that we would like to keep the same order and number of sentences as in the Breton version of the file. This could prove to be a useful dataset for future work on Breton--French speech-to-speech translation.

This is a terrible restriction put on both corpus evolution (and particularly the french one), and it's not clear to how maintaining such mapping would be useful.

About dataset preservation when change occurs on the sentence: #1049

About About alternatives corpus, a lot of languages laboratories all around Europe already worked on this for years (usually with a proven methodologies):
* http://wortschatz.uni-leipzig.de/en/download . A couple of sample files to explore:
  * fra_wikipedia_2010_100K-sentences.txt
  * fra_mixed-typical_2012_100K-sentences.txt
* https://deep-sequoia.inria.fr/fr/ (sequoia-8.1/sequoia.surf.conll)
* http://www.atilf.fr/spip.php?article915 (Not available, but could be requested)
* http://www.cnrtl.fr/ probably contains some good ones
Tagging a couple of GH users working with french corpus @prismsoundzero @bsagot @simonmeoni @kimgerdes @gguibon @mmcauliffe

mikehenrty:
Thanks everyone for the great discussion here! A couple of thoughts.

Let's forget about trying to keep translation alignment with the Breton set. It was a good starting point to have both corpora, but at this point I would rather focus on fixing the issues individually than trying to keep the alignment. There are other projects out there for translations, Common Voice is for Speech-to-text data.

Yes, spelling numbers is indeed better than having the numbers listed. It's ok for single digits, but when you get to things like: `2018`, there are several ways to say that. Spelling it out is better, `two-thousand eighteen` or `twenty eighteen`.

Indeed I would love to see more text added to our corpora. But I cannot read French (or most of the languages on Common Voice), so for now we are relying on you the community to flag and address issues. I'm just here to push the merge button when things are ready (and at some point we will give access to other members to do that instead of me). So please continue to have this discussion! And remember, we are all on the same team :) https://github.com/mozilla/voice-web/blob/master/CODE_OF_CONDUCT.md

Please tag me when these fixes are ready to merge. Thanks!

lissyx:
BTW, anyone interested in contribution on french sentences should have a look at https://github.com/mozfr/besogne/wiki/Common-Voice-Fr

lissyx:
@mikehenrty I guess this is now ready to be merged ?

PR #1196, opened 2018-07-01
French places names / Sentences

Lulucmy:
Hi,

I made a list of French sentences (for books of the 18th and 19th centuries) about French places.
Some are from Jean-Jacques Baude, Victor Hugo, or Théophile Gautier.
I will add some other sentences from other books if I find good ones.


mikehenrty:
Thank you for the contribution @Lulucmy !

two quick thoughts:

1.) Can we make sure there are no sentences over 14 words? We need to try and keep the sentences around 5 seconds to speak
2.) Can we spell out the numbers (ie. `fifteen twenty-four` rather than `1524`). This makes it less ambiguous

Lulucmy:
Yes we can spell all the numbers, I changed everything for this, and there is no very long sentences.

mikehenrty:
thank you @Lulucmy !

PR #1228, opened 2018-07-08
Create corpus-litterature-part3

belkacem77:
Daily language: social activities winthin the kabyle villages, daily expressions


belkacem77:
Tahnks to @S-Amiri @Amazigh-bedar . Tha same sentences are also available on https://github.com/belkacem77/tuddar

Issue #912, opened 2018-04-06
CSS Sanity Restoration


Gregoor:
We considered adding a component styling library, like styled-components, but with those we have to pay the extra size-price (between 8kb-14kb, depending on the lib). So, a better approach might be to take what we already have to a better place. Even if we make the decision to switch to one of those libs after, the effort wouldn't be in vain, since cleaner DRYer styles are easier to port.

So this is a meta-issue for tracking everything that we can do to make our styling saner. Discussion, other ideas and PRs very much wanted & appreciated!

- [x] Standardise
  - [x] Margins/Padding (atm we probably have >~ 10 different values) 
  - [x] Font Sizes (atm >~ 8 differnet values)
  - [x] Colors
  - [x] Transitions
- [x] Prefer nesting over selector repetition ([example](https://github.com/mozilla/voice-web/commit/411bd75a57ac368bcf44a0f697f5137fca3ba0c0))
  - While deep nesting is hard to read, I prefer it over the more error-prone selector repetition (also deep nesting being difficult to read, might incentivize componentizaation)
- [x] Bootstrap-like mobile-first breakpoints ([example](https://github.com/mozilla/voice-web/commit/6415f246b6a42c59a610fcf27fff38e7bbf73d05))
  - When we're done, we can delete the [old breakpoints](https://github.com/mozilla/voice-web/blob/6415f246b6a42c59a610fcf27fff38e7bbf73d05/web/src/components/vars.css#L63-L67)
- [x] Require CSS files in their respective component js files
- [ ] Styleint
  - [ ] order of attributes (imo it should be "position > display > margin > padding > size > color/opacity/etc. > transition > sub-selectors")
  - [ ] forbid new values for colors etc (to enforce standardisation)
  - [ ] run sth like `stylelint --fix` with our `prettier` task
- [ ] stop depending on route name for css classes
  - in `layout.tsx` we set the content's classname to the route and depend on those names all over the place
- [ ] don't select for tags or specific components. give context specific classNames

PR #1237, opened 2018-07-09
Use nesting for CSS

BynariStar:
Hi! First-time contributor here. Thought I'd pick CSS nesting from #912 as a simple starter task.

Let me know if there are any issues or if I missed any files.


Gregoor:
That's awesome, thank you!

PR #1243, opened 2018-07-11
Extract French parliaments debates strings

lissyx:


lissyx:
As I worried, some strings where modified between the extraction made, on the crowdin space, so it makes it very hard to precisely map back the data, and thus go back to the root cause of a bad string.

lissyx:
Yes, that's it. IMHO, we don't need thorough verification on that, just another pair of eyes to make sure. If you tell me it looks okay, we'll merge it: some strings have been changed from the original source, as I mentionned earlier, so going back to the real original file will be painful.

lissyx:
@mikehenrty Looks like I made no big mistake, this is good enough to be merged.

PR #1249, opened 2018-07-12
Update corpus-litterature-part1.txt

belkacem77:
remove 2 sentences due to urban language


mikehenrty:
@belkacem77 i just see one sentence removed, not two. is that on purpose?

belkacem77:
Sorry, the first is already removed.

PR #1259, opened 2018-07-12
Automatically add hyphens on contribution cards

Djfe:
Beware: I tested this only on voice.mozilla.org by using the inspector.

Before:
firefox_2018-07-12_23-52-34

After:
firefox_2018-07-12_23-52-39

What do you think?

https://caniuse.com/#feat=css-hyphens
https://developer.mozilla.org/docs/Web/CSS/hyphens

### DIFF
--- a/web/src/components/pages/contribution/contribution.css
+++ b/web/src/components/pages/contribution/contribution.css
@@ -259,6 +259,9 @@
             box-shadow: 0 6px 12px 0 rgba(0, 0, 0, 0.05);
             overflow: auto;
             word-wrap: break-word;
+            -ms-hyphens: auto;
+            -webkit-hyphens: auto;
+            hyphens: auto;
 
             opacity: 1;
             transition: transform var(--transition-duration-slow) var(--easing),


Djfe:
I'm not so sure anymore :/
firefox_2018-07-13_00-00-35

mikehenrty:
@Djfe is the idea to save space? @Gregoor do you have any thoughts?

Djfe:
the idea was, that some words are long enough to fill almost a whole line without this (vierzehnjährigen)

with this enabled, other words could fill up remaining space, so that there are less lines

Djfe:
for very short sentences that might look unnecessary though

Gregoor:
I like this idea, didn't know about that attribute. Awesome! Maybe we could conditionally add a class (like `hyphenate`) when a sentence is long enough?

Djfe:
Sounds good.
Maybe 50 characters works well (or just 40).
I'm ready to give it a test on allizom :)

jf99:
I think the longest word I've seen so far is vierundvierzigtausendeinhundert, which is too long for a single line. So you probably want to check not only for the length of the sentence, but also for the length of the longest word in it.

Djfe:
I just realized, that hyphens uses the lang attribute of the html tag as the reference for the language:
https://blog.kulturbanause.de/2012/02/silbentrennung-im-browser-css-eigenschaft-hyphens/
Without it, it will screw up. Common Voice already implements it correctly. I just wanted to mention it here, because I didn't know about it, when I tested it out and wrote the PR.

@jf99 Words that are too long for the div could be hyphanized by:
https://www.w3schools.com/cssref/css3_pr_word-wrap.asp
(I just realized that the div already has that attribute hehe)

The specs:
https://drafts.csswg.org/css-text-3/#hyphens-property
https://drafts.csswg.org/css-text-4/#propdef-hyphenate-limit-zone

The hyphenate limit zone/limit chars (maybe also limit lines) properties seem to implement just what we wanted here :)
no Javascript needed 😍
it's still a draft and only IE supports it 😞
https://bugzilla.mozilla.org/show_bug.cgi?id=css-text-4
https://msdn.microsoft.com/en-us/library/hh771869(v=vs.85).aspx

So we might still need a workaround using js or we wait instead.

The easiest thing might be to add hyphens: auto when there's is a word with more than x chars or when the sentence is over x chars long.

Gregoor:
Nice to see that we anticipated what's being drafted :)

MichaelKohler:
What's the status here?

phirework:
So I definitely don't want a heavy JS-based solution that checks word-length and hyphenates based on that, and even checking each sentence feels like overkill. My instinct is that hyphenated short words will be more confusing than the occasionally line-dominant long word, in which case I'd prefer to close this without merging. What do you think?

Djfe:
I don't care too much anymore ^^
It's fine by me, whatever you choose to do.

imo the outcome with well-placed hyphens looks nice and it would've been nice to implement it with a simple css-based approach here but css text 4 is still a draft so it probably won't be possible for a while.

The code-base probably changed quite a bit in the meantime anyway so a simple merge won't work.
I had a lot of fun, testing with the attribute but I see why it's not quite feasible, yet, to implement it in this way.
And it looks ok without this change.

PR #1262, opened 2018-07-15
Create copus-litterature-part7.txt

belkacem77:
Own sentences from non published texts I wrote, daily expressions.


belkacem77:
Thanks @mikehenrty if you think that it's better to send long files corpus let me know. I'm splitting them into parts of 200 and 500 sentences.

mikehenrty:
splitting them is no problem. whatever works best for you. thanks @belkacem77 !

PR #1264, opened 2018-07-16
Add some LibreTheatre data

lissyx:
Regenerate with:

for f in $(wc -l data/libretheatre/*.txt | sort -nr | head -n 50 | grep txt | awk '{ print $2 }');
do
	F=$(basename `echo $f`);
	T="libretheatre_$F";
	head -n 55 $f > $T;
done;


lissyx:
@mikehenrty I guess we should be able to merge that

mikehenrty:
@lissyx, this one still has some `--` and `...`, which I think we want removed from the dataset. what do you think?

lissyx:
@mikehenrty I should have a fix now, regenerating data

PR #1282, opened 2018-07-21
[fr] 39 phonemes

drzraf:
Aside from words/sentences recognition, phonemes recognition is an interesting field having its own set of possible applications (eg, when it comes to learning language/letters/phonemes and pronunciation or overcoming some related diseases...) [although we probably lack children voices]

There is already a couple of scientific [corpus split on the phoneme level](https://www.ortolang.fr/market/corpora?filters=%7B%22corporaType.id%22:%5B%22speech_corpora%22%5D,%22annotationLevels.id%5B%5D%22:%5B%22phonetic_transcription%22%5D%7D&viewMode=tile&orderProp=rank&orderDir=desc) for French language.

Also not a specialist in this field, but I _wonder_ whether creating an additional audio corpus of individual phonemes would provide useful. (Note: in that case, how could voice-web hint users with context: Eg: `Please only pronounce the sound of `**[l]**` (like in "lit", "halle" or "allée")`?

References:
https://github.com/mozfr/besogne/wiki/Common-Voice-fr
https://fr.wikipedia.org/wiki/Liste_des_graphies_des_phonèmes_du_français
https://github.com/persephone-tools/persephone
@alexis-michaud @hadware @PaulBoersma

### DIFF
--- a/server/data/fr/phonemes.txt
+++ b/server/data/fr/phonemes.txt
@@ -1,0 +1,43 @@
+# voyelles
+[i] nid	jeep	abbaye	Annie
+[e] dé	écolé	pied
+[ɛ] mais	frais	être	parfait
+[a] laver	ami	tuba
+[ɑ] âme	âtre	pâte
+[ɔ] dehors	bol	Victor	Paul
+[o] hôte	osé	beau	tôt	pot
+[u] où	vous		coût	loup
+[y] jus	sur	Ursule
+[ø] jeu	feu	veux	Denis	
+[œ] œil	fleur	beurre	seul
+[ə] venin	monsieur
+[ɛ̃] vin	train	pain
+[œ̃] brun	humble	parfum	lundi
+[ɑ̃] an	ampoule	encre
+[ɔ̃] on	ombre	tronc	maison
+# semi-consonnes
+[j] yaourth	crayon	sommeil
+[w] oie	oui	moi	whisky
+[ɥ] aujourd'hui	huit	huile
+# consonnes
+[p] pipe	cap	appel
+[t] tarte	patte	théorie
+[k] casse	sac	paquet
+[b] bal	bébé	abbé
+[d] monde	raid	midi
+[g] gaffe	tag	bague
+[f] vif	photo	fou
+[s] si	casse	nation
+[ʃ] chat	tache	achat
+[v] va	cave	avril
+[z] zèbre	base	dixième
+[ʒ] jeu	tige	âgé
+[l] lit	halle	allé
+[ʀ] rat	car	carré
+[m] Mamie	âme	homme
+[n] nous	âne	tonne
+[ɲ] gagner	peigne	vigne
+# non-natives
+[h] hop!	hue!
+[ŋ] camping	viking	swing
+[x] (jota)	arthrose	


hadware:
Not entirely sure I understand why I've been summoned to this discussion, but the SAMPA phonetic alphabet is maybe a good start (and it's pretty stardard as well). It's used in Mbrola for instance. Here is is for French https://www.phon.ucl.ac.uk/home/sampa/french.htm

alexis-michaud:
Hi! That's a topic I'm interested about. Thanks for the 'ping' and feel free to ask more. Here are some thoughts. 

- About phoneme inventory: Yes, it's definitely good to have a set of recordings that match phoneme inventories. The approved order of business is to identify a speaker, work out her/his phonemic system, and _then_ do recordings. So there will be a good match between the audio and transcription. It's a terrible shame when someone asks a speaker to do a recording and then pairs the recording with a pre-defined transcription that does not match the transcription, as when a textbook reports a [/ɑ/ vowel](https://fr.wikipedia.org/wiki/Liste_des_graphies_des_phon%C3%A8mes_du_fran%C3%A7ais#/%C9%91/) distinct from /a/, and provides recordings in which the two phonemes are pronounced in the same way (this contrast is lost in contemporary 'standard' French, though some people can still produce a difference, at a push). That's an issue with a new recording campaign conducted online: it is misleading to assume the same phonological system in all speakers. Speech data collection is an underestimated challenge: some views on this topic are expressed [here](https://halshs.archives-ouvertes.fr/halshs-01026295v4/document)!

- About phoneme recognition as a research field: 
I think that specialists of Automatic Speech Recognition for French are by now accustomed to doing recognition with a language model, recognizing bigger units (words, sentences...) not phonemes, and they may find it strange to think of using a 'standalone' phoneme recognition tool like [Persephone](https://github.com/persephone-tools/), because current tools for widely-used languages like French yield output (complete sentences) that is of so much more practical usefulness. But phoneme recognition remains a worthwhile research topic, I would think, for instance to identify 'noncanonical' realizations of phonemes.

- About children's voices: 

> [although we probably lack children voices]

The issue of how far it is possible to do automatic phonemic transcription of children's voices has been raised by Pr. Birgit Hellwig (Köln=Cologne), with special reference to a vast corpus of children's voices from Papua New Guinea. It would indeed be good to find out from what age this is feasible. My own guess would be that this is basically _not_ possible with toddlers because very young learners' phonological systems are in a state of flux (but that would nonetheless be worth testing). 

Corpora of children exist but they are usually not under a CC-0 license because of privacy & other issues. It should be fine running stats on these corpora to create acoustic models or language models for Natural Language Processing, but it's problematic to let random humans browse the files (audio/video). Maybe the solution to work out with corpus copyright owners would be to allow 'data mining' for Natural Language Processing purposes without allowing open access to the files. But the fact that Google & others are reported to have broken some promises they made about not using data for other purposes than those listed in agreements (for instance, when dealing with e-mail correspondences from students at universities) could make data curators wary of opening their collections to this kind of mining.

- I don't work on French and would recommend that you get in touch with colleagues working in the 'PFC' project, [Phonologie du Français Contemporain](https://modyco.fr/fr/optranscorpus/pfc.html).

mikehenrty:
Thanks for the discussion.

That said, I'm not exactly sure what this PR is trying to do. What will happen now is that a user will be presented with:
```
[a] laver	ami	tuba
```
and will have to read that. How does one read `[a]`? This is not a nice user experience. What are you hoping to achieve here?

Gregoor:
I'm closing this due to inactivity. Happy to reopen if wanted.

PR #1285, opened 2018-07-22
Normalize audio for verification

### LINKED ISSUE #336: Adjust volume automatically when listening to multiple recordings
Many music players can do this, why not use it here too?

Just adjust the audio pitch when hearing to sentences. I always have to adjust the volume up and down and by using such an automatically adjustment I could prevent this.

est31:
Implement audio normalization on the client, using a method based on ReplayGain.

Fixes #336

### DIFF
--- a/web/src/components/listen-box/listen-box.tsx
+++ b/web/src/components/listen-box/listen-box.tsx
@@ -7,6 +7,7 @@ import StateTree from '../../stores/tree';
 import { FontIcon, OldPlayIcon, OldRedoIcon } from '../ui/icons';
 import URLS from '../../urls';
 import { LinkButton } from '../ui/ui';
+import NormalizedPlayer from '../normalized-player';
 
 const VOTE_NO_PLAY_MS = 3000; // Threshold when to allow voting no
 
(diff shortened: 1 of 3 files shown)


est31:
@Gregoor please take a look.

I've tested the changes locally. They do work.

est31:
How can Travis CI be fixed?

est31:
Caniuse [says](https://caniuse.com/#search=webaudio) that all of the [supported browsers](https://github.com/mozilla/voice-web/blob/master/docs/browser-support.md) should have WebAudio.

est31:
Also note that this needs [changes in the S3 bucket configuration](https://docs.aws.amazon.com/en_us/AmazonS3/latest/dev/cors.html#how-do-i-enable-cors) to enable CORS requests from voice.mozilla.org.

Gregoor:
I'm getting CORS warnings when trying it out, we might have to add CORS headers allowing the AWS domains.
Another thing that came to mind is that we would also want to use this for the listen widget on the homepage as about 40% of all listens still happen there.

est31:
@Gregoor the CORS settings need to be adjusted in the Bucket, not on the side of the website. See the page I've linked above. As for the listen widget, sure will do.

Gregoor:
Oh, pardon read over that. Thanks for the pointer!

Gregoor:
Works great for me locally! Thanks for also bringing it to the old widget. I'm awaiting the config change of the staging/production buckets and then I'll merge it!

est31:
Rebased, as #1337 caused conflicts.

Gregoor:
Thanks for keeping it synced! I was told the infra change is next up on the list of todos, I hope that means soon 🙃

lissyx:
@est31 @Gregoor When can we have that live? It would be such an awesome improvement!

Gregoor:
Thanks for the reminder. Full plate this week 🙈 Let's take this live!

Gregoor:
@est31 I reverted this for now (eedf104347c80bc6fbe86b51f298fe290d054f13), as it broke Safari (#1463) and I couldn't figure out how to unbreak it. What I found out: 
Safari doesn't properly implement decodeAudioData / or doesn't implement the new Promise based version of it. But even when I used the old callback syntax, it wouldn't work.

est31:
Oh sorry to hear that it broke Mac OS. I don't have an apple device so can't be of much help, but I've found [this](https://github.com/mohayonao/promise-decode-audio-data) library. @Gregoor could you maybe try whether using it fixes the issue?

PR #1301, opened 2018-07-26
added german sentences with counties capitals and more geographical names

ubahnverleih:
New german sentences:
Mainly added sentences of countries and their capitals. Also some more international places (Regions, cities, rivers, mountains, oceans, planets, ...) and a few travel related sentences.


mikehenrty:
thanks @ubahnverleih !

PR #1326, opened 2018-08-07
[fr] removing antiquated and racist/offensive words from some French sentences

jeanbaptisteb:
Removing the words "nègre" and "négresse" which are antiquated and racist words in France's French. See https://www.quora.com/How-offensive-is-%E2%80%9Cn%C3%A8gre%E2%80%9D-in-French . I replaced them with non-offensive words, or rewrote or removed the whole sentence when it was not possible. These offensive words comes from old theater plays, which were automatically extracted from Wikisource; it explains how they got there.


lissyx:
While this is technically valid french wording, I get that until we have a way to perform contributions with a flag allowing for offensive content, we might want to remove that. Though @jeanbaptisteb I'd prefer we fix that on the `commonvoice-fr` repo and then update the dataset here: this way, we avoid the risk of regressions when updating / generating new data.

jeanbaptisteb:
@lissyx OK no problem!

PR #1336, opened 2018-08-09
Add Esperanto

bidaian:
I'm not sure we (eo) are ready for production, as contributable language. If we are, then I guess we should be listed here.

### DIFF
--- a/locales/contributable.json
+++ b/locales/contributable.json
@@ -5,6 +5,7 @@
   "cy",
   "de",
   "en",
+  "eo",
   "fr",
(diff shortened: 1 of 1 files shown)


Gregoor:
Hi @bidaian, thanks for the Esperanto contributions. Those locale JSON files are generated automatically. And a language is automatically added to the contributable-ones, once it has at least 1k sentences. Esperanto is currently at ~300.

(Sorry for the delayed response, and thanks for the nudge, Lucy)

PR #1341, opened 2018-08-11
more corrections in German

jf99:
Not so sure about the year numbers. In other files, they are written without hyphens.
neunzehn-hundert-achtzehn vs. neunzehnhundertachtzehn

Sure, it's easier to read with hyphens, but I think without them is the formerly correct way. Also, we should probably have a consistent way to write such year numbers.

Shall I remove the hyphens?


jf99:
Pushed another round of corrections and removed the hyphens. @Gregoor Ready to merge.

PR #1355, opened 2018-08-14
Add Capitalize...Ending punctuation.

anzart:
Capitalize the first word and end the sentence with appropriate ending punctuation.


kdavis-mozilla:
Depending on the language correct capitalization may be required. For English it's not strictly necessary.

However, the data set is of more use when a capitalization is consistent and correct.

mikehenrty:
@kdavis-mozilla great thank you. what about punctuation? should we be adding this to sentences?

@belkacem77 do you know if capitalization is required for Kabyles? Or is it optional like in English?

belkacem77:
Capitalization is optional. It doesn't change the sense, phonology, grammatical class .....

PR #1356, opened 2018-08-14
Add Capitalize...Ending punctuation.

anzart:
Capitalize the first word and end the sentence with appropriate ending punctuation.

### DIFF
--- a/server/data/kab/named-entities1.txt
+++ b/server/data/kab/named-entities1.txt
@@ -1,304 +1,304 @@
-adabu uswaḍ n usliqqel
-aɣerbaz amenzu rabeḥ belqasem
-aɣlif amezwaru
-aɣlif n ṣṣehd
-aɣlif n tdawsa d yimezdaɣ d useggem n sbiṭrat
-aɣlif n tedrimt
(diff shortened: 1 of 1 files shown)


jf99:
It might be a good idea to wait with such mass editing only changing capitalization and punctuation until a new contribution experience is online. #1049 What do you think?

mikehenrty:
@jf99 what would we be waiting for? im not sure how long or what the resolution will be from the discussion in #1049

jf99:
@mikehenrty I had the hope that bucket-aware correction might become part of the new sentence contribution process. Well, you know better than me, if this is true and when it's going to happen.

IMO, the consideration should be as follows: All sentences that get modified via PR after they have already been in use are potentially in two buckets. So, to keep the data sets clean, they will all have to be put into the train bucket when the data is released to the public. The current series of PRs modifies ~70% of all kab sentences. The PRs should be merged if it looks like enough new sentences will be added soon enough to attain the desired train/dev/test ratio.

belkacem77:
Within Speech recognition, there is no difference in Kabyle between A and a. But all named entities we used are capilized,

PR #1369, opened 2018-08-16
more sentences from mansur and ilnar

ftyers:
More proofread public domain sentences from @mansayk  and @IlnarSelimcan.


mikehenrty:
awesome!

PR #1416, opened 2018-08-24
Catalan: Create cities.txt

jmontane:
Cities from Andorra, Balear Islands, Catalonia, and Valencia zones


mikehenrty:
thanks @jmontane !

PR #1436, opened 2018-09-05
WIP: Hacking some Kiosk UI

lissyx:


web/src/components/index.css
```diff
@@ -8,6 +8,7 @@
 @import url('listen-box/listen-box.css');
 @import url('modal/modal.css');
 @import url('pages/data/data.css');
+@import url('pages/kiosk/kiosk.css');
```
Gregoor: We transitioned to requiring css files in their related components. The files that are still required in `index.css` are the older pages (where I wasn't sure how it'd affect the cascade, as it changes the order).

web/src/components/index.css
```diff
@@ -365,6 +374,18 @@ body:not(.ios) .recording #background-container {
     }
 }
 
+@media (--mobile) {
```
Gregoor: Those breakpoints are also old, the newer pages use more granular ones (4 different sizes). For more on CSS: https://github.com/mozilla/voice-web/issues/912

web/src/components/pages/contribution/listen/listen.tsx
```diff
@@ -140,6 +148,8 @@ class ListenPage extends React.Component<Props, State> {
         (clip, i) => (i === clipIndex ? { ...clip, isValid } : clip)
       ),
     });
+
+    updateKiosk({ listenFinished: true });
```
Gregoor: Why set this here and in line 189?

web/src/components/pages/contribution/success.tsx
```diff
@@ -153,12 +161,42 @@ class Success extends React.Component<Props, State> {
...
+                <h3 />
+              </Localized>
+
+              <Localized id="kiosk-listen-submit">
+                <LinkButton rounded to={URLS.KIOSK} />
+              </Localized>
+            </div>
+          )}
+
+        {!wizardFinished && (
```
Gregoor: Does that still show in non-Kiosk mode then?

web/src/components/pages/contribution/success.tsx
```diff
@@ -153,12 +161,42 @@ class Success extends React.Component<Props, State> {
           </div>
         )}
 
-        <ContributeMoreButton>
-          {type === 'speak' ? <MicIcon /> : <PlayOutlineIcon />}
-          <Localized id="contribute-more" $count={SET_COUNT}>
-            <span />
-          </Localized>
-        </ContributeMoreButton>
+        {wizardFinished &&
```
Gregoor: Still would prefer an enum for the kiosk status 🙊 Is that still on the menu?

web/src/components/pages/kiosk/kiosk.css
```diff
@@ -0,0 +1,46 @@
...
+#kiosk-container {
+    & .kiosk a {
+        font-size: var(--font-size-small);
+        font-weight: 600;
+        text-decoration: underline;
+    }
+}
+
+#common-voice-kiosk {
+    & {
```
Gregoor: Haven't seen this `& {` kind of rule declaration before. Is there a difference to directly putting the rules under `#common-voice-kiosk`?

web/src/components/pages/kiosk/kiosk.tsx
```diff
@@ -0,0 +1,200 @@
...
+}
+
+interface KioskProps
+  extends LocalizationProps,
+    PropsFromState,
+    PropsFromDispatch,
+    RouteComponentProps<any> {}
+
+class KioskPage extends React.Component<KioskProps> {
+  state: User.State = {
```
Gregoor: there's some leftover copypasta. i.e. you don't need all that user state

web/src/stores/root.ts
```diff
@@ -47,6 +49,7 @@ const store = createStore(
       sentences: undefined,
       uploads: undefined,
       user: undefined,
+      kioskprogress: undefined,
```
Gregoor: nit: I'd camelcase that

web/src/components/pages/kiosk/kiosk.tsx
lissyx: Am I ? I thought it was required to display the email field below

web/src/stores/root.ts
lissyx: Sure.

web/src/components/pages/contribution/listen/listen.tsx
lissyx: Because that's the only proper place I could find: when we vote, or when we have no clip to show. Happy to do that at some other, unique place for both case if you see where :)

web/src/components/pages/contribution/success.tsx
lissyx: It should, yes, though I'm not 100% sure we need to hide it ?

web/src/components/pages/kiosk/kiosk.tsx
lissyx: Oh, and when I was only pulling email, it would complain about missing fields :)

web/src/components/index.css
lissyx: Fixed as well.

web/src/components/pages/kiosk/kiosk.tsx
lissyx: My mistake, I can do it

web/src/components/pages/contribution/success.tsx
lissyx: Retesting, it's not that disturbing IMHO, so I'll leave it.

lissyx:
Closing this one.

PR #1443, opened 2018-09-09
Create kab-team2.txt

belkacem77:
Kabyle Sentences from Tatoeba. Reviewed.
These sentences are from the same team working also on the kab locale of Common Voice.


belkacem77:
@mikehenrty  please would you like add the file. It's from our work on tatoeba.

PR #1451, opened 2018-09-12
Create kab-team3.txt

belkacem77:
More sentences from Tatoeba. Thanks also to @Amazigh-bedar  @S-Amiri @Nacira Abrous


belkacem77:
@mikehenrty  please would you like add the file. It's from our work on tatoeba.

PR #1472, opened 2018-09-19
feature/new-profile-login

Gregoor:
- use a singleton for the mysql wrapper to simplify depending on it
- collect all user_clients for a given client_id and email
- remove old admin login
- move demographic data options into separate file


Gregoor:
@fiji-flo @akatsoulas this one got a bit too bulky in scope, sorry. I usually do try to do things in a more granual way. The luxury of working alone is that you can do drive-by refactorings in one step. I'll try to better myself!

PR #1485, opened 2018-09-23
more German language corrections

jf99:
Just 3 minor corrections.


RobinvanderVliet:
What will happen to people that already contributed pronunciations of sentences that are changed by this pull request? Their files should probably be deleted.

jf99:
According to https://github.com/mozilla/voice-web/issues/1049#issuecomment-402437632 they will not be deleted. The old recordings are stored along with the old text, and the new recordings will be stored with the new text. So the audio always fits to the text. Note that the readers are supposed to pronounce exactly what they read, not what they think would be correct.

PR #1486, opened 2018-09-24
add more sentence for zh-HK

terrylau2013:
all source written by myself , hk gov and wiki


irvin:
維基百科的授權是 CC:BY-SA，不能使用做錄音的語料

mikehenrty:
Correct, we cannot use Wikipedia. Can you remove that from this PR @terrylau2013 ?

terrylau2013:
wiki.txt deleted

mikehenrty:
thank you @terrylau2013 !

One final question, is there a license for the sentences you got from the HK goverment?

terrylau2013:
@mikehenrty , I'm not explicitly to request a license from hk gov, but most of data are from hk gov open data -- https://data.gov.hk
according to the "TERMS AND CONDITIONS OF USE"

You are allowed to browse, download, distribute, reproduce, hyperlink to, and print in their original format the Data for both commercial and non-commercial purposes on a free-of-charge basis on condition that:-

you shall comply with the Terms of Use;
you shall identify clearly the source of the Data and acknowledge the Government ownership of the intellectual property rights in the Data and in all copies thereof including but not limited to paper copies, digital copies and copies placed on other websites;
you shall indemnify the Government against any allegations or claims of infringement of the rights of any person and all costs, losses, damages and liabilities incurred by the Government, which in any case arise directly or indirectly in relation to your use, reproduction and/or distribution of the Data;
You shall also give proper attribution to the Government and DATA.GOV.HK.

mikehenrty:
> You shall also give proper attribution to the Government and DATA.GOV.HK.

Unfortunately, this is a deal breaker for us. We make a promise to our users (and to the world) that we will publish the voice data with a CC-0 license, but if the sentence source has an attribution license we cannot publish under CC-0.

In this case, I think we would need special permission from the HK government to use that data. Just to note, this is not an insurmountable challenge. We were able to obtain permission from the French government to use their transcripts fairly easily. It's just a matter of finding the correct person to ask. Unfortunately, we would need to rely on you or someone in HK to make this connection.

terrylau2013:
Got it, I can try to ask.

sammyfung:
Thanks for Irvin telling me about this issue, IMO, I think it is not easy to get a special permission from HK Gov (OGCIO and departments providing these data) for data retrieve thru data.gov.hk.

sammyfung:
@terrylau2013 Which datasets on data.gov.hk you use ? May you provide source URLs ?

mikehenrty:
>  If I base on data from data.gov.hk and complete it to full sentence, what license is it?

When you modify and re-pulish (in other words, "remix"), you still need to adhere to the original license. So if it was cc-4, your new "remixed" dataset needs to be cc-4 with attribution to the original publisher. Does that answer your question?

irvin:
Hi @terrylau2013 

Could you send my your email address?  ( irvin at moztw.org ) 
We just meet some people from data.gov.hk at Taiwan, and your example may be a good example to help them adopt more open licenses.

Gregoor:
Thanks for your efforts! We're about to launch a beta of our sentence collection tool, please submit your sentence through that once it's ready. It'll be announced in our Discourse very soon:
https://discourse.mozilla.org/t/readme-how-to-see-my-language-on-common-voice/31530/

PR #1492, opened 2018-09-25
add Hakha Chin data from working group

ftyers:
Here are the first 1000 sentences from our Hakha Chin working group at IU.

### DIFF
--- a/server/data/cnh/chin-project.txt
+++ b/server/data/cnh/chin-project.txt
@@ -1,0 +1,999 @@
+Keimah ka min cu Peng a si.
+Indianapolis ah ka um.
+Indiana University ah sianginn ka kai.
+Ka dam ve, kaa lawm.
+Khoi ka ah dah atu an um?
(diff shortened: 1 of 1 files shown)


mikehenrty:
awesome @ftyers ! do you know anyone who would be interested in localizing the website content into Hakha Chin?

https://pontoon.mozilla.org/projects/common-voice/

ftyers:
@mikehenrty already done as part of a summer project. We're working with some different organisations on this. :) They're proofreading the next 1000 sentences now and we should have 5000 in a few weeks or so.

mikehenrty:
that is so cool @ftyers ! 

as for enabling it on the main site, what would you need to test? note that we've had problems enabling languages with small corpora due to sentence splitting (dev/test/train buckets). is testing on staging not enough?

ftyers:
@mikehenrty the idea would be to show it to the people proofreading so they can make some recordings and see how it works. But if it's problematic from the point of data splits, it's no problem we can wait. :)

PR #1496, opened 2018-09-27
Add preliminary support for Asturian

softastur:
As the Asturian team has translated the Common Voice's webpage, the same team (from Softastur the association account) uploads the first 100 phrases to add preliminary support for Asturian.

This phrases are extracted from one of the Asturian member's mind. We are in conversations to get more phrases from Academia de la Llingua Asturiana (Academy of Asturian Language). Right now we are reviewing phrases from a Asturian sayings book whose author relicensed it's creation

Let us know if there is some requirement which does not meet or is missing.


mikehenrty:
Awesome work @softastur team!! We'll get these merged for now, and look forward to more sentences soon :)

PR #1511, opened 2018-10-11
add another 250 sentences

jf99:
Again, thanks @Ordoviz for proofreading!


jf99:
Now it's actually 500 sentences, because @Ordoviz quickly reviewed the next batch. Awesome!

PR #1518, opened 2018-10-13
[fr] add French sentences with (sometimes spelled) most common french firstnames and lastnames

nicolaspanel:
Add 5000 new french sentences generated from source: INSEE (public) using https://github.com/lissyx/commonvoice-fr/blob/4e6b78f6d58b0bef0e0014c6000bca4a553bb66e/names.py script.

### DIFF
--- a/server/data/fr/names.txt
+++ b/server/data/fr/names.txt
@@ -1,0 +1,5000 @@
+Marie Martin
+Monsieur Jean Bernard
+Monsieur Thomas, Pierre Thomas
+Michel M I C H E L
+Robert R O B E R T
+Jeanne Richard R I C H A R D
+mon nom c'est Philippe Dubois D U B O I S
+Durand, ça s'épelle D U R A N D
+Moreau, ça s'écrit M O R E A U
+nom Laurent, prénom Alain
+Jacques Simon
+Monsieur Bernard Michel
+Monsieur Lefebvre, Marcel Lefebvre
+Daniel D A N I E L
+David D A V I D
+Robert Roux R O U X
+mon nom c'est Paul Morel M O R E L
+Bertrand, ça s'épelle B E R T R A N D
+Fournier, ça s'écrit F O U R N I E R
+nom Girard, prénom Georges
+Nicolas Fontaine
+Madame Françoise Lambert
+Madame Dupont, Monique Dupont
+François F R A N Ç O I S
+Rousseau R O U deux S E A U
+Catherine Vincent V I N C E N T
+mon nom c'est Nathalie Muller M U deux L E R
+Lefevre, ça s'épelle L E F E V R E
+Faure, ça s'écrit F A U R E
+nom Andre, prénom Christophe
+Joseph Mercier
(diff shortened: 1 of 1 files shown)


mikehenrty:
thanks for this! have you seen this @lissyx?


> Michel M I C H E L

is part of the point to get people to spell out words (ie. pronounce individual letters)?

lissyx:
@mikehenrty I did saw this on my repo, https://github.com/lissyx/commonvoice-fr/pull/4 and spoke with Nicolas. Indeed, the goal is to get people to spell out words, but speaking with Kelly about that he fears this might need care on the language model side, and I was unsure if it would be clear enough to people reading sentences on the website.

mikehenrty:
from my perspective, it won't be clear to users that they need to spell out words. if we want to include the alphabet, maybe we should include certain sentences like "the letter a", rather than asking people to spell out words.

@lissyx, do you think it's useful to include these names and entities in this PR, but exclude the spelling part?

lissyx:
@mikehenrty I think there's value if we include less common names, we likely already have a lot of names with the french parliament dataset. I also think that we should not add too much of it :)

nicolaspanel:
Thank you @mikehenrty @lissyx  for your interest in this PR

> Kelly fears it might need care on the language model side

Since being able to handle spelled words, names, phone numbers or even emails is a common use case, I think it's worth investigating further on the LM part.

> it won't be clear to users that they need to spell out words

I agree it can be confusing. Maybe we could 
 - focus on explicit cases. ex: `martin M A R T I N`
 - include a notice on https://voice.mozilla.org/fr/speak
 - include a few hundreds samples first to see what people think / understand

lissyx:
> Since being able to handle spelled words, names, phone numbers or even emails is a common use case, I think it's worth investigating further on the LM part.

The only thing, is that until we can be sure it's okay with the LM, we'd need a way to be able to exclude those, in case it triggers unwanted behaviors. @mikehenrty Any idea?



>     * focus on explicit cases. ex: `martin M A R T I N`
> 
>     * include a notice on https://voice.mozilla.org/fr/speak

I guess it's worth a try, with proper notice? What do you think @mikehenrty ?

nicolaspanel:
@mikehenrty @kdavis-mozilla what do you think ?

nicolaspanel:
@lissyx
I think it is more a `mozilla` question

lissyx:
Ah right. We need to ask @Gregoor if we can have some hints on how things should be said in ambigous cases.

Gregoor:
I'm happy to include some of the sentences and see if users get it. If it works without UI changes, great. Otherwise I'd prefer we'd leave it out unless the ML folk here think it's too valuable to miss out on. It'd require a full-stack change I think, to have notes for specific sentences.

Gregoor:
Thanks for your efforts! We're about to launch a beta of our sentence collection tool, please submit your sentence through that once it's ready. It'll be announced in our Discourse very soon:
https://discourse.mozilla.org/t/readme-how-to-see-my-language-on-common-voice/31530/

PR #1519, opened 2018-10-14
New Esperanto sentences from "Proverbaro Esperanta"

RobinvanderVliet:
New Esperanto sentences from "Proverbaro Esperanta". This is a collection of Esperanto proverbs released in the year 1910. It is in the public domain, because the author has been dead for more than 100 years.

Only one sentence has been changed, the parenthesis have been removed from "Esti sub la ŝuo (de sia edzino)." because it would otherwise confuse a person trying to read it.


mikehenrty:
thanks @RobinvanderVliet ! these look great

PR #1524, opened 2018-10-17
Bring back normalization

### LINKED ISSUE #1498: Bring back audio normalization
In order to hotfix #1463, the PR to implement audio normalization #1285 has been reverted. I think this was meant as a temporary measure until a good fix for that issue is found.

In the PR thread, I've [suggested](https://github.com/mozilla/voice-web/pull/1285#issuecomment-422404670) a possible proper fix for #1463 that would work with normalization enabled.

Just filing an issue because the discussion is already 2 weeks old and I didn't want to forget about it.

est31:
Fixes #1498.

Should have safari support now by using the promise-decode-audio-data package the way its author described in [this gist](https://gist.github.com/mohayonao/89fd55526cef5a60364e860695a4bf75). But please test it because I don't have any safari lying around to test it. I've tested it on Firefox though and here normalization works again!

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "postcss-loader": "^2.0.7",
     "prettier": "^1.7.0",
     "prom-client": "^10.1.1",
+    "promise-decode-audio-data": "^0.4.1",
     "proxy-agent": "^2.0.0",
     "random-js": "^1.0.8",
     "react": "^16.3.0",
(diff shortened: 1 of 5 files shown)


mikehenrty:
@Gregoor, review when you get back?

Gregoor:
Thanks for bringing it back @est31! I'll have a Mac with me tomorrow, but it might be another week until we merge it to production (if it works as expected).

Gregoor:
Works! Thank you!

est31:
@Gregoor thank you!

PR #1544, opened 2018-11-04
add jf99-batch04.txt

jf99:
Another 250 sentences reviewed by @Ordoviz.


jf99:
@Gregoor Please also merge #1545 so that "erst mal" and "noch mal" are always written the same way.

Gregoor:
Done 👍

PR #1598, opened 2018-11-18
Use dualstack for S3

tambry:
Currently non-dualstack S3 endpoints are used, meaning that requests are made only over IPv4, even if the user's browser would prefer IPv6.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -15,7 +15,7 @@
     "@types/node": "^10.3.0",
     "@types/passport": "^0.4.3",
     "@types/request-promise-native": "^1.0.14",
-    "aws-sdk": "^2.67.0",
+    "aws-sdk": "^2.362.0",
     "body-parser": "^1.18.2",
     "common": "*",
     "commonmark": "^0.28.1",
--- a/server/src/config-helper.ts
+++ b/server/src/config-helper.ts
@@ -47,6 +47,7 @@ const DEFAULTS: CommonVoiceConfig = {
   ADMIN_EMAILS: '[]', // array of admin emails, as JSON
   S3_CONFIG: {
     signatureVersion: 'v4',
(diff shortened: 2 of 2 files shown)


Gregoor:
Hi, thanks for this! The CI shows a type error regarding the config key name, are the typings off there?

tambry:
It seems that the typings have an error. [The typings are for `useDualStack`](https://github.com/aws/aws-sdk-js/blob/cc29728c1c4178969ebabe3bbe6b6f3159436394/lib/config_use_dualstack.d.ts#L7), while [the option checked for is `useDualstack`](https://github.com/aws/aws-sdk-js/blob/cc29728c1c4178969ebabe3bbe6b6f3159436394/lib/region_config.js#L51).  
I think since variable names are case sensitive, specifying the uppercased one wouldn't actually work. I've filed https://github.com/aws/aws-sdk-js/issues/2364.

tambry:
@Gregoor Took a little while for an upstream fix and a new package release, but it should be good to go now.

Gregoor:
Awesome, thank you for that!

tambry:
@Gregoor Do let me know once it's deployed, so I can give it a quick test just in case. I was unable to test it myself with S3.

Gregoor:
@tambry it's already deployed on the [staging server](https://voice.allizom.org)

tambry:
@Gregoor Thank you, wasn't aware of the staging server. Seems to be working fine.

I'll see if I can get IPv6 for the web server itself too. [Made an upstream PR](https://github.com/nubisproject/nubis-terraform/pull/253) for an improvement necessary for that already.

Gregoor:
That's great! You're taking us to the future.. or present I guess? 🙈

PR #1600, opened 2018-11-19
Added nb-NO, Norwegian Bokmål

jurgemaister:
The translation files for Norwegian Bokmål is already in the repo, but not included in this file.


Gregoor:
Thanks for looking into this! That file is actually automatically updates once a locale has been translated >= 95%.  Until then you can look at it on [staging](https://voice.allizom.org/nb-NO).

PR #1601, opened 2018-11-19
fixed a typo in en/DNGros.txt

gyp:


DNGros:
Sorry about that... Good catch 👍

PR #1613, opened 2018-11-25
Add 140 more sentences.

DNGros:
Includes a mix of names and cities that might not have been covered previously. Also adds a couple of examples with medical drug names, which should pose a challenge for a system to understand and can be used to test an ability to generalize based off phonetic sound.


DNGros:
Closing as has been left unmerged for over a month and the project seems to be moving to a different system. See #341.

PR #1633, opened 2018-11-28
Enable IPv6 dualstack for the website

tambry:
Taking the website into the present. 🎉

### DIFF
--- a/nubis/terraform/main.tf
+++ b/nubis/terraform/main.tf
@@ -29,7 +29,7 @@ module "worker" {
 }
 
 module "load_balancer" {
-  source       = "github.com/nubisproject/nubis-terraform//load_balancer?ref=v2.3.1"
+  source       = "github.com/nubisproject/nubis-terraform//load_balancer?ref=6b91794839523ab5b3806824369efde2f61b3c17"
   region       = "${var.region}"
   environment  = "${var.environment}"
   account      = "${var.account}"
@@ -48,7 +48,7 @@ module "dns" {
   environment  = "${var.environment}"
   account      = "${var.account}"
   service_name = "${var.service_name}"
-  target       = "${module.load_balancer.address}"
+  target       = "${module.load_balancer.dualstack_address}"
 }
 
 resource "aws_db_parameter_group" "slow_query_enabled" {


Gregoor:
Awesome! Thanks for the lift in your Dolorean.

@gozer anything else we'd need for this?

tambry:
@Gregoor It should work fine, since it's a load balancer and internally it proxies the request over IPv4.

tambry:
@Gregoor @gozer Any news on this?

Gregoor:
Oh sorry, fell off my radar. Thanks for the reminder! I wanted to wait for feedback from @gozer first, but I think we can just give it a spin on staging for now and if it works it works.
Thanks for taking us to the future!

tambry:
@Gregoor The DNS change seems to have been applied, however connections are still refused over IPv6. I suspect this may be because it's necessary to whitelist IPv6 also in the worker security group (though that seems somewhat strange to me).  
Beforehand, could you provide me the `terraform.tfstate` to see if there's anything else wrong?

Gregoor:
Note that atm it's only on staging atm (https://voice.allizom.org), which atm has (unrelated) cert problems. Or did you try IPv6 on staging?

tambry:
I tried on staging.

Gregoor:
Great! Do you know where I'd find that tfstate? I don't know much about Terraform. Maybe @gozer knows?

tambry:
tfstate should be in the same folder as you're running the Terraform commands in.

Gregoor:
I think our Jenkins runs it (I don't 😁). In its logs nothing matches tfstate though.

tambry:
Seems like you're correct about Jenkins.  
The IPv6 dualstack domain name has been also deployed to production by now. Fortunately IPv6 actually not working doesn't really affect anything (besides a slight latency increase), due to Happy Eyeballs.

The `.tfstate` seems to be stored [at `http://state.nubis.${var.account}.${var.nubis_domain}/aws/${var.region}/${var.arena}.tfstate`](https://github.com/nubisproject/nubis-terraform/blob/develop/info/main.tf#L6). [`var.account` is unspecified in this repo](https://github.com/mozilla/voice-web/blob/master/nubis/terraform/variables.tf#L2), so it's probably specified in Jenkins.  
@Gregoor Mind trying to find it, placing it in `http://state.nubis.${var.account}.nubis.allizom.org/aws/us-west-2/core.tfstate` and seeing if you can fetch that?

PR #1662, opened 2018-12-04
Support for Universal Links in iOS app

dabinat:
This adds support for clicking a link in the user's web browser or mail app and having it launch the Common Voice app and navigate to the URL.

However, it needs the following server-side changes in order to be fully functional:

1. Edit the app ID at developer.apple.com and enable Associated Domains.

2. The authorization URL currently starts with https://auth.mozilla.auth0.com. Because Mozilla doesn't own auth0.com, there's no way to make it work with Universal Links. So the URL in the email needs to be changed to start with voice.mozilla.org and then redirect to the auth0 URL.

### DIFF
--- a/ios/voicebank-ios-wrapper.xcodeproj/project.pbxproj
+++ b/ios/voicebank-ios-wrapper.xcodeproj/project.pbxproj
@@ -53,6 +53,7 @@
 		3E42BC781ECA43B8002E3CDA /* Recorder.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Recorder.swift; sourceTree = "<group>"; };
 		3EC11D901EF45B6D00589896 /* BrowserViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = BrowserViewController.swift; sourceTree = "<group>"; };
 		3EE991BB1ED8B81800432FC5 /* Reachability.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Reachability.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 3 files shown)


Gregoor:
Thanks for jumping in so quickly! That's a great point RE the auth0 domain, I think we might not be able to do anything about it, but I'm investigating that now.

Gregoor:
It sounds like it's not at all straightforward, to change the email-link. I'm looking again into MediaRecorder polyfills, so that we don't need the iOS app anymore.

dabinat:
Any reason why the Github / Google / Firefox account authentications can't be brought back? They were there on day one of the user profile release and worked well for me until they were removed. Or passwords? They would be the simplest.

I know the app only exists because of the inability to record on iOS, but I think it's still useful because it's WAY more convenient than a browser. I think the following example will illustrate the difference.

Here's how I have to authenticate in my browser:
1. Type in voice.mozilla.org manually because bookmarks seem to be broken in the latest version of Firefox for iOS. (Seriously, bookmarking and copying links seems to copy the referrer rather than the URL.)
2. Click the square icon and select Login.
3. Type my email address.
4. Click Send Email.
5. Launch mail and open the email.
6. Tap and hold on the link and select Copy (VERY CAREFULLY because if you accidentally 3D Touch it, it invalidates the link).
7. Paste in Firefox.

I have to do this multiple times per day. Sigh.

Here's how I was authenticating in the app while Github authentication was allowed:
1. Launch app.
2. Click square icon and select Login.
3. Click Github and login to Github.
4. Approve the authentication request in Github.

And that's just the first time I use it. Every time after that I just launch the app and start using it straight away. So the difference between the iOS app and the browser is the difference between 7 steps and 1 step.

Gregoor:
We're at a work week currently, but @mbransn will outline our reasoning around this sometime next week.

PR #1674, opened 2018-12-05
Updating sentences in jf99.txt

nmstoker:
Fix some spelling mistakes and a few awkward phrases


server/data/en/jf99.txt
```diff
@@ -169,23 +169,23 @@ Taking part in road traffic is potentially dangerous after donating blood.
...
 This picture is pure art.
 A crowd of people stood around the fountain.
 When she entered the dance floor, everyone in the nightclub was looking at her.
 Wikipedia is a non-profit organization.
-The clock of the town center took damage from a lightening strike.
+The clock in the town center was damaged by a lightning strike.
 Water boarding is a terrible form of torture.
 The lawyer successfully defended his client.
-When the judge spoke the death sentence, the defendant showed no emotion.
+When the judge pronouced the death sentence, the defendant showed no emotion.
```
davidak: You probably mean:
```suggestion
When the judge pronounced the death sentence, the defendant showed no emotion.
```

nmstoker: Yes, you're quite right. Thanks for spotting that David!

davidak:
Accepting the suggested changes created a new commit.

@Gregoor to have just one commit, squash and merge.

Gregoor:
@davidak That's what I did. Squash and merge means that all commits will be rebased into a single commit (which is called squashing) and then merged. It will still a yield a commit, since that's the atomic unit of change in git.

PR #1675, opened 2018-12-06
Create option to sign up for emails through footer

### LINKED ISSUE #1408: Sign Up for Emails through the Footer
footer_01
footer_03
nav_footer 2 hover
nav_footer 2 done
footer_02

jkmdev:
## Description

Closes #1408 

This PR adds a component that gets inserted into voice-web's footer. The component will validate the input entered and give the user feedback on their submission through the submission button to the rightmost of the component. If their input is invalid or a server side error happened in the submission of their request the user will see an error. If their email is valid and their request was successful, they will see that their submission was successful.

This PR makes slight changes to how the responsiveness of the footer, and adds a new icon called `exclamation.svg` to the project.

I wanted to mention that the server-side code isn't submitting emails that I send to it, this is because I don't have `BASKET_API_KEY` in my `config-helper` file. I wasn't sure if this was something I'd need to fix on my end so I thought I'd make a PR as of now.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- I see some existing errors in the console but they don't seem to be related to my changes; they still exist in master.
- Some of the tests don't pass but they don't seem related to my changes, they still exist in master.


web/locales/en/messages.ftl
```diff
@@ -109,6 +109,7 @@ cookies = Cookies
 faq = FAQ
 content-license-text = Content available under a <licenseLink>Creative Commons license</licenseLink>
 share-title = Help us find others to donate their voice!
+email-subscription-title = SIGN UP FOR EMAIL UPDATES
```
Gregoor: can we have this in normal casing and then do CSS text-transform to uppercase it.

web/src/components/sign-up/sign-up.css
```diff
@@ -0,0 +1,55 @@
+@import url('../vars.css');
+
+.submission {
+    height: 45px;
+    display: flex;
+}
+
+input {
```
Gregoor: Unfortunately our css import system doesn't namespace files, so these selectors apply globally. Can we namespace it under a className like .newsletter-sign-up?

web/src/components/sign-up/sign-up.tsx
```diff
@@ -0,0 +1,88 @@
+import * as React from 'react';
```
Gregoor: (Comment on the file name, not on this line):
Could we rename it to something that includes newsletter? I feel like sign-up might be confused with the authorization flow.

web/src/components/sign-up/sign-up.tsx
```diff
@@ -0,0 +1,88 @@
...
+class SignUp extends React.Component<PropsFromState, State> {
+  constructor(props: PropsFromState, context: any) {
+    super(props, context);
+    const { api } = this.props;
+    this.state = {
+      email: '',
+      buttonStyle: 'submit-button',
+      buttonHtml: <OldPlayIcon className="icon" />,
+    };
+    this.handleChange = this.handleChange.bind(this);
```
Gregoor: nit: in other components we tend to do binding by assigning functions to properties, e.g. `handleChange = (event) => {}` (similar for state, then you won't need a constructor).

web/src/components/sign-up/sign-up.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  handleChange(event: any) {
+    this.setState({ email: event.target.value });
+    if (this.state.buttonStyle != 'submit-button') {
+      this.setState({ buttonStyle: 'submit-button' });
+      this.setState({ buttonHtml: <OldPlayIcon className="icon" /> });
```
Gregoor: It's a bit of an anti-pattern in React to have components in the state and I don't think it's necessary in this case. We could remodel the state, so that depending on the status of the form (e.g. initial/error/submitted), it renders different stuff. That'd also allow us to reduce the state properties down to 2.

web/src/components/sign-up/sign-up.tsx
```diff
@@ -0,0 +1,88 @@
...
+    } else {
+      this.setState({ buttonStyle: 'error-button' });
+      this.setState({ buttonHtml: <CautionIcon className="icon" /> });
+    }
+  };
+
+  private subscribeToNewsletter = () => {
+    this.props.api
+      .subscribeToNewsletter(this.state.email)
+      .then(result => {
```
Gregoor: In the rest of the codebase we use async await, to streamline async code.

web/src/components/sign-up/sign-up.tsx
```diff
@@ -0,0 +1,88 @@
...
+      })
+      .catch(err => {
+        this.setState({ buttonStyle: 'error-button' });
+        this.setState({ buttonHtml: <CautionIcon className="icon" /> });
+        throw new Error(err.message);
+      });
+  };
+
+  render() {
+    const { api } = this.props;
```
Gregoor: unused variable

web/src/components/sign-up/sign-up.tsx
jkmdev: Makes sense. Would you want to go with subscribe-newsletter instead?

Gregoor: That sounds good to me!

jkmdev:
Fun stuff. Looks like some changes that were made in master were overridden in this pr despite me updating my code. Fixing this now

web/src/components/subscribe-newsletter/subscribe-newsletter.tsx
```diff
@@ -0,0 +1,92 @@
...
+class SignUp extends React.Component<PropsFromState, State> {
+  state = {
+    email: '',
+    buttonStyle: 'submit-button',
+    isSubmitted: false,
+    isValidEmail: true,
+  };
+
+  handleChange = (event: any) => {
+    this.setState({ email: event.target.value });
```
Gregoor: You can set multiple state properties in one call, by putting them into the same object, e.g.: `setState({email: 'something', isSubmitted: false})`.

It's also a good pattern in React to not use derived state (unless it's costly to compute). In this case that would be isValidEmail (which can be derived from email) and buttonStyle (which can be derived from isValidEmail and isSubmitted).

Gregoor:
Finally got around to looking at this, sorry for the delay. Also I squashed all the commits together, apparently killing the reference to this PR in the process (though you're still the author of the commit). So this has been merged in: 0a457e9e01afc0e4abf69b7a9e18996584968532

Gregoor:
Thanks again for putting the work in!

I made a couple of changes to the [NewsletterSubscribe component](https://github.com/mozilla/voice-web/blob/0a457e9e01afc0e4abf69b7a9e18996584968532/web/src/components/layout/subscribe-newsletter.tsx). Most notably:
- use only 2 state variables, the others are deducible from those
- use form, label & button elements to make the site more accessible and keyboard friendly. Clickable divs aren't tabbable by default. And when you wrap an input and a submit button with a form you get enter-submit for free
- remove the email regex, modern browsers do email validation on inputs with type=email

rileyjshaw:
It looks like this has been merged in with https://github.com/mozilla/voice-web/commit/0a457e9e01afc0e4abf69b7a9e18996584968532, so this is safe to close. Thanks all!

PR #1703, opened 2018-12-18
update homepage legal image files

mbransn:
- images are now correctly mapped to their name / list number
- images are now all same pixel dim
- images are now 2x density

@Gregoor please feel free to merge if it looks like this is all done correctly


Gregoor:
Awesome, thanks for the PR 🙂 🤓

mbransn:
Thanks for the merge. 👍

Issue #1217, opened 2018-07-03
Pressing "No" on the last clip should stop the audio from playing


connorshea:
While validating clips, I ran into a clip which was just the word "F**ker" on loop with the bass boosted. When I pressed "No", I expected the audio to stop playing, but it continued until the clip ended. When the clip is malicious (admittedly a rare occurrence in my experience), this is pretty annoying.

It continued to play while on this screen:

<img width="1426" alt="screen shot 2018-07-03 at 12 08 23 pm" src="https://user-images.githubusercontent.com/2977353/42237072-f923fe80-7eb9-11e8-8ab4-e58c5fb71f23.png">

I'm not sure if this was because the clip happened to be the 5th clip in the set, or if this happens for all clips no matter what?

mikehenrty:
I'm sorry you had to listen to that! @connorshea

connorshea:
It didn't bug me much, I've seen worse :P Just wanted to make sure others don't have to deal with the same thing :)

poush:
Similarly with the case of "Yes" as well.

poush:
@Gregoor, Can I take this?

Gregoor:
Of course :)

PR #1709, opened 2018-12-20
Fixed: Clicking Yes/No control doesn't stop the playing audio

poush:
Fixed: #1217 

I think this quick fix should solve this bug.

### DIFF
--- a/web/src/components/pages/contribution/listen/listen.tsx
+++ b/web/src/components/pages/contribution/listen/listen.tsx
@@ -130,7 +130,7 @@ class ListenPage extends React.Component<Props, State> {
     const { clips } = this.state;
     const clipIndex = this.getClipIndex();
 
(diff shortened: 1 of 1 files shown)


web/src/components/pages/contribution/listen/listen.tsx
```diff
@@ -130,7 +130,7 @@ class ListenPage extends React.Component<Props, State> {
     const { clips } = this.state;
     const clipIndex = this.getClipIndex();
 
-    clearInterval(this.playedSomeInterval);
```
Gregoor: Why is it necessary to remove that line?

poush: This line is already present inside stop() method.

Gregoor: D'oh! Good point 🙈

PR #1740, opened 2019-01-07
Importing french CC-0 book

lissyx:
https://framabook.org/vieux-flic-et-vieux-voyou/


lissyx:
Thanks @goofy-mdn who did the work here :)

PR #1765, opened 2019-01-20
English spelling and grammar fixes

dabinat:
Big batch of fixes to improve spelling, grammar and address awkward phrasing.

### DIFF
--- a/server/data/en/DNGros.txt
+++ b/server/data/en/DNGros.txt
@@ -1,7 +1,7 @@
 How big can elephants be?
 Move these files into the pictures folder.
 Where did you find that apple?
-Teresa, Mildrid, Ralph, and Vonda all arrived yesterday evening.
+Teresa, Mildrid, Ralph and Vonda all arrived yesterday evening.
 Sometimes I wonder how there can be so many stars in the sky.
 Does the venue offer free WiFi?
 President Herbert Hoover had two children.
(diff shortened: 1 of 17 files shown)


Gregoor:
Thanks for the effort, I do have concerns about changing sentences in ways that aren't disambiguating (like changing casing). From what I've skimmed it looks like none of these would affect how sentences are pronounced. Yet in our backend, a new sentence would be created for each of these entries.
The problem with that is that for ML purposes we're creating 3 different buckets, which should have no overlap in sentences and users. If we have the "same" sentence multiple times (which we have no way of detecting atm), the separation of the buckets isn't as pure anymore, so we degrade the quality of the dataset.

At least that's my current understanding. @kdavis-mozilla, do you share that view?

dabinat:
What about lines that are grammatically incorrect and awkward to say? Half of these I end up rejecting because the speaker corrects the bad grammar and the other half of the time the person pauses or says it in an unnatural way.

e.g.
If you want a discount, you’ll have to provide a ID.
A chip of the old block.
You can’t sacrifice his life for you personal vendetta.
Leonardo DiCaprio is staring in the new movie.

While other lines are just plain nonsensical:

The voice, Chest voice the real voice, Head voice falsetto.
Men that are not included in the line, the reserve, Militia Never mind, Farewell.
The hull husk and the full in plenty Virtue is the highest good.

Is it better to just remove such lines from the project completely?

Gregoor:
Thanks, that's a good point. I think it makes sense to correct those, because they affect how people pronounce the sentences (or in the latter case, reduces the amount of nonsense 😁 ).

PR #1783, opened 2019-01-30
Sentence Collector - validated sentences export - 2019-01-30-18-45

MichaelKohler:
@Gregoor first export, would be great if you could review as this is the first one (of course not the sentences themselves, but rather the format and everything)


nukeador:
Really excited about this. Let's make sure people is able to see the progress, it's super motivating!

Cheers.

Gregoor:
Awesome, thanks again :) Looks good to me!

PR #1799, opened 2019-02-07
Export Sentence Collector - 2019-02-07

MichaelKohler:


dabinat:
It’s disappointing that sentences with errors are still making it through despite the review process.

e.g.

The only way too fly is first class
A mimim is a musical term corresponding to a half note
According to the poet, April is the cruellest month
Cathy took two hundred and ten pounds out of her bank account,
Doctor's write out many prescriptions for patients.
He looked at the handkerchief more, closely.
His central heating had stopped working,
Once the jews harp began emitting its barbaric rhythms, Michael was helpless.
The correct pronunciation is.
The drawer drew a lifelike image of the man.

nukeador:
I suspect that no matter what the process is there is always a % of errors. What we need to ensure is that this % is the lowest possible and enable mechanisms to fix them when encountered.

@dabinat I don't know how deeply you checked on errors for this export but can you make an estimation on the % of sentences with minor (misspelled) and major (invalid characters) errors?

jf99:
Imo, this pull requests shows that a couple of things are missing:

- We need a language specific place where we can discuss contribution rules within the community (including a language specific set of rules that can be complemented by the community). This place should ideally be integrated in the sentence-collector tool.
- We need a way to directly correct a wrong sentence in such a way that the original author notices it.
- Since it is too late for the sentences that are already in use, we need an official way to correct the wrong ones.
- We need a whitelist for allowed acronyms. Only for acronyms of which there really is just one correct pronunciation (such as BMW, BKA or DDR but *not* IT, DECT).

Referring to the German part of this PR, these are the main issues:
- Someone circumvented the prohibitions of acronyms by inserting dashes ("I-T-Sicherheit", "B-K-A").
- When I reviewed sentences, I tried to correct all sentences that I considered incorrect. That is, I added the corrected versions to the queue. As a result, some of the sentences in the corpus are present in both their correct and their incorrect form. Examples:
Der frühe Vogel kann mich mal.
Der frühe Vogel kann mich Mal.
Die Tanzlehrerin ist vierfache deutsche und zweifache Europameisterin im Standard- und Latein-Tanz.
Die Tanzlehrerin ist vierfache deutsche und zweifache Europameisterin im Standart Latein Tanz.
Greifswald ist eine sogenannte Universitätsstadt.
Greifswald ist eine so genannte Universitätsstadt.
- There are some sentences with regional slang. E.g.
Da kiekste wa?
Shouldn't an STT system recognize such a sentence as
Da guckst du, was?
This is a point which needs discussion at such a language specific place, which I requested above.

dabinat:
@nukeador That list was the entirety of the errors I found.

But the frustrating part is that many of these I saw in my review queue and voted down, yet they still ended up getting approved. I have seen other errors since and voted them down, but are they all going to end up being approved in the next batch?

nukeador:
Let's move to a better place for this conversation

https://discourse.mozilla.org/t/feedback-about-validated-sentences/35885

PR #1800, opened 2019-02-07
Fix back button arrow svg position.

m0ksem:
Set back button arrow to center of button for contribution. It was broken for me. 
image.
image


Gregoor:
Hi, thanks for getting in on that and fixing it! We already have a fix up on staging though: https://voice.allizom.org/en/speak

Issue #1768, opened 2019-01-21
Add convenient way to immediately re-record a line


lucaswerkmeister:
When I mess up a line, to re-record it I have to stop the recording, move the mouse over to the “repeat recording” button, click it, and start the recording again. I think for power users it would be nice to offer a faster way to do this: a keyboard shortcut to do all this at once, so that you can instantly repeat the line.

Something similar to this was proposed in #​2 of https://github.com/mozilla/voice-web/issues/313#issuecomment-317114368 (cc @jdittrich and @mikehenrty), though not quite as focused on the “power user” usecase, and the issue has since been closed. (I do agree with @Gregoor that the page now lets me do the things I need, re-recording wise, I just wish it was faster.)

Alternatively, extreme power users apparently [assign this and other actions to a dedicated controller](https://twitter.com/ubahnverleih/status/1087111164461543425) (cc @ubahnverleih), so I guess that’s always an option :)

mritunjaygoutam12:
should it be closed ?

Gregoor:
Yes, thanks!

PR #1803, opened 2019-02-13
Added keyboard shortcut (shift + clip number) to immediately re-record a line

mritunjaygoutam12:
Added keyboard shortcut (shift + clip number) to immediately re-record a line.
clip number indexed like 1,2,3,4,5.
solves #1768

### DIFF
--- a/web/src/components/pages/contribution/speak/speak.tsx
+++ b/web/src/components/pages/contribution/speak/speak.tsx
@@ -164,7 +164,7 @@ class SpeakPage extends React.Component<Props, State> {
     this.audio.setVolumeCallback(this.updateVolume.bind(this));
 
     document.addEventListener('visibilitychange', this.releaseMicrophone);
-    document.addEventListener('keyup', this.keyUprerecording);
+    document.addEventListener('keyup', this.handleKeyUprerecording);
 
     if (
       !this.audio.isMicrophoneSupported() ||
@@ -176,7 +176,7 @@ class SpeakPage extends React.Component<Props, State> {
   }
 
   async componentWillUnmount() {
-    document.addEventListener('keyup', this.keyUprerecording);
+    document.addEventListener('keyup', this.handleKeyUprerecording);
 
     document.removeEventListener('visibilitychange', this.releaseMicrophone);
     if (!this.isRecording) return;
@@ -187,7 +187,7 @@ class SpeakPage extends React.Component<Props, State> {
     return this.state.recordingStatus === 'recording';
   }
 
-  private keyUprerecording = async (event: any) => {
+  private handleKeyUprerecording = async (event: any) => {
     let index = null;
     //for both sets of number keys on a keyboard with shift key
     if (event.which === 35 || event.which === 49) {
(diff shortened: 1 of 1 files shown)


web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -184,6 +187,30 @@ class SpeakPage extends React.Component<Props, State> {
     return this.state.recordingStatus === 'recording';
   }
 
+  private keyUprerecording = async (event: any) => {
```
Gregoor: I'd recommend a name like `handleKeyUpReRecord`

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -184,6 +187,30 @@ class SpeakPage extends React.Component<Props, State> {
...
+    let index = null;
+    //for both sets of number keys on a keyboard with shift key
+    if (event.which === 35 || event.which === 49) {
```
Gregoor: Can we use `event.key` instead? It's more descriptive than numbers.

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -184,6 +187,30 @@ class SpeakPage extends React.Component<Props, State> {
...
+      index = 1;
+    } else if (event.which === 34 || event.which === 51) {
+      index = 2;
+    } else if (event.which === 37 || event.which === 52) {
+      index = 3;
+    } else if (event.which === 12 || event.which === 53) {
+      index = 4;
+    }
+
+    if (index !== null) {
```
Gregoor: Does this work when the given index doesn't have a recording yet? I'd imagine it doesn't.

web/src/components/pages/contribution/speak/speak.tsx
mritunjaygoutam12: done

web/src/components/pages/contribution/speak/speak.tsx
mritunjaygoutam12: yes, but then there will be little change in value returned from event
In case of event.key ->      shift +1 = "end" or "!"
                                          shift +2 = "arrowdown " or "@"
                                         or more

web/src/components/pages/contribution/speak/speak.tsx
mritunjaygoutam12: yes, this works even when it doesn't have a recording yet

mritunjaygoutam12:
@Gregoor please review it 🙂

Gregoor:
Thanks for the ping!

web/src/components/pages/contribution/speak/speak.tsx
Gregoor: Good point! `event.code` should do the trick! That would yield `Digit1` (even when shift is pressed).

web/src/components/pages/contribution/speak/speak.tsx
Gregoor: Huh, I'm positively surprised that it does. The page is more resilient than I thought. Great!

web/src/components/pages/contribution/speak/speak.tsx
mritunjaygoutam12: is there something left to be done in this PR.

Gregoor: If you could switch to using `event.code` that'd be great. That way it'd be clear which keys do what, just by reading the code.

mritunjaygoutam12: awesome. Done

mritunjaygoutam12:
I think changes complete. Ready for merge.

PR #1805, opened 2019-02-14
Spanish accents (from EN wikipedia)

nukeador:
Spanish accents from https://en.wikipedia.org/wiki/Spanish_dialects_and_varieties#Sets_of_variants

### DIFF
--- a/web/src/stores/demographics.ts
+++ b/web/src/stores/demographics.ts
@@ -61,6 +61,19 @@ export const ACCENTS: any = {
   eo: {
     internacia: 'Internacia',
   },
+  es: {
+    nortepeninsular: 'España: Norte peninsular (Asturias, Castilla y León, Cantabria, País Vasco, Navarra, Aragón, La Rioja, Guadalajara, Cuenca)',
+    centrosurpeninsular: 'España: Centro-Sur peninsular (Madrid, Toledo, Castilla-La Mancha)',
+    surpeninsular: 'España: Sur peninsular (Andalucia, Extremadura, Murcia)',
+    canario: 'España: Islas Canarias',
+    mexicano: 'México',
+    americacentral: 'América central',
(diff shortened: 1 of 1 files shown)


capiscuas:
The East Spain accents are missing, also galician.
https://en.wikipedia.org/wiki/File:Spanish_dialects_in_Spain-en.png

I would add as:
    estepeninsular: 'España: Este peninsular (Baleares, Cataluña, Comunidad Valenciana)',
    galicia: 'España: Galicia'

nukeador:
The original article didn't include them as accents but as separate languages (Galician and Catalan)

> In purple, the major variations and dialects of Castilian/Spanish in Spain.
In other colors, the extent of the other languages of Spain in the bilingual areas.

capiscuas:
Yeah but people who don't speak those languages or those languages(Catalan and Galician) are not their mother tongue, they should still be able to choose Castillian/Spanish and their region(Galicia) for example.

I am not a linguistic but I think that it makes sense for bi-lingual regions or bi-lingual countries.

capiscuas:
It is interesting issue.
image

In my case Catalan and Castilian/Spanish are my 2 mother tongues: in Catalan I can select Valencia(my dialect), but in the case of Castilian I can't choose my area(East Spain) cause it only shows dialects of regions non-bilingual. 
Should I then just select Catalan->Valencia?

What if somebody lives in valencia but only speaks Castilian?

nukeador:
Maybe we should include all regions in one of the existing buckets?

Galicia and Catalonia in the Northern and Valencia in the Central?

capiscuas:
Galicia is rather Northern
and Catalonia, Valencia and Baleares(islands) are rather Eastern(new bucket, bcs our accents are influenced by the fact that most of us are native-bilingual)

PR #1806, opened 2019-02-14
Another italian review to improve the sentences

Mte90:
We worked on cleaning, fix typos and remove duplicates.


Gregoor:
Thanks, I'm gonna merge this one, but could we in the future do this for sentences that are wrong:

create a PR to remove this and then use the sentence collection tool to add it?

Mte90:
ok but right now in this pr we didn't added new sentences only fix on the previous.
about the sentence collector I don't know if fit the italia ncase because we have our linter for the sentences

Gregoor:
We can create that linter into the process. @nukeador might have some pointers for you for that.

RE it just being changes: Common Voice doesn't know of corrections, so they just appear as new sentences.

nukeador:
Daniele, we can talk about this but definitely we need hundreds of thousands sentences per language if we want to reach someday the 2000hrs, there is going be to always a need for new sentences ;-)

PR #1840, opened 2019-02-21
double click on stop makes recordingStatus to waiting which leads to …

### LINKED ISSUE #1706: Clicking twice on stop recording button breaks button
When the stop recording button is pressed twice, it breaks the button. It disappears and can't be used again without refreshing the page. I think commit 79b1511 introduced this bug, but I am not sure.

This is how the button looks like afterwards:
image

mritunjaygoutam12:
double click on stop makes recordingStatus to waiting which leads to disable icon.
Fixed #1706

### DIFF
--- a/web/src/components/pages/contribution/speak/speak.tsx
+++ b/web/src/components/pages/contribution/speak/speak.tsx
@@ -282,8 +282,6 @@ class SpeakPage extends React.Component<Props, State> {
     if (this.state.recordingStatus === 'waiting') return;
     const isRecording = this.isRecording;
 
-    this.setState({ recordingStatus: 'waiting' });
-
     if (isRecording) {
       this.saveRecording();
       return;
--- a/web/src/components/pages/contribution/speak/speak.tsx
+++ b/web/src/components/pages/contribution/speak/speak.tsx
@@ -282,37 +282,41 @@ class SpeakPage extends React.Component<Props, State> {
     if (this.state.recordingStatus === 'waiting') return;
     const isRecording = this.isRecording;
 
-    await this.setState({ recordingStatus: 'waiting' });
+    this.setState({ recordingStatus: 'waiting' });
 
     if (isRecording) {
       this.saveRecording();
       return;
     }
 
-    setTimeout(async () => {
-      try {
-        await this.audio.init();
-        await this.startRecording();
-      } catch (err) {
-        if (err in AudioError) {
-          this.setState({ error: err });
-        } else {
-          throw err;
-        }
+    try {
+      await this.audio.init();
+      await this.startRecording();
+    } catch (err) {
+      if (err in AudioError) {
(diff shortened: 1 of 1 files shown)


Gregoor:
I don't think we should completely remove the waiting state, as it indicates that recording hasn't started yet. Maybe what we should do is make the second click not do anything (i.e. if it's already in a waiting state)?

mritunjaygoutam12:
I agree it indicates that recording hasn't started yet. But I not able to find any anywhere where we are using this state ( "waiting" recording status) except disabling button.

mritunjaygoutam12:
@Gregoor . I got it the "waiting" here. 
Actual problem is on double click is, this.audio.init() called twice which throw err(Failed to execute 'start' on 'MediaRecorder': The MediaRecorder's state is 'recording'.) in catch statement which stop execution of rest of code in try statement and it leads to no status("waiting") change of disabled button.

solution (I tried. It worked for me) : setTimeout after setStateRecording. Time equals to RECORD_STOP_DELAY in saveRecording();

Benefit of this solved issue is it give speaker fast and smooth way to stop and start mic.

Gregoor:
I don't think recordings shorter than 500ms are realistic tbh. Also setTimeout seems sub-optimal, as it doesn't tackle the core problem. Let me try to recap what the problem is:
When quickly clicking the button twice, the method is run twice which leads to it erroring, as the audio initialization can only happen once. The `recordingStatus == 'waiting'` check & early-return should guard against it, but it doesn't because setState is async. Unfortunately await doesn't do anything against that, setState doesn't return a promise and even if it would, the method could still be called twice.
I guess one way to fix that would be to also set a class property (i.e. not in state) which we'd check instead of the state property.

mritunjaygoutam12:
@Gregoor review it. It is working fine.

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -302,19 +302,25 @@ class SpeakPage extends React.Component<Props, State> {
...
+    } catch (err) {
+      this.setState({
+        recordingStatus: null,
+      });
+    }
   };
 
   private saveRecording = () => {
-    const RECORD_STOP_DELAY = 500;
+    const RECORD_STOP_DELAY = 300;
```
Gregoor: could we revert this back to 500 or is it necessary? I think we had some reason why we wanted to go with that instead of a lower number.

mritunjaygoutam12: No, it not necessary. I forget to revert it.

mritunjaygoutam12:
Reverted the record_stop_delay to 500.

Gregoor:
Thanks for the fix and the patience! :)

mritunjaygoutam12:
Thank you for your repeated reviewing and advice on my PR.

PR #1860, opened 2019-02-28
Fix conflicting shortcuts while listening to Esperanto

b-bdf:


Gregoor:
Heya, thanks for the contribution The localizations are managed through Pontoon though: https://pontoon.mozilla.org/projects/common-voice/

PR #1873, opened 2019-03-03
webpack setup for image loading

### LINKED ISSUE #1705: Use webpack for bundling assets
The /img directory thing is suboptimal, for two reasons:

- can't co-locate assets with their views
- no build processing timeline (e.g. for optimizing PNGs)

mritunjaygoutam12:
Fixes #1705 
It helps to import image as module and build with webpack.


Gregoor:
Great, thanks for laying some of the groundwork! I guess it'd also be great to have this for CSS image imports (e.g. `background: url("/img/something.png");`). And we can also use this PR to move all the images over (which I can help with later this week)!

mritunjaygoutam12:
@Gregoor after some try I found  It is already working. only we have to change the path to relative. Example: 
`background: url("/img/something.png");`   -------> wrong
                 `background: url("../../../img/something.png");`     -------> right (relative path of Image)

Gregoor:
Oh that's cool! The relative paths get a bit ugly, but I think we already have that same problem in our TS files, so at some (later) point we should add a way to absolutely resolve paths (which I think also requires some typescript configuration).

Could you revert the messages.ftl commit, as it's not related to this PR (and already fixed in master)?

Thanks!

mritunjaygoutam12:
fixed. Now, we can use `background: url("~image/something.png"); `

mritunjaygoutam12:
@Gregoor completed .

mritunjaygoutam12:
WIP

mritunjaygoutam12:
Now, I checked it on my side. please review it @Gregoor .

web/src/components/layout/logo.tsx
```diff
@@ -1,10 +1,10 @@
 import * as React from 'react';
 import { LocaleLink } from '../locale-helpers';
+import * as logo from '../../../img/cv-logo-one-color-white.svg';
```
Gregoor: Oh, so the `~img/cv-logo-one-color-white.svg` wouldn't work here?

mritunjaygoutam12: Oh, may be now it is done

mritunjaygoutam12:
every where in tsx file I have used `Robot`
but webpack it not catching the image from here. we can change it to `Robot` . it is working fine. @Gregoor what should we do here ?

Gregoor:
> 
> 
> every where in tsx file I have used `Robot`
> but webpack it not catching the image from here. we can change it to `Robot` . it is working fine. @Gregoor what should we do here ?

Yeah I think I like the latter better, despite it being more code. Also because we'd then get warning when the file can't be found, right?

mritunjaygoutam12:
yeah awm, I was not thing of this benefit. Ok then I will do this now.

mritunjaygoutam12:
may be now it is fixed .

Gregoor:
I have to stop merging on my machine, that keeps for some reading discarding author information and doesn't close the PR. 

This has been merged in https://github.com/mozilla/voice-web/commit/40a34ba1fc071ca2d0d8da4782202f9dc8e044ae

Some changes I made was de-duplicating images that were duplicated and creating images folder for dirs that have many images.

Thankslots I think this everything much saner!

mritunjaygoutam12:
It is always great to see your code in master :heart:

PR #1874, opened 2019-03-04
add zh-tw sentences from Taipei Sprint and g0v hackathon

irvin:


Gregoor:
Great! Have they been reviewed / could we put them in the sentence collector to make sure they fit the rules?

irvin:
@Gregoor Yes, We had check them together in the end of event, and I had also review each single of them. 
Please if we could avoid put them into sentences collector, because that would also me again to review (and we have few thousand more sentences in the coming weeks)

irvin:
(totally 2980 new sentences for now)

irvin:
after discuss with the local community members, we decides to dump all the sentences into collection tools.

PR #1875, opened 2019-03-04
Update native-names.json

jurikolo:
Added Latvian name


Gregoor:
Hello, thanks for the PR! This file is updated automatically, using the native name that is given in the language's `message.ftl`. To add its native name, you'd need to use our localization too, Pontoon: https://pontoon.mozilla.org/projects/common-voice/

PR #1932, opened 2019-04-02
saveclip at backend error handling

### LINKED ISSUE #1459: Let clip saving errors bubble up
Didn't happen on production in the last time, but we noticed that those errors are caught too early.

mritunjaygoutam12:
Fixes #1459

### DIFF
--- a/server/src/lib/clip.ts
+++ b/server/src/lib/clip.ts
@@ -123,66 +123,74 @@ export default class Clip {
     const sentenceFileName = folder + filePrefix + '.txt';
 
     // if the folder does not exist, we create it
-    await this.s3
-      .putObject({ Bucket: getConfig().BUCKET_NAME, Key: folder })
-      .promise();
+    try {
+      await this.s3
+        .putObject({ Bucket: getConfig().BUCKET_NAME, Key: folder })
+        .promise();
 
-    // If upload was base64, make sure we decode it first.
-    let transcoder;
-    if ((headers['content-type'] as string).includes('base64')) {
-      // If we were given base64, we'll need to concat it all first
-      // So we can decode it in the next step.
-      const chunks: Buffer[] = [];
-      await new Promise(resolve => {
-        request.on('data', (chunk: Buffer) => {
-          chunks.push(chunk);
+      // If upload was base64, make sure we decode it first.
+      let transcoder;
+      if ((headers['content-type'] as string).includes('base64')) {
+        // If we were given base64, we'll need to concat it all first
+        // So we can decode it in the next step.
+        const chunks: Buffer[] = [];
+        await new Promise(resolve => {
+          request.on('data', (chunk: Buffer) => {
+            chunks.push(chunk);
+          });
+          request.on('end', resolve);
         });
-        request.on('end', resolve);
+
+        const passThrough = new PassThrough();
+        passThrough.end(
+          Buffer.from(Buffer.concat(chunks).toString(), 'base64')
+        );
+        transcoder = new Transcoder(passThrough);
+      } else {
+        // For non-base64 uploads, we can just stream data.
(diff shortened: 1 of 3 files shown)


mritunjaygoutam12:
@Gregoor review it, please

Gregoor:
I will review it tomorrow 🤞

web/src/services/api.ts
```diff
@@ -67,7 +67,7 @@ export default class API {
       return;
     }
     if (response.status >= 400) {
-      throw new Error(await response.text());
+      throw response;
```
Gregoor: why not wrap it in an error object?

server/src/lib/clip.ts
```diff
@@ -120,64 +120,71 @@ export default class Clip {
...
+        path: clipFileName,
+        sentence,
+        sentenceId: headers.sentence_id,
+      });
+      await Awards.checkProgress(client_id);
 
-    response.json(filePrefix);
+      response.json(filePrefix);
+    } catch (error) {
+      console.log(error);
```
Gregoor: let's use `console.error` instead

server/src/lib/clip.ts
```diff
@@ -120,64 +120,71 @@ export default class Clip {
...
+      response.status(error.statusCode || 500).json(error);
```
Gregoor: let's just return a generic error message here, error will be meaningless to the client anyway.

mritunjaygoutam12: Ok, then just send some error message then in below speak.tsx only check for that error message to differentiate between the errors ??

server/src/lib/clip.ts
mritunjaygoutam12: :+1:

web/src/services/api.ts
mritunjaygoutam12: :+1:

server/src/lib/clip.ts
mritunjaygoutam12: let me do this

Gregoor: I think it'd be best if we just console.error'd it here and output something generic, as the client doesn't need to know about the internals here.

mritunjaygoutam12: :+1:

mritunjaygoutam12:
conflict resolved :relaxed:

web/src/services/api.ts
```diff
@@ -66,8 +66,8 @@ export default class API {
       location.reload();
       return;
     }
-    if (response.status >= 400) {
-      throw new Error(await response.text());
+    if (response.statusText === 'Error in saveClip server') {
```
Gregoor: I don't think we should replace the generic error handling here. Also the generic case would also cover this case wouldn't it?

server/src/lib/clip.ts
```diff
@@ -123,66 +123,75 @@ export default class Clip {
...
+      });
+      await Awards.checkProgress(client_id);
 
     response.json(filePrefix);
 
     Basket.sync(client_id).catch(e => console.error(e));
+      response.json(filePrefix);
+    } catch (error) {
+      console.error(error);
+      response.statusMessage = 'Error in saveClip server';
```
Gregoor: maybe name it something like `save_clip_error` instead, and then check directly for it in `speak.tsx`?

mritunjaygoutam12:
little new approach. Steps:
>>>  Catching error in backend clip.ts and responded it with a status code and message. status code because it can be handled by previous generic error handling in services/api.ts

>>>Throwing separate new error for save_clip_error  in api.ts because of previous error throw hole error as text so I am unable to operate on the error in speak.tsx

Gregoor:
Great, looks good to me. Thanks for your hard work and patience with me :)

PR #1943, opened 2019-04-12
removed and fixed incorrect Mongolian sentences

sharavsambuu:
There were Mongolian sentences difficult to say and incorrect. Removed and fixed some of them.

### DIFF
--- a/web/locales/ha/cross-locale.ftl
+++ b/web/locales/ha/cross-locale.ftl
@@ -1,0 +1,4 @@
+## Languages
+
+contribute = taimaka
+get-involved-button = Shiga a dama da kai


MichaelKohler:
For the sentence-collector.txt the main data source is the database where https://common-voice.github.io/sentence-collector/# pushed sentences to. Deleting them here won't help, as with the next export they would just be added again. At some point, can't say when though, we will create a script to delete those sentences (https://github.com/Common-Voice/sentence-collector/issues/198). Until then, can you please write those down in a text file line by line? Then we could take that full text file as input and delete them.

Thanks
Michael

sharavsambuu:
Thank you for letting me know, where should create a text file?

MichaelKohler:
That's a good question. I guess an issue at https://github.com/Common-Voice/sentence-collector to keep track would be best. Then others could have a look at the list as well.

Thanks

PR #1963, opened 2019-04-25
Remove bad sentences from wiki import

dabinat:
110K bad sentences were removed and 20K sentences had abbreviations or symbols converted to words. See https://github.com/dabinat/cvtools/blob/master/sentence_validator.py for the exact criteria used for removal.

This should greatly reduce the number of partial or unreadable sentences, but it didn't get all of them and manual validation will still be required.


nukeador:
Thanks for this, adding @Gregoor for review.

According to my math this means around 8% of the 1,5M sentences had issues. Which is not too bad but it's great to have it fixed.

@dabinat do you think that with this PR we have reduced the error rate to 1% or less?

Gregoor:
Awesome, thanks for your hard work @dabinat!

Sampled a bit through the diff and looks like a great improvement.

dabinat:
> @dabinat do you think that with this PR we have reduced the error rate to 1% or less?

My script is better at detecting the second half of a partial sentence than the first half, so there are still partial sentences remaining. The script also does nothing about misspellings or nonsensical sentences.

There are also sentences with difficult foreign names that don't have accented characters. I have heard a lot of speakers struggling with them when I validate. I feel like those are more subjective and will need to be manually flagged up where appropriate.

So I think <=1% is unlikely, but my script has removed most of the egregious errors.

dabinat:
This PR removed all sentences with accented characters but I'm still seeing sentences with those characters on the Speak page, e.g.

> Ruins of Misión El Descanso remain within the community of El Descanso.

I checked and this sentence no longer exists within the wiki.en.txt file.

Did something go wrong with the merge or are the old sentences still cached somewhere?

Gregoor:
Sorry for the confusion, after the last deploy took down the DB I disabled the sentence import. It'll redo it on the next deploy.

PR #1964, opened 2019-04-25
Fix incorrect translate

alihoseiny:
Incorrect and meaningless translation of the `clips-uploaded` fixed.


nukeador:
Hello,

Localizations are handled via pontoon, any fixes should come from there

https://pontoon.mozilla.org/fa/common-voice/

Thanks for your contributions!

Cheers.

PR #1973, opened 2019-04-29
Clarify format of Auth0 credentials

nmstoker:
The keys for use with Auth0 had a slight mismatch between the keyname given (SECRET) and that in the code (CLIENT_SECRET) so simply clarifying that here in this minor update


Gregoor:
Thank you Neil!

PR #1989, opened 2019-05-01
Remove 31K bad sentences and 30K duplicates

dabinat:
This PR strips out sentences from the wiki dump with missing words and attempts to reduce the number of foreign person and place names in the text. This amounts to 31K removals.

It also removes 30K duplicate sentences from the file.

Some remaining sentences were cleaned up too, to remove errant quote and full stop characters.


Gregoor:
Thanks for all the hard work cleaning this up! Were the 30k exact duplicates? I.e. did you find them by checking each sentence if it already exists?

I'm asking so that we can feed it back into the wiki scraper once someone finds the time for it.

dabinat:
> Were the 30k exact duplicates? I.e. did you find them by checking each sentence if it already exists?

Yep, that’s right.

JAGulin:
Were they removed from further recording requests or entirely?
@dabinat Out of curiosity, did anyone check how many of these that had been:
a) recorded b) accepted c) rejected d) remain in any work queue?
That could serve both as an evaluation of the filter criteria and the acceptance rules.

dabinat:
The goal was to try to remove sentences like this:

> Recently, a very different pigment has been found in some marine γ-proteobacteria: proteorhodopsin.

> Bacterial chlorophyll b has been isolated from Rhodopseudomonas spp.

> The name was contrived by the railway for publicity and to outdo Llanfairpwllgwyngyllgogerychwyrndrobwllllantysiliogogogoch.

> Pavel Melnikov-Pechersky, Velimir Khlebnikov, Pyotr Boborykin, Eugenia Ginzburg, Leo Tolstoy, Daniil Mordovtsev, Alexander Tarasov-Rodionov.

Sentences like this were indeed recorded and people either missed out the words, said “something” instead of the long word, or attempted a really bad pronunciation. I think one guy got a couple of them right.

For some reason, people don’t skip sentences and just try to read them anyway. So IMO it’s disrespectful to volunteer time to have a lot of sentences that speakers can’t pronounce and validators still have to listen to, especially if those words aren’t in common usage anyway.

This PR also removed sentences that were objectively wrong like:

> Only the plural form "" is used even when addressing one person.

> They also released a best of album, "".

> is applied to species he described.

These sentences were removed by a script, so some “good” ones were caught up in the process but I made an effort to ensure this was minimal.

Someone could hunt through the 30K to find the 200(?) ones caught up but that seems like a wasted effort when there are 1.4 million sentences remaining in the file.

JAGulin:
I fully agree with your decision @dabinat and I meant it just as a quick question. Sorry, I was unclear. I believe this will not be a one-time cleanup, especially considering all languages, so learning from the experience should be rewarding.

My main thought was that these "faulty" sentences could also be removed from the verification process if they were already recorded. They will eventually go away, but it would be interesting to know how many that had time to slip through. I'm not sure if there was a tool to remove recordings made.

Secondary thought was to challenge the current "2 votes" consensus. Given the assumptions, the faulty sentences should mainly be rejected in verification step. It should be possible to check that and evaluate.

That said, I was just curious if any measurements had been done. I fully understand that it would be a decision of where to spend time.

Issue #1994, opened 2019-05-02
Search button is not correctly displayed


rleitan:
Environment:
Windows 10, Mac OS 10.14 - Firefox, Chrome
[STR]:
1.Open https://voice.allizom.org/en/faq-new

[Expected result]:
"Search for answers" placeholder should be black 
Search button should be visible

https://app.zeplin.io/project/5a9eec5c53a9a2685450d4be/screen/5c92b7a7e6cbce24f53dfe6e

[Actual result]:

![2019-05-02_12h13_57](https://user-images.githubusercontent.com/41474886/57066457-e60d2b00-6cd4-11e9-8224-3dc3913dc2ab.png)

rleitan:
@munteanu210

arnavvats:
Hi, would like to take up the issue.

arnavvats:
We can disable the auto search or keep this the current way. But how do we improve the UX?
@Gregoor can you give any suggestions?

munteanu210:
@arnavvats the search functionality was thoroughly discussed and there is a set way we want it to work. We don't want the auto search disabled, because it was done on purpose after discussing it with the team. Also, a decent amount of work went into the auto search to make it work fast and pull results in the way it currently does. I will fix the search icon so that it is displayed correctly on all browsers. As for allowing users to click on the search icon - that is doable. (Here is what I mean by Search button: [https://www.screencast.com/t/8P9db3r7d4sS](https://www.screencast.com/t/8P9db3r7d4sS)) However, we can also keep the Search button there for display and aesthetics purposes. As long as the auto search works well (and it does), users will use the auto search and will not need to press on the Search button.

arnavvats:
okay, nice.

Issue #1981, opened 2019-04-30
Two sentences repeat on about us page


zerdo90:
On https://voice.allizom.org/en/about page under the "≥ 2 No votes" and "Common Voice Dataset" headlines there is the same text: "When a user rejects a voice clip it returns to the Queue. If rejected a second time, the voice clip is moved to the Clip Graveyard." 

I suppose under the "Common Voice Dataset" should be something else?

![about](https://user-images.githubusercontent.com/26620773/56953661-57b27100-6b3d-11e9-9b8a-8b1309a40ac9.jpg)

Gregoor:
Thanks for the report!

cc @munteanu210

munteanu210:
@zerdo90 thanks for identifying this! I am working on changing it already.

munteanu210:
@mbransn @georai These two texts were identical on the design. See [screenshot](https://www.screencast.com/t/CfJ1ggl6pb). Please provide the correct / unique texts per each of these two sections, and I will update them.

PR #2019, opened 2019-05-07
Fixes for #1994 and #1981

munteanu210:


TheoChevalier:
@Gregoor  Heads up that the string change in the PR is not picked up by localizers in Pontoon unless you update the string ID (or like for French, that someone notices that the translation doesn’t match anymore — thanks @lissyx )

Gregoor:
thanks for the ping, Theo! I've renamed the string.

PR #2032, opened 2019-05-11
2018-05-11 Sentence Collector - validated sentences

MichaelKohler:


MichaelKohler:
I'll follow-up with a newer PR in a few minutes.

PR #2039, opened 2019-05-18
Fix "checked" property of checkboxes

FujiHaruka:
I clicked checkboxes below, but check marks don't appear.

I found `"checked"` property was not passed to the checkbox components, so I fixed it.

スクリーンショット 2019-05-18 22 11 52

(In addition, I fixed react "key" warning of list)


Gregoor:
Awesome, thank you!

PR #2040, opened 2019-05-21
Wiki cleanup

dabinat:
Removes 9000 partial sentences and sentences with unpronounceable foreign or scientific terms. Expands abbreviations for 1400 sentences.


nukeador:
@Gregoor let's have this merged and let's make sure we incorporate this rules to the English extraction.

PR #2041, opened 2019-05-22
Avatar clip tag initialisation

mritunjayz:


mritunjaygoutam12:
WIP

mritunjayz:
@Gregoor now you can review it !

server/src/lib/api.ts
```diff
@@ -86,7 +95,8 @@ export default class API {
       bodyParser.raw({ type: 'image/*' }),
       this.saveAvatar
     );
-
+    router.post('/user_client/avatarclip', this.saveAvatarClip);
```
Gregoor: I'd recommend going with `avatar_clip` for consistency's sake

mritunjayz: done

mritunjayz:
done !

web/src/components/pages/profile/avatar-setup/avatar-setup.tsx
```diff
@@ -145,6 +259,25 @@ class AvatarSetup extends React.Component<Props> {
           </label>
         </div>
 
+        <div style={{ display: isProduction() ? 'none' : 'block' }}>
```
Gregoor: Another way to do this is to write:
```jsx
{isProduction() && <div>someContent</div>}
```
Which decreases the amount of things that need to be rendered and computed in the false-branch of that conditional. But since this is non-critical here, I'll just merge it in.

mritunjayz: Got it ! double &&

PR #2047, opened 2019-05-24
hide input feild

mritunjayz:


mritunjayz:
This can be merged !

mritunjayz:
@Gregoor

Gregoor:
Won't the input stop being clickable if we hide it? I.e. can one still upload an avatar?

mritunjayz:
Yes, It works only hide the default HTML input box. Otherwise, everything will be the same.

Gregoor:
Awesome, thanks!

PR #2059, opened 2019-06-04
voice avatars of other profile owners

mritunjayz:
To discover voice avatars of other profile owners by hovering(desktop) or tapping (touchscreen) on their avatar of the dashboard's leaderboard card for user list.


mritunjaygoutam12:
Backend code is done.
frontend WIP

mritunjayz:
ready for review @Gregoor

server/src/lib/model/leaderboard.ts
```diff
@@ -4,6 +4,14 @@ import { getLocaleId } from './db';
 import { getConfig } from '../../config-helper';
 import lazyCache from '../lazy-cache';
 import { getMySQLInstance } from './db/mysql';
+import Bucket from '../bucket';
+import { S3 } from 'aws-sdk';
+import { AWS } from '../aws';
+import Model from '../model';
+
+const s3: S3 = AWS.getS3();
+const model: Model = new Model();
```
Gregoor: explicit type annotation isn't necessary here or in the line above I think, as they can be inferred

web/src/components/pages/dashboard/stats/leaderboard-card.tsx
```diff
@@ -88,10 +93,29 @@ class UnconnectedLeaderboard extends React.Component<Props, State> {
...
+    if (!this.state.avatarClipPlaying) {
+      this.setState({ clipPosition: position });
+
+      const audio = new Audio(clipUrl);
+      this.setState({ avatarClipPlaying: true });
+      audio.play();
+
+      audio.onended = () => {
+        this.setState({ avatarClipPlaying: false });
+        this.setState({ clipPosition: null });
```
Gregoor: can we group those calls? i.e. `setState({ a: false, c: null})`

web/src/components/pages/dashboard/stats/leaderboard-card.tsx
```diff
@@ -44,12 +45,16 @@ const FetchRow = (props: React.HTMLProps<HTMLButtonElement>) => (
 interface State {
   rows: { position: number; [key: string]: any }[];
   isAtEnd: boolean;
+  avatarClipPlaying: boolean;
+  clipPosition: number;
```
Gregoor: I think the state could be reduced to a single property, `playingClipIndex?: number`. And if it's null, no avatar clip is playing. Or am I missing something?

web/src/components/pages/dashboard/stats/leaderboard-card.tsx
```diff
@@ -118,8 +142,14 @@ class UnconnectedLeaderboard extends React.Component<Props, State> {
...
           </div>
-          <div className="avatar-container">
+          <div
+            className="avatar-container"
+            title="Click to play avatar"
+            onClick={() =>
+              this.playAvatarClip(row.avatarClipUrl, row.position)
+            }>
             <Avatar url={row.avatar_url} />
+            {clipPosition === row.position ? <PlayIcon /> : ''}
```
Gregoor: nit, preferred:
```jsx
{clipPosition === row.position && <PlayIcon />}
```

web/src/components/pages/dashboard/stats/leaderboard-card.tsx
```diff
@@ -118,8 +142,14 @@ class UnconnectedLeaderboard extends React.Component<Props, State> {
             {row.position < 9 && '0'}
             {row.position + 1}
           </div>
-          <div className="avatar-container">
+          <div
```
Gregoor: Can we make this a button? That would make it more accessible for screen-readers and also reachable with keyboards.

server/src/lib/model/leaderboard.ts
mritunjayz: sorry?

Gregoor: Sorry, I'll try again:

TypeScript can infer the type of `s3` from the `AWS.getS3()` call, so you don't have to explicitly write `s3: S3` (and same for `model`).

mritunjayz: :thumbsup:

web/src/components/pages/dashboard/stats/leaderboard-card.tsx
mritunjayz: Yes! It 's works fine.

mritunjayz:
changes done !

Gregoor:
Awesome, thanks for addressing everything so thoroughly!

One more thing: Can we use the `isProduction()` flag again to "hide" it on prod? It might require us to branch the rendering of `<Avatar>` to use the old container.

Gregoor:
Oh another "one more thing", it also shouldn't be a button if the user has no avatar clip!

PR #2077, opened 2019-06-19
Validated sentences export - 2019-06-19

MichaelKohler:


gion-andri:
@MichaelKohler This export seems to include quite many strings for rm-sursilv. We only have about 2000 approved strings; this export includes about 5500 strings. This seems the number of all strings uploaded to the sentence collector.

MichaelKohler:
@gion-andri thanks for catching this, that would have been very bad. Definitely my fault!

PR #2078, opened 2019-06-19
2019-06-19 Validated SC export

MichaelKohler:


mozillakab:
may be someone could merge.
We're still waiting for new sentences.

nukeador:
Sorry, we are currently in a work week so response times are going to be longer than usual ;)

PR #2079, opened 2019-06-19
Sentence Collector

4746:
Update sentence collector uk


MichaelKohler:
@4746 where are these sentences coming from? I can't find them in the sentence collector and I've done an export a few hours ago, which also includes a few uk sentences. If those are new sentences, please add them through https://common-voice.github.io/sentence-collector/#/ so they can get reviewed.

Thanks!

Michael

4746:
Ok

PR #2090, opened 2019-07-01
fixed register section layout response

mritunjayz:


Gregoor:
Is this related to some issue and if not where does the problem occur?

mritunjayz:
Sorry, for the no description!
Yeah, problem in width on home page of register section if you are not login. https://voice.mozilla.org/en

mritunjayz:
didn't you get this ? @gregoor

Gregoor:
I did, just had a busy week. Thanks for the patience. 

Just looked at it and I don't think it does what we want. This limits the spread of the background waves. We do want them to be fullscreen.

rileyjshaw:
The issue this addresses was fixed in https://github.com/mozilla/voice-web/pull/2255.

PR #2106, opened 2019-07-14
Updated outdated FAQ item text

Adrijaned:
The FAQ item in question is "When will you release Common Voice data in other languages?", updated was according to my best knowledge and capabilities. Untested, made only in GitHub web editor for reasons. Probably squash before merging.


Gregoor:
Awesome, thanks for reaching out and taking a stab at copy.

@mbransn shall we let Jay take a stab at it as well or are we happy with this new copy?

mbransn:
Just one change and then I'm good for merge -- update the last sentence to be: 

> New languages are added to Common Voice for voice contribution when 5000 approved sentences have been collected.

Gregoor:
Awesome, thanks you two!

PR #2129, opened 2019-07-30
Use a non t2 instance type for production

The-smooth-operator:


The-smooth-operator:
@gozer for a sanity check

Issue #2133, opened 2019-07-31
Voice Avatar Setup polish


Gregoor:
- [x] missing buttons after recording (https://app.zeplin.io/project/5a9eec5c53a9a2685450d4be/screen/5d12406cd17e52705b13c595)
- [x] missing icons on the buttons: mic icon for "create voicewave", stop icon for "recording"
- [x] counter style not like in the designs (https://app.zeplin.io/project/5a9eec5c53a9a2685450d4be/screen/5d12406ad17e52705b13c452)
- [x] missing screen for when user already has voice wave (https://app.zeplin.io/project/5a9eec5c53a9a2685450d4be/screen/5d124066bcecab70f893236b)

Gregoor:
@mritunjayz 

- [ ] counter style doesn't match Zeplin yet
- [ ] microphone request should come before countdown starts counting
- [ ] no play button after recording voice wave
- [ ] icon colors are off

Gregoor:
- [ ] missing separator between voice avatar recording and avatar image (https://app.zeplin.io/project/5a9eec5c53a9a2685450d4be/screen/5d1240645259df7049af5d71)
- [ ] preload voice wave when visiting the avatar page, so that content doesn't jump later

moz-bozden:
Closing as stale.

PR #2137, opened 2019-07-31
play for already clip url

mritunjayz:
WIP
#2133


mritunjayz:
WIP

web/src/components/pages/profile/avatar-setup/avatar-setup.css
```diff
@@ -76,6 +113,23 @@
         padding-left: 22px;
     }
 
+    .recreate-voice {
+        width: 470px;
+        height: 58px;
+        font-family: OpenSans;
+        font-size: 16px;
```
Gregoor: It looks like these values are more or less verbatim from Zeplin. I'd recommend not doing that, as many of these are just the default values and create noise in the code. And for some others we also have variable we could use here. For sizes it's often preferable to keep things dynamic, so that they respond to different screen sizes.

web/src/components/pages/profile/avatar-setup/avatar-setup.tsx
```diff
@@ -309,39 +339,65 @@ class AvatarSetup extends React.Component<Props> {
...
-              />
+                onClick={this.uploadAvatarClip.bind(this)}>
+                <ShareIcon />
+                <Localized id="ready-to-upload"></Localized>
+              </Button>
+            </div>
+          )}
+          {avatarClipUrl === 'empty' ? (
+            <Localized id="about-avatar-clip">
+              <p className="Create-a-custom-voic" />
```
Gregoor: nit: typo
nit2: let's stick to lowercase class names

web/src/components/pages/profile/avatar-setup/avatar-setup.css
mritunjayz: Ah, I was thinking I have to keep little sync in code with Zeplin. I understand !

Gregoor: I see, sorry for the confusion and thanks for the fix!

Issue #2146, opened 2019-08-04
picture change ui problem on persian


geeksesi:
see this screenshot

![image](https://user-images.githubusercontent.com/28778964/62423693-fddfbb00-b6d8-11e9-9bc0-4e70b71a92f6.png)

also please change persian font ( open source [Vazir font](https://github.com/rastikerdar/vazir-font) is good )

PR #2151, opened 2019-08-06
Avatar setup4

mritunjayz:


mritunjayz:
also includes #2146

web/src/components/pages/profile/avatar-setup/avatar-setup.css
```diff
@@ -127,22 +131,35 @@
...
-        width: 100%;
+        width: 508px;
     }
     .start-in {
         position: relative;
-        top: -8px;
+        top: -12px;
         width: 39px;
         height: 18px;
+        font-family: ;
```
Gregoor: this looks a bit like zeplin boilerplate, do we need all these rules?

mritunjayz: yes, this is affecting the style if not mentioned !

PR #2158, opened 2019-08-09
minor avatar setup cleanups

Gregoor:


web/src/components/pages/profile/avatar-setup/avatar-setup.tsx
```diff
@@ -96,19 +95,28 @@ interface PropsFromDispatch {
...
 interface Props extends LocalizationProps, PropsFromState, PropsFromDispatch {}
 
-class AvatarSetup extends React.Component<Props> {
-  state = {
+interface State {
+  isSaving: boolean;
+  isRecording: boolean;
+  isPlaying: boolean;
+  counter: number;
+  clipStatus: 'fetching' | 'notStarted' | 'started' | 'starting' | 'recorded';
```
Gregoor: That way we don't actually set an invalid status

web/src/components/pages/profile/avatar-setup/avatar-setup.tsx
```diff
@@ -96,19 +95,28 @@ interface PropsFromDispatch {
...
+class AvatarSetup extends React.Component<Props, State> {
+  state: State = {
     isSaving: false,
-    recordingStatus: false,
-    avatarClipPlaying: false,
+    isRecording: false,
+    isPlaying: false,
     counter: 3,
-    clipStatus: 'notStarted',
+    clipStatus: 'fetching',
```
Gregoor: By starting in this state we don't have flickering UI caused by a clip being fetched (and the status then changing)

web/src/components/pages/profile/avatar-setup/avatar-setup.tsx
```diff
@@ -96,19 +95,28 @@ interface PropsFromDispatch {
...
+  avatarClipUrl: string | null;
+}
+
+class AvatarSetup extends React.Component<Props, State> {
+  state: State = {
     isSaving: false,
-    recordingStatus: false,
-    avatarClipPlaying: false,
+    isRecording: false,
+    isPlaying: false,
```
Gregoor: more consistent boolean naming

web/src/components/pages/profile/avatar-setup/avatar-setup.tsx
```diff
@@ -205,9 +212,7 @@ class AvatarSetup extends React.Component<Props> {
     this.recordingStartTime = Date.now();
     this.recordingStopTime = 0;
     this.setState({
-      // showSubmitSuccess: false,
-      recordingStatus: true,
-      error: null,
```
Gregoor: we didn't use the error state, so I discarded it

web/src/components/pages/profile/avatar-setup/avatar-setup.tsx
```diff
@@ -265,7 +268,7 @@ class AvatarSetup extends React.Component<Props> {
   };
 
   private updateAvatarClip = () => {
-    this.setState({ avatarClipUrl: 'empty', counter: 3 });
```
Gregoor: let's use `null` instead for that case

web/src/components/pages/profile/avatar-setup/avatar-setup.tsx
```diff
@@ -304,6 +301,8 @@ class AvatarSetup extends React.Component<Props> {
       },
     };
 
+    const hasClip = avatarClipUrl !== null;
```
Gregoor: more explicit name for a condition we're using multiple times

PR #1064, opened 2018-06-06
French accents

TheoChevalier:
R? @lissyx 
cc @mikehenrty

PR #2162, opened 2019-08-14
Add breton accents

tornoz:
Here is a list for breton accents (https://en.wikipedia.org/wiki/Breton_language#Geographic_distribution_and_dialects).
I took the #1064 PR as an example, tell me if I missed something.


phirework:
This looks great, thanks!

Issue #2178, opened 2019-08-22
Voice avatar recording should not be available on mobile


viorelaioia:
**Env:**
iOS 12.4, Firefox an Safari

**STR:**

1. Login to CV staging
2. Navigate to Settings - Avatar

**Expected:**
Voice avatar recording should not be available, as recording is not working on mobile

**Actual:**
"Recreate Voicewave" button is available, even though recording is not working.

Gregoor:
cc @mritunjayz

mritunjayz:
Yeah, I see. Thanks @viorelaioia

mritunjayz:
@Gregoor @dabinat @viorelaioia voice avatar should not be available on mobile which does not support recording? And what should be the screen then without recording?

mritunjayz:
Actually, I am asking what should be the title then. Should it be same  "Add an avatar to your profile and say “hello!” " ?

rleitan:
> Actually, I am asking what should be the title then. Should it be same "Add an avatar to your profile and say “hello!” " ?

@mbransn

mbransn:
@mritunjayz @Gregoor first -- which mobile browsers aren't supporting recording?

mritunjayz:
I only know about Firefox but I think these are all iOS 12.4, Firefox and Safari as mentioned by @viorelaioia

mbransn:
Got it, forgot that it's not Sept. yet and Apple hasn't released the new update to hopefully rectify this. So it'd be best to just hide the voice avatar divs until they are functional on iOS. Once they're functional we'll reveal them. @Gregoor thoughts?

Gregoor:
All iOS browsers, Safari and Edge don't support recording yet. And both Edge & Safari beta do, so salvation is imminent :)
Agree with megan wrt hiding them.

Gregoor:
fixed in #2182

PR #2182, opened 2019-08-22
no recording for mobile screen

mritunjayz:


mritunjayz:
solves #2178

web/locales/en/messages.ftl
```diff
@@ -324,6 +324,7 @@ off = Off
 on = On
 add-avatar-title = Add an avatar to your profile
 add-avatar-clip = Add an avatar to your profile and say “hello!”
+avatar-clip-title = An avatar to your profile!
```
Gregoor: this english looks incomplete

web/src/components/pages/profile/avatar-setup/avatar-setup.css
```diff
@@ -26,9 +26,23 @@
         left: 15%;
         right: 15%;
         padding-top: 10px;
-        @media only screen and (max-width: 375px) {
+
+        @media only screen and (max-width: 415px) {
             left: 6%;
             right: 6%;
+            display: none;
```
Gregoor: I don't think we want to actually hide the recording on mobile in general. Rather only on browsers that don't support recording

web/locales/en/messages.ftl
mritunjayz: I just wanna ask this after opening a PR ! that what should I show as the title if no recording option is available !

mritunjayz: No, it is not from copy doc

mritunjayz:
I think for now we can go for this until the design is not ready for it

Issue #2193, opened 2019-08-26
Please change the main title Profile/Avatar 


mbransn:
Instead of "Add an avatar to your profile and say “hello!”  please change to "Give your avatar a voice"

@mritunjayz @rleitan

mritunjayz:
I was little busy. Now , I will tackle these issues.

mritunjayz:
close it !

Issue #2192, opened 2019-08-26
Add line separator between the voice waves and right column Profile/Avatar


mbransn:
<img width="543" alt="Screenshot 2019-08-23 00 23 32 3" src="https://user-images.githubusercontent.com/18686986/63713481-9955de80-c7f4-11e9-80a8-025167d8f28a.png">

cc @mritunjayz @rleitan

mritunjayz:
close it !

PR #2194, opened 2019-08-28
change voice-wave avtar title

mritunjayz:
#2193


mritunjayz:
#2192

web/locales/en/messages.ftl
```diff
@@ -322,7 +322,7 @@ skip-submission-note = Note: You will still need to select between Speak or List
 off = Off
 on = On
 add-avatar-title = Add an avatar to your profile
-add-avatar-clip = Add an avatar to your profile and say “hello!”
+add-avatar-clip = Give your avatar a voice
```
Gregoor: can we change the id of this string? that way communities in other languages will be prompted to add a new string

mritunjayz: Yeah, right

Gregoor:
grazie :) 🚀

Issue #2198, opened 2019-08-30
Stop button continues to play my voice avatar


viorelaioia:
**STR:**

1. Record a voice avatar
2. Play the voice avatar
3. While the avatar is playing, click the Stop button, in order to stop playing

**Expected:**
The voice avatar stops playing

**Actual:**
The voice avatar plays until the recording is over, clicking the Stop button has no effect.

mritunjayz:
close it !

Issue #2187, opened 2019-08-26
Hover action on voice avatar


mbransn:
When the voice wave is created in the menu and dashboard, all the voice waves needs to start on the hover (rather than click).

cc @mritunjayz @rleitan

mritunjayz:
I have added hover action instead of onclick but should it be in avatar setup page too ? @mbransn

mritunjayz:
should close it !

PR #2199, opened 2019-09-01
on hover play for audio instead of click

mritunjayz:


mritunjayz:
#2187 #2198

web/src/components/layout/user-menu.tsx
```diff
@@ -38,23 +40,31 @@ export default function UserMenu() {
...
             trackVoiceAvatar('self-listen', locale);
             if (account.avatar_clip_url !== null && !showAnimation) {
-              const audio = new Audio(account.avatar_clip_url);
-              audio.play();
+              audioRef.current.src = account.avatar_clip_url;
+
+              audioRef.current.play();
               setShowAnimation(!showAnimation);
-              audio.onended = () => {
+              audioRef.current.onended = () => {
```
Gregoor: Why not attach these event listeners in the JSX?

web/src/components/layout/user-menu.tsx
```diff
@@ -38,23 +40,31 @@ export default function UserMenu() {
         className={'user-menu ' + (showMenu ? 'active' : '')}
         onMouseEnter={() => setShowMenu(true)}
         onMouseLeave={() => setShowMenu(false)}>
+        <audio preload="auto" ref={audioRef} />
         <button
           className="toggle"
           title="click to play avatar"
-          onClick={() => {
```
Gregoor: For mobile we'd still need tap to listen, so I think we'd wanna have both mouseEnter and click

web/src/components/layout/user-menu.tsx
mritunjayz: yeah, we can

PR #1275, opened 2018-07-19
Create copus-daily-lang.txt

belkacem77:
Some sentences containing kabyle public people names: Authors, Singers, Politicians, with birth and place dates, work... Daily language.

Issue #1257, opened 2018-07-12
add UI feedback when using yes/no keyboard shortcuts


Djfe:
shortcuts shouldn't only blink green/red when being clicked by cursors but also when shortcuts are used.  
the easiest solution might be to use the DOM and execute the click method on a button, which in return should also trigger the css, right?

Gregoor:
Thanks for raising this, that feedback would indeed be valuable. Unfortunately it wouldn't be as simple as simulating a click as as that style is currently only applied on hover.

Undigon:
Well, we can try using ```<input type="checkbox">``` and a bit of CSS+JS to implement onClic styling. We should then use some method to remove the checked property once the phrase has been updated. Is there any listener for that or should we apply a timed delay?

I've tried doing something like that before so I know it could be done (changing the colour "on checked" at least).

For reference, I got the idea some months ago from here: [Codepen example](https://codepen.io/oloman/pen/odnqy).

zecakeh:
Also there should be visual feedback when you click/use the shortcut for YES but the sentence hasn't been played yet. So that people know it's because the full length of the audio file hasn't been played yet.

moz-bozden:
Closing as stale...

PR #2219, opened 2019-09-17
Visual feedback for shortcuts (Fixes #1275)

xdumaine:
Add a visual indicator for buttons triggered with keyboard shortcuts.

Yes/No/Skip

Yes/No:

shortcut-feedback

Skip (Listen and Speak):

shortcut-feedback-2

I also fixed some broken styling around the disabled states of the yes/no buttons. I have not yet added anything to address the comment on the thread about attempting to click/trigger Yes before the clip is ended. That might need some more design thought.

### DIFF
--- a/web/src/components/pages/contribution/contribution.tsx
+++ b/web/src/components/pages/contribution/contribution.tsx
@@ -105,7 +105,7 @@ interface Props extends LocalizationProps, PropsFromState {
   shortcuts: {
     key: string;
     label: string;
-    action: () => any;
+    action: (e: KeyboardEvent) => any;
   }[];
   type: 'speak' | 'listen';
   skipTriggered?: boolean;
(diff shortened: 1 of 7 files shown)


web/src/components/pages/contribution/contribution.tsx
```diff
@@ -242,11 +255,11 @@ class ContributionPage extends React.Component<Props, State> {
       onSubmit,
       type,
     } = this.props;
-
     if (
       event.ctrlKey ||
       event.altKey ||
       event.shiftKey ||
+      event.metaKey ||
```
xdumaine: I noticed that using `cmd-r` to refresh on mac was triggering a record instead of a refresh.

web/src/components/vars.css
```diff
@@ -8,6 +8,7 @@
     --near-black: #4a4a4a;
     --grey: #e7e5e2;
     --darker-grey: #e6e5e3;
+    --disabled-grey: #cacaca;
```
xdumaine: This color is the browser's default disabled color grey on buttons, and it looks good to me, where the other already defined greys did not.

Gregoor:
Great stuff, thanks! This also fixes #1257!

Unfortuantely I don't have time to review atm.

PR #2220, opened 2019-09-17
Add better support for localhost dev without setting up an s3 account

xdumaine:
This works for me, though I’m not sure if it’s the best solution long term, since it adds s3proxy to the CSP and requires a host file change. The host file change is still far easier than setting up an s3 bucket and creds though. It seems that the puppet config overwrites the CSP anyway, so that part might not be a problem?

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -18,6 +18,9 @@ services:
       - voice-web
     environment:
       - S3PROXY_AUTHORIZATION=none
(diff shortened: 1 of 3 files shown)


skanderbenalaya:
I'm having issues with CSP blocking mp3 files, i tried your solution but it didn't work for me.
Is there any other workaround this issue with the new state of the project ?

PR #2225, opened 2019-09-20
Fix #1295: Contribute more button rerendering

### LINKED ISSUE #1295: "Ready to do 5 more" blinks when hovered and sending clips
Every time the progress bar progresses on top (so I guess every time a clip is sent), the "Ready to do 5 more?" button blinks when it is hovered.
It doesn't look good and it also prevents people from actually clicking on the button before all the clips are sent.

colelawrence:
This addresses [the rerendering issue](https://github.com/mozilla/voice-web/issues/1295) affecting the "Ready to do 5 more" button on the success page.

I approached this by narrowing down the cause of the render, which is the result of several functions being re-declared on every render. So, by passing direct references to the properties ".resetState" and ".upload" in speak.tsx and also moving the component's creation outside of the `Success` component's render cycle, we are helping React reconcile the component properly.

This strategy might surface several other opportunities for improving performance of these components (by unnesting component functions) if that has ever been an issue.

`/en/speak`
image

### DIFF
--- a/web/src/components/pages/contribution/success.tsx
+++ b/web/src/components/pages/contribution/success.tsx
@@ -31,6 +31,26 @@ const GoalPercentage = ({
   </span>
 );
 
+const ContributeMoreButton = (props: {
+  hasAccount: boolean;
+  children: React.ReactNode;
+  onClick: () => any;
+}) =>
+  props.hasAccount ? (
+    <Button
+      className="contribute-more-button"
+      rounded
+      onClick={props.onClick}
+      children={props.children}
+    />
+  ) : (
+    <TextButton
+      className="contribute-more-button secondary"
+      onClick={props.onClick}
+      children={props.children}
+    />
+  );
+
 function Success({
   getString,
   onReset,
@@ -90,22 +110,6 @@ function Success({
   const finalPercentage = Math.ceil(
     (100 * (contributionCount || 0)) / goalValue
   );
-
-  const ContributeMoreButton = (props: { children: React.ReactNode }) =>
-    hasAccount ? (
-      <Button
-        className="contribute-more-button"
-        rounded
-        onClick={onReset}
-        {...props}
-      />
-    ) : (
-      <TextButton
-        className="contribute-more-button secondary"
-        onClick={onReset}
-        {...props}
-      />
-    );
 
   const goalPercentage = (
     <GoalPercentage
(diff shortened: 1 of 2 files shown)


web/src/components/pages/contribution/success.tsx
```diff
@@ -31,6 +31,26 @@ const GoalPercentage = ({
   </span>
 );
 
+const ContributeMoreButton = (props: {
+  hasAccount: boolean;
+  children: React.ReactNode;
+  onClick: () => any;
+}) =>
+  props.hasAccount ? (
```
mritunjayz: are we getting hasAccount as props? may be not !

web/src/components/pages/contribution/success.tsx
```diff
@@ -31,6 +31,26 @@ const GoalPercentage = ({
...
+    <Button
+      className="contribute-more-button"
+      rounded
+      onClick={props.onClick}
```
mritunjayz: I think it should be props.reset

web/src/components/pages/contribution/success.tsx
colelawrence: It was moved into the props here so we can declare `ContributeMoreButton` outside the scope of the parent function component.

This prevents us from redefining the `ContributeMoreButton` every time the parent function component renders.

web/src/components/pages/contribution/success.tsx
colelawrence: Are you suggesting that the prop be named `reset` instead of `onClick`?

web/src/components/pages/contribution/success.tsx
mritunjayz: Ah, got it below. Nice

web/src/components/pages/contribution/success.tsx
mritunjayz: yes, what do you think ?

colelawrence: I think that `onClick` is more accurate in the scope of this component. If we called this `reset`, then we would be leaking responsibilities of the parent component down into this component. On it's own, the ContributeMoreButton is just a button, and the user of the button can listen to "onClick", but it makes less sense to listen to "reset."

Does that make sense?

rileyjshaw:
Thanks for the fix @colelawrence! Sorry for the delay in accepting this.

PR #2227, opened 2019-09-23
Remove 12k wiki sentences containing duplicates, foreign and scientific terms

dabinat:
12,000 sentences were removed from the English wiki dump containing:

-  Duplicates where the only difference was capitalization
-  Partial sentences
-  Difficult-to-pronounce scientific terms (e.g. "Dromaeosauridae")
-  Sentences containing character sequences not commonly used in English, which are indicative of foreign names (e.g. "Kiirunavaara")
-  Apostrophe patterns not commonly used in English (e.g. "V'tosh ka'tur")

It also tidies up some of the sentences by removing errant punctuation and replacing abbreviations with the full word.


nukeador:
Thanks for this.

PR #2228, opened 2019-09-24
add event name to avoid audio play on mouse leave as the playAvatar f…

mritunjayz:
add event name to avoid audio play on mouse leave as the playAvatar function is the same for onEnter and onLeave event


Gregoor:
Accidently merged this one too early (i.e. in a broken state). Could you open a new PR once its unbroken?

PR #2232, opened 2019-09-26
Consolidate resources in a single account.

The-smooth-operator:
These recourse were created in a separated AWS account. Now that Gregor is steppping out, we want to have everything in the same account

### DIFF
--- a/nubis/terraform/main.tf
+++ b/nubis/terraform/main.tf
@@ -154,3 +154,49 @@ module "cache" {
   client_security_groups = "${module.worker.security_group},${module.sync.security_group}"
   engine                 = "redis"
 }
+
+module "cv_mandarin" {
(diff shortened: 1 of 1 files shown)


The-smooth-operator:
@Gregoor before we run this through Gozer, can you please have a quick sanity check?
If working as expected, this changes will create:
 - 2 S3 buckets
 - 1 t2.small EC2 instance in the same VPC where the DB runs.

The-smooth-operator:
Hey @gozer can you please take a look if this looks good?

MichaelKohler:
What's the status on this one?

The-smooth-operator:
I'd say this one is deprecated. I'm going to talk to Gregor for clarifying if we have something to do, in that case I'll open a new PR.

PR #2234, opened 2019-09-28
Updated description text

Adrijaned:
Updated description text for languages in development on the Languages page, to reflect the actual contents of the cards displayed.


phirework:
Hi @Adrijaned, thanks for submitting this PR! We will actually be revisiting the entire UX on that page, and since every new string in the localization docs needs to be re-submitted to Pontoon to be translated into all the other languages, it doesn't make sense right now to go through that effort when it's entirely possible the whole page will change.

PR #2243, opened 2019-10-14
Implemented onboarding modal

catherine917:
In this pull request, I've made the following changes:

- Add onboarding modal 

- You can import it into any page and trigger it for testing


web/src/components/onboarding-modal/onboarding-modal.tsx
```diff
@@ -0,0 +1,87 @@
+import * as React from 'react';
+import { useState } from 'react';
+import { Button } from '../ui/ui';
+import { PlayOutlineGreenIcon, MicIcon, SkipIcon } from '../ui/icons';
+import Modal from '../modal/modal';
+import URLS from '../../urls';
+import { LocaleLink } from '../locale-helpers';
+import './onboarding-modal.css';
+
+const STEPS: Array<any> = [
```
rileyjshaw: Not a blocker, but it would be nice to add a type here.

```
interface StepTypeWithoutImage {
  title: string;
}
interface StepTypeWithImage extends StepTypeWithoutImage {
  img: string;
  alt: string;
}
type StepType = StepTypeWithoutImage | StepTypeWithImage;
```

web/src/components/onboarding-modal/onboarding-modal.tsx
```diff
@@ -0,0 +1,87 @@
...
+  {
+    title: 'Track your team and indvidual progress on the challenge dashboard',
+    img: './images/dashboard.png',
```
rileyjshaw: These `img` values need to be wrapped in a `require` call to resolve to the correct path.

```
// ...
img: require('./images/dashboard.png'),
// ...
```

web/src/components/onboarding-modal/onboarding-modal.tsx
```diff
@@ -0,0 +1,87 @@
...
+    alt: 'dashboard',
+  },
+  {
+    title: 'Invite friends and colleagues to join your challenge team',
+    img: './images/invite.jpg',
```
rileyjshaw: https://github.com/mozilla/voice-web/pull/2243/files#r335200438

web/src/components/onboarding-modal/onboarding-modal.tsx
```diff
@@ -0,0 +1,87 @@
...
+    alt: 'invite',
+  },
+  {
+    title:
+      'Earn points by contributing your voice and validating recorded clips',
+    img: './images/contribute.png',
```
rileyjshaw: See above (https://github.com/mozilla/voice-web/pull/2243/files#r335200438)

web/src/components/onboarding-modal/onboarding-modal.tsx
```diff
@@ -7,37 +7,50 @@ import URLS from '../../urls';
...
     title:
       'Earn points by contributing your voice and validating recorded clips',
-    img: './images/contribute.png',
+    img: require('./images/contribute.png'),
     alt: 'contribute',
   },
   {
     title: 'Ready to get started?',
+    img: '',
+    alt: '',
```
rileyjshaw: Typescript allows optional types like so:

```
interface Step {
  title: string;
  img?: string;
  alt?: string;
}
```

However, I think the type definition I gave in my previous comment was even more descriptive:

```
interface StepTypeWithoutImage {
  title: string;
}
interface StepTypeWithImage extends StepTypeWithoutImage {
  img: string;
  alt: string;
}
type StepType = StepTypeWithoutImage | StepTypeWithImage;
```

So, I think we can do without the blank `img` and `alt` fields here.

catherine917: @rileyjshaw Yep, I have tried this way. But in the below code. But when I use `data.img` or `data.alt` it will throw the error `Property 'img' does not exist on type 'StepType'.
 Property 'img' does not exist on type 'StepTypeWithoutImage'`

PR #2246, opened 2019-10-15
Add a welcome modal for the Open Voice Challenge

rileyjshaw:
This commit adds a skeleton welcome modal for the upcoming partner challenge. Because the Open Voice Challenge is not yet underway, the modal is gated behind an `!isProduction()` check.

Screen Shot 2019-10-15 at 12 24 22 PM

NOTE: The modal still needs to be:

- linked to an updated T&C, and
- hooked up to a working signup endpoint.

Test plan:

- Navigate to localhost:9000/?ovchall=1

Modal should pop up. Button should be disabled.

- Check the T&C box.

Button should be enabled.

- Click the button.

Page should redirect to /login.

### DIFF
--- a/web/src/components/layout/layout.tsx
+++ b/web/src/components/layout/layout.tsx
@@ -32,6 +32,7 @@ import Nav from './nav';
 import UserMenu from './user-menu';
 import * as cx from 'classnames';
 import { isStaging } from '../../utility';
+import WelcomeModal from '../welcome-modal/welcome-modal';
 
 const LOCALES_WITH_NAMES = LOCALES.map(code => [
   code,
@@ -57,6 +58,7 @@ interface LayoutState {
   hasScrolled: boolean;
   hasScrolledDown: boolean;
   showStagingBanner: boolean;
+  showWelcomeModal: boolean;
 }
 
 class Layout extends React.PureComponent<LayoutProps, LayoutState> {
@@ -69,6 +71,7 @@ class Layout extends React.PureComponent<LayoutProps, LayoutState> {
     hasScrolled: false,
     hasScrolledDown: false,
     showStagingBanner: isStaging(),
+    showWelcomeModal: this.props.location.search.includes('ovchall=1'),
   };
 
   componentDidMount() {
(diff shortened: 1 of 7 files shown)


web/src/components/layout/layout.tsx
```diff
@@ -162,6 +166,17 @@ class Layout extends React.PureComponent<LayoutProps, LayoutState> {
...
       <div id="main" className={className}>
+        {showWelcomeModal && (
+          <WelcomeModal
+            onRequestClose={() => {
+              this.setState({ showWelcomeModal: false });
+            }}
+            onClick={() => {
+              window.location.pathname = '/login';
+            }}
+            team="SAP"
```
rileyjshaw: Note: eventually this will be determined by query strings, but for now I just hardcoded it.

web/src/components/layout/layout.tsx
```diff
@@ -162,6 +166,17 @@ class Layout extends React.PureComponent<LayoutProps, LayoutState> {
 
     return (
       <div id="main" className={className}>
+        {showWelcomeModal && (
+          <WelcomeModal
+            onRequestClose={() => {
+              this.setState({ showWelcomeModal: false });
+            }}
+            onClick={() => {
+              window.location.pathname = '/login';
```
rileyjshaw: I went through several iterations of fancy react-router magic before landing on this. If you have any examples of a cleaner redirect to uncontrolled routes, please let me know!

Also: we probably need to pass information with this redirect, to ensure the extra Open Voice Challenge fields (eg. `team`) get populated correctly. We're working on a solution as the backend comes together, but any input is appreciated.

Gregoor: Oh I'd go with a regular ol' `<a>`. We do pass some state around for login (namely locale and when changing the email the old_email) here: https://github.com/mozilla/voice-web/blob/master/server/src/auth-router.ts#L130-L140

Hope that helps!

rileyjshaw: @Gregoor the button needs to be disabled until the terms & conditions box is checked; how would you recommend handling the disabled state? I thought there may be some a11y concerns with trying to fake a "disabled" link by changing the href, but yeah, I'm not happy with the button solution either… 😕 If you know of anywhere else on the site where this pattern is followed I'd appreciate a link!

hidde: `aria-disabled` could work to convey state, swapping button for link seems not so great as I suspect it would unexpectedly change semantics (unexpectedness can affect certain users disproportionately)

Gregoor: Great, thankslots for the guidance!

PR #2248, opened 2019-10-16
Add Open Voice Challenge option to the info page

rileyjshaw:
This commit updates privacy and mailing list options on the
/profile/info page.

Screen Shot 2019-10-15 at 5 13 26 PM

NOTE: As communicated in the TODOs, this commit is not ready for
production.

Test plan:

- Remove the `false && ...` condition from /web/src/components/pages/profile/info/info.tsx
- Visit localhost:9000/en/profile/info

The new option should appear, with new styles.


web/locales/en/messages.ftl
```diff
@@ -8,6 +8,8 @@ email-input =
 submit-form-action = Submit
 loading = Loading…
 email-opt-in-info = I'd like to receive emails such as goal reminders, my progress updates and newsletters about Common Voice.
+email-opt-in-info-title = Join the Common Voice mailing list
+email-opt-in-info-sub = Receive emails such as goal reminders, progress updates and newsletters about Common Voice.
```
rileyjshaw: I'm still learning how the localization system works. As you can see, this is _very_ similar to the `email-opt-in-info` string above. When an edit like this is needed, if I update the initial value will changes eventually propagate to new locales? Or is it necessary to create a new ID like I did here?

web/src/components/pages/profile/info/info.css
```diff
@@ -124,6 +132,12 @@
                 text-decoration: none;
             }
         }
+
+        a.terms {
```
rileyjshaw: I'm writing some pretty loosey-goosey CSS, so let me know if there's a style guide / ordering preference / etc that you were adhering to. One example that comes to mind is that I'm omitting leading `&`s where possible.

web/src/components/pages/profile/info/info.tsx
```diff
@@ -323,25 +323,52 @@ function ProfilePage({
             </Tooltip>
 
             <div className="checkboxes">
+              {/* TODO(riley): Once Open Voice Challenge backend is set up,
+                               only show this to enrolled users. */}
+              {/* TODO(riley): Connect `onChange` and `checked` to the user
+                               model. */}
+              {false && (
```
rileyjshaw: 😅 I don't even want to put this behind `isProduction` yet...

web/src/components/pages/profile/info/info.tsx
```diff
@@ -323,25 +323,52 @@ function ProfilePage({
...
+                    </Localized>
+                  </>
                 }
                 onChange={handleChangeFor('sendEmails')}
                 checked={sendEmails}
               />
 
               {!user.account && !isSubmitted && (
-                <React.Fragment>
+                <>
```
rileyjshaw: Any reason to use `React.Fragment` over `<>`?

Gregoor: I think the TS compiler needed a little longer to be ready for it, so no good reason not to use `<>` now :)

web/src/components/pages/profile/info/info.css
Gregoor: I've been a loose goose myself, as you might've noticed. Had some ambitions about that a year ago, but that's about it in terms of style guide:
https://github.com/mozilla/voice-web/issues/912

web/locales/en/messages.ftl
Gregoor: Unfortunately we do need new IDs when we want the localization community to notice. So you did the right thing!

web/src/components/pages/profile/info/info.css
rileyjshaw: Oh cool, thanks for the link!

PR #2249, opened 2019-10-16
creating the concept of joinable teams

can-007:
Just 1 migration to create 2 table: teams and user_client_teams.


can-007:
I have doubts on the origin of teams: who created them? How they are managed? Are they fixed and can only be updated manually in DB, or are they dynamically created from the web page?

rileyjshaw:
@can-007 great question! Teams will be created by us manually in the DB for now. In 2020 we may create a signup flow, but for now we will have under 10 teams that we can create manually.

server/src/lib/model/db/migrations/20191015151914-add-teams-and-user-client-teams.ts
```diff
@@ -0,0 +1,28 @@
+export const up = async function(db: any): Promise<any> {
+  return db.runSql(
+    `
+    CREATE TABLE teams (
+      team_id CHAR(36) NOT NULL PRIMARY KEY,
+      teamname TEXT NOT NULL DEFAULT (''),
+      leader CHAR(36) NOT NULL,
+      invite_link TEXT DEFAULT NULL DEFAULT (''),
+      created_at DATETIME DEFAULT now(),
+      logo_url TEXT DEFAULT NULL DEFAULT (''),
```
rileyjshaw: Great idea! This is a nice touch.

server/src/lib/model/db/migrations/20191015151914-add-teams-and-user-client-teams.ts
```diff
@@ -0,0 +1,28 @@
+export const up = async function(db: any): Promise<any> {
+  return db.runSql(
+    `
+    CREATE TABLE teams (
+      team_id CHAR(36) NOT NULL PRIMARY KEY,
+      teamname TEXT NOT NULL DEFAULT (''),
+      leader CHAR(36) NOT NULL,
```
rileyjshaw: I don't think teams will have leaders in the pilot – for now we can leave this out.

server/src/lib/model/db/migrations/20191015151914-add-teams-and-user-client-teams.ts
```diff
@@ -0,0 +1,28 @@
...
+    );
+    
+    CREATE TABLE user_client_teams (
+      client_id CHAR(36) NOT NULL,
+      team_id CHAR(36) NOT NULL,
+      created_at DATETIME DEFAULT now(),
+      PRIMARY KEY (client_id, team_id),
+      FOREIGN KEY (client_id) REFERENCES user_clients(client_id),
+      FOREIGN KEY (team_id) REFERENCES teams(team_id)
+    );
```
rileyjshaw: `teams` to `user_clients` is a one-to-many relationship, so we could also do this with a foreign key on the user. Sorry if I didn't make that clear! Did you store a junction table like this for safety / efficiency, or do you think a foreign key would be better?

can-007: sorry, I didn't realize it is a one-to-many relationship. My first thought was that maybe in the future, people could join different teams or create multiple teams (and be a role like leader to manage the teams, like changing the logo etc.).

server/src/lib/model/db/migrations/20191015151914-add-teams-and-user-client-teams.ts
```diff
@@ -0,0 +1,28 @@
+export const up = async function(db: any): Promise<any> {
+  return db.runSql(
+    `
+    CREATE TABLE teams (
+      team_id CHAR(36) NOT NULL PRIMARY KEY,
+      teamname TEXT NOT NULL DEFAULT (''),
+      leader CHAR(36) NOT NULL,
+      invite_link TEXT DEFAULT NULL DEFAULT (''),
```
Gregoor: I'd go for a name like `invite_token` since this probably wouldn't/shouldn't hold a whole URL

server/src/lib/model/db/migrations/20191015151914-add-teams-and-user-client-teams.ts
```diff
@@ -0,0 +1,28 @@
+export const up = async function(db: any): Promise<any> {
+  return db.runSql(
+    `
+    CREATE TABLE teams (
+      team_id CHAR(36) NOT NULL PRIMARY KEY,
+      teamname TEXT NOT NULL DEFAULT (''),
```
Gregoor: `team_name` for consistency's sake

server/src/lib/model/db/migrations/20191015151914-add-teams-and-user-client-teams.ts
can-007: what does the token looks like? will it be more than , like, 255 chars?
I am thinking maybe the token is unique and of type varchar(255).

server/src/lib/model/db/migrations/20191017135015-add-challenge-team-to-user-clients.ts
```diff
@@ -0,0 +1,12 @@
+export const up = async function(db: any): Promise<any> {
+  return db.runSql(
+    `
+    ALTER TABLE user_clients ADD COLUMN challenge_team VARCHAR(255) DEFAULT NULL;
+    ALTER TABLE user_clients ADD CONSTRAINT uc_ibfk_1 FOREIGN KEY (challenge_team) REFERENCES teams(team_name);
```
rileyjshaw: This looks fine to me, but I just want to run a quick confidence check. Will the default `NULL` value pass this constraint? Just want to make sure, since this will run across the entire `user_clients` table.

rileyjshaw: (Update: tested it on my end, it works!)

rileyjshaw:
Cool! I just did some testing on my end and this checks out :) excited to finally land this, thanks for the work!

PR #2250, opened 2019-10-16
Add challenge tab to dash board for the Open Voice Challenge

catherine917:
- For now, the challenge tab is visible for all users. Thus I add `isProduction` to hide it. If the backend ready, I will modify it. 

- I reuse `AwardsPage` and add it to the challenge dashboard. Until the challenge page ready, I will replace it.


web/src/components/pages/dashboard/dashboard.css
```diff
@@ -174,3 +173,123 @@
...
+    > .label-team::after {
+        content: 'Team points';
+    }
+    > .divider {
+        height: 100%;
+        width: 1px;
+        background-color: var(--light-grey);
+    }
+}
+.inviteButton {
```
rileyjshaw: Kebab-case is preferred: `invite-button`

web/src/components/pages/dashboard/dashboard.css
```diff
@@ -174,3 +173,123 @@
...
+    width: 120px;
+    height: 38px;
+    min-width: auto;
+    background-color: #629ff4;
+    border: 0;
+    color: var(--white);
+    text-transform: uppercase;
+    letter-spacing: 0.8px;
+    > .plusIcon {
```
rileyjshaw: Kebab-case is preferred: `plus-icon`

web/src/components/pages/dashboard/dashboard.css
```diff
@@ -174,3 +173,123 @@
...
+        width: 14px;
+        height: 14px;
+    }
+    span {
+        margin: 0 auto;
+    }
+    .score {
+        font-weight: 600;
+    }
+    > .label {
```
rileyjshaw: I'd like to keep our selectors as simple as possible. Can we remove the child combinator (`>`) from all of these styles, or is it needed for some reason?

web/src/components/pages/dashboard/dashboard.tsx
```diff
@@ -63,8 +66,20 @@ const TopBar = ({ dashboardLocale }: { dashboardLocale: string }) => {
   }, []);
 
   return (
-    <div className="top-bar">
+    <div className={`top-bar ${isChallengeTabVisible ? 'challenge' : ''}`}>
       <nav>
+        {/* only visible for users enrolled in challenge*/}
+        {isChallengeTabVisible ? (
```
rileyjshaw: Nitpick: if you're conditionally rendering a single component, I think the preferred style is `{isChallengeTabVisible && ( ...` rather than a ternary. See https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator.

web/src/components/pages/dashboard/dashboard.tsx
```diff
@@ -25,6 +27,7 @@ const TopBar = ({ dashboardLocale }: { dashboardLocale: string }) => {
   const [, toLocaleRoute] = useLocale();
   const account = useAccount();
   const [showTitleBarLocales, setShowTitleBarLocales] = useState(true);
+  const isChallengeTabVisible = isProduction() ? false : true;
```
rileyjshaw: Nitpick: same as `const isChallengeTabVisible = !isProduction();`

web/src/components/pages/dashboard/dashboard.tsx
```diff
@@ -148,6 +163,8 @@ const TopBar = ({ dashboardLocale }: { dashboardLocale: string }) => {
           </select>
         )}
       </div>
+      {/* only visible for users enrolled in challenge*/}
+      {isChallengeTabVisible ? <ChallengeBar /> : null}
```
rileyjshaw:
```
{isChallengeTabVisible && <ChallengeBar />}
```

web/src/components/pages/dashboard/dashboard.css
```diff
@@ -174,3 +173,123 @@
...
+        > .plusIcon {
+            margin: 0;
+        }
+    }
+    .languages.challenge {
+        display: none !important;
+    }
+}
+
+@media (--xs-down) {
```
rileyjshaw: Thank you for adding all of these responsive styles!! 😃

web/src/components/pages/dashboard/dashboard.tsx
```diff
@@ -97,7 +112,7 @@ const TopBar = ({ dashboardLocale }: { dashboardLocale: string }) => {
         </LocaleNavLink>
       </nav>
 
-      <div className="languages">
+      <div className={`languages ${isChallengeTabVisible ? 'challenge' : ''}`}>
```
rileyjshaw: The challenge pilot is going to be English-only. I think it makes sense to prevent showing the language bar for the Challenge tab, and instead put in some static text that says "Language: ENGLISH" like in the mock-ups.

web/src/components/pages/dashboard/dashboard.tsx
```diff
@@ -169,10 +186,28 @@ function DashboardContent({
...
+      <span className="plusIcon"></span>
+    </Button>
+  </div>
+);
+
 const PAGES = [
   { subPath: URLS.STATS, Page: StatsPage },
   { subPath: URLS.GOALS, Page: GoalsPage },
   { subPath: URLS.AWARDS, Page: AwardsPage },
+  { subPath: URLS.CHALLENGE, Page: AwardsPage },
```
rileyjshaw: Can we try rendering `StatsPage` here instead? `AwardsPage` has some visual bugs (it's full-width, unlike the other pages), so I think `StatsPage` will give a better idea of what the layout will actually look like.

Thanks!

web/src/components/pages/dashboard/dashboard.css
catherine917: Thanks, @rileyjshaw.
Your suggestion is helpful. But I think these selectors' name is too general. If import other styles, I concern about the same name selectors existed. The styles will be overwritten.

web/src/components/pages/dashboard/dashboard.css
```diff
@@ -217,7 +218,24 @@
         background-color: var(--light-grey);
     }
 }
-.inviteButton {
+.languages.challenge {
+    display: none !important;
+}
```
rileyjshaw: I'd prefer a conditional render in the `.tsx` file – that way we don't clutter the DOM with unused components. I'll make that change in a future commit, don't worry about it!

web/src/components/pages/dashboard/dashboard.css
rileyjshaw: I agree that `.label` would be too general, but since you’ve nested it, the selector will already resolve to `.points .label`. My comment was on the child combinator (`>`); I think it’s overkill to set the selector as `.points > .label`. Removing the `>` should still be specific enough, and makes the selector more resilient to changes in the component structure.

It’s just a small nitpick, certainly not a blocker. I'll land your changes now, thanks for the awesome work!

catherine917: You are right. I will pay attention to it in my future work.

PR #2251, opened 2019-10-16
Fixed typos and poorly translated expressions

vowk1n:
Also made the spacing consistent with the English version.


nukeador:
Hi @vowk1n 

Thanks for your PR. For localization content we don't accept PT directly, since we use the Mozilla's pontoon localization system to manage and merge localizations.

For Spanish you can make your changes here:

https://pontoon.mozilla.org/es/common-voice/

A community reviewer will check and approve them, and then they will be auto-incorporated into the website in our next deployment.

Thanks!

PS: Por favor, no dudes en hacerme ping si encuentras algún problema con pontoon, si tienes dudas sobre la localización al español puedes preguntar a la comunidad en los foros de Common Voice en español https://discourse.mozilla.org/c/voice/es ¡Gracias!

PR #2255, opened 2019-10-19
Fix horizontal overflow

### LINKED ISSUE #2239: Horizontal Scrolling on main page should not be possible
On the main page (voice.mozilla.org) it's currently possible to scroll horizontally without actually having more content.

Screenshot 2019-10-02 at 21 04 44

saranshbarua:
Added a quick CSS rule to hide overflow-x that fixes #2239. @Gregoor @rileyjshaw please take a look.

### DIFF
--- a/web/src/components/index.css
+++ b/web/src/components/index.css
@@ -41,9 +41,6 @@ body,
     width: 100%;
 }
 
-body {
-}
-
 h1,
 h2,
 h3,
--- a/web/src/components/register-section/register-section.css
+++ b/web/src/components/register-section/register-section.css
@@ -1,6 +1,8 @@
 @import url('../vars.css');
 
 .register-section {
+    overflow-x: hidden;
+
     & .top {
         margin: 0 auto;
         display: flex;


rileyjshaw:
Ah, thanks so much! I've been meaning to get around to this.

We should avoid hiding overflow on `body` if possible. I think the offending elements are within the `.signup-info` container, specifically `.waves` and `.bg`.

Rather than hiding overflow, I'd much rather fix the underlying issue. _Why_ are these elements wider than the page?

Thanks for pitching in! Ping me if you get a chance to update the PR.

rileyjshaw:
@saranshbarua are you able to take a look at this?

saranshbarua:
@rileyjshaw I looked at the problem and you are right. The problem is with .waves which has a width of 100vw. And it lies under the div with class .top which has a margin: (0 auto). which shifts the .waves to the right. Now, the waves can be shifted to the left but the value can't be determined because it is governed by the rule of margin: (0 auto). One better fix I found is to hide the overflow but this time for the class .register-section which fixes the problem. Applying this rule to any children element cuts the image of the waves which looks absurd. So this is my proposed solution. WDYT? Should I update the PR?

rileyjshaw:
Thank you for looking into this! That solution sounds good to me, let's get
this updated and merged into master 🎉

saranshbarua:
@rileyjshaw I just updated the PR, PTAL. Hope this helps :)

rileyjshaw:
Thanks so much for your contribution!

PR #2257, opened 2019-10-21
challenge signup backend task

can-007:
Implementation of challenge sign-up backend task:
1) pass challenge_team between` /login `and` /callback`;
2) in the `/callback`, update challenge_team of the user identified by the email who has already signed up and without a team; (this is for signing up and teaming up separately)
3) pass the challenge_team from `/callback` to frontend and get saved with the email in the Profile page. (this is for signing up and teaming up altogether)
4) pass the challenge_team from `/api/v1/user_client` and `/api/vi/user_clients` to frontend to display the challenge page;
5) include challenge_team in the downloaded data from the _settings_ page.
Gratitude to @catherine917 for the help on the frontend.
Gratitude to @rileyjshaw  for the technical clarification on the workflow.

How to test it (DON'T FORGET TO ADD THE TEAM IN `teams` TABLE):
1) check that the "`Challenge`" tab will show on "`Dashboard`" page after user joined a team;
teamup

2) check DB that the challenge_team has correct team name;
3) download the _my data_ from the settings page, and see challenge_team is in it.


can-007:
@rileyjshaw , sorry for the late pull request, please review this in case of major mistakes.

rileyjshaw:
> in the /callback, update challenge_team of the user identified by the email who has already signed up and without a team; (this is for signing up and teaming up separately)

Thanks for adding this! That's an important use case.

common/user-clients.ts
```diff
@@ -16,4 +16,5 @@ export type UserClient = {
   votes_count?: number;
   awards?: any[];
   custom_goals?: CustomGoal[];
+  challenge_team?: string;
```
rileyjshaw: This will probably look different after the schema changes, eg:

```
challenge_enrollments?: {
  ...
  team: string,
}[];
```

But this looks good for now!

server/src/lib/model/user-client.ts
```diff
@@ -243,6 +245,18 @@ const UserClient = {
     return true;
   },
 
+  // update the challenge_team of the user who is already signed up but have not joined any team
+  // [BUG]: there are exceptions if the team doesn't exist
```
rileyjshaw: We should add a constants file. I'll do that in a future diff, thanks for calling out the bug here. Please add:

```
// TODO(riley): Hook this up to a constants file.
```

server/src/lib/model/user-client.ts
```diff
@@ -243,6 +245,18 @@ const UserClient = {
...
+  async updateChallengeTeam(
+    email: string,
+    challenge_team: string
+  ): Promise<boolean> {
+    return await db.query(
+      'UPDATE user_clients SET challenge_team = ? WHERE email = ? AND has_login AND challenge_team IS NULL',
```
rileyjshaw: Should this be `… AND challenge_team IS NOT NULL`?

web/src/components/pages/dashboard/dashboard.tsx
```diff
@@ -34,7 +33,7 @@ const TopBar = ({
   const [, toLocaleRoute] = useLocale();
   const account = useAccount();
   const [isAboveMdWidth, setIsAboveMdWidth] = useState(true);
-  const isChallengeEnrolled = !isProduction();
+  const isChallengeEnrolled = account && account.challenge_team;
```
rileyjshaw: Love it!

server/src/auth-router.ts
```diff
@@ -114,6 +118,14 @@ router.get(
       }
       response.redirect('/profile/settings?success=' + success.toString());
     } else {
```
rileyjshaw: We should probably put the `challenge_team`-specific logic under an `else if (challenge_team) {`, right?

Also, at the bottom of the `else if` you're adding, please call:

```
response.redirect(redirect || basePath + 'login-success?ovchall=1');
```

That will let us show a different front-end flow for newly-enrolled users.

server/src/auth-router.ts
```diff
@@ -137,6 +149,7 @@ router.get('/login', (request: Request, response: Response) => {
             }
           : {}),
         redirect: query.redirect || null,
+        challenge_team: query.challenge_team,
```
rileyjshaw: `query.challenge_team || null`

server/src/auth-router.ts
```diff
@@ -114,6 +118,14 @@ router.get(
       }
       response.redirect('/profile/settings?success=' + success.toString());
     } else {
+      // pass the challenge_team to frontend
+      user.challenge_team = challenge_team;
```
rileyjshaw: I'm confused about what's happening here. I'm missing context, but `user` looks like a temporary object?

web/src/components/pages/profile/info/info.tsx
```diff
@@ -160,7 +160,9 @@ function ProfilePage({
       locales: locales.filter(l => l.locale),
       visible: JSON.parse(visible.toString()),
       client_id: user.userId,
+      challenge_team: user.userClients[0].challenge_team,
```
rileyjshaw: I think that we should save this onto `userFields` instead, so it's accessible like any other field. That probably involves threading it into the store at `/web/src/stores/user.ts`.

server/src/lib/model/user-client.ts
can-007: My understanding is that NULL challenge_team means the user has not enrolled yet, he/she may have signed up independently, not from the invitation link.

This also implies that users can't switch teams, enrollment happens only once.
I forgot to ask you about the implications, sorry.

server/src/auth-router.ts
can-007: This is INDEED quite confusing! But the `user `belongs to the `request `object. 
The authentication server (or SSO server, be it auth0 or mozzila SSO) embedded the user information (email, verified or not, etc.) inside the `request `object, and send it back to `/callback`.
With the redirection below, the user object is passed to the frontend. That's how the email shows up in the `Profile `page.

server/src/auth-router.ts
can-007: If backend redirect `ovchall=1`, frontend will pop up the invite modal.
I guess it should and will be three new consecutive on-boarding modals in the recent future.
Thus, normally, invitation aside, there will be no ovchall=1 in the redirection, and no on-boarding modals.

web/src/components/pages/profile/info/info.tsx
can-007: maybe, but values in userFields come from the DB, now, challenge_team is like email, have not been persisted into DB.

PR #2258, opened 2019-10-24
challenge related schema as well as adaptations

can-007:
add challenge related tables and remove challenge_team from user_clients table;
add invite url processing, use names like ovchall, team and invite instead;
group ovchall, team and invite into enroll object and do modifications between frontend and backend;
removed the team part from the download;

[NOTE]
This is not pretty.
Login needs at least 3 parameters to pass around, back and forth between backend and frontend.
There should be some optimizations later.


rileyjshaw:
This looks great, thanks @can-007. I'm going to make a few changes based on feedback from the marketing team, eg. switch all query parameters to something readable (`ovchall` => `challenge`).

rileyjshaw:
Hi @can-007, I made some comments and pushed a few changes to this PR. Please review when able.

server/src/lib/model/user-client.ts
```diff
@@ -245,17 +251,34 @@ const UserClient = {
...
-    return await db.query(
-      'UPDATE user_clients SET challenge_team = ? WHERE email = ? AND has_login AND challenge_team IS NULL',
-      [challenge_team, email]
-    );
+    if (email && team_name && challenge_id && team_url_token) {
+      let client_id = await Promise.all([UserClient.findClientId(email)]);
+      if (client_id) {
+        const res = await db.query(
+          `INSERT IGNORE INTO enroll (id, challenge_id, url_token, client_id, team_id)
+           VALUES (null, ?, ?, ?, (SELECT id FROM teams WHERE challenge_id=? AND name=? AND url_token=?))`,
```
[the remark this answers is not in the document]
can-007: yeah, good point

server/src/lib/model/db/migrations/20191024074951-challenge-related-schemas.ts
```diff
@@ -0,0 +1,73 @@
...
+              ADD FOREIGN KEY (challenge_id) REFERENCES challenges(id);
+              
+    /* Records each enrollment (a user signing up for a team) across all challenges. */
+    CREATE TABLE enroll (
+      id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+
+      client_id CHAR(36) NOT NULL,
+      team_id INT UNSIGNED NOT NULL,        /* enroll means user becomes a team member of one challenge */
+      challenge_id INT UNSIGNED NOT NULL,	 
+      url_token CHAR(12) NOT NULL,          /* no foreign key constraints because team's url might change */
```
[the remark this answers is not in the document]
can-007: It seems that I misunderstood this. I thought url_token of enrollments comes from the url_token of teams.
Another invited_by is good, tracking who invites who. And let DB generate it on every insertion of enrollments.
But what about the first enrollment url_token within every team? I am thinking it could be anything, and check if the user is the first enrollments of the team. First enrollment of a team with non-matching url_token gets accepted, other enrollments of a team with non-matching url_token is rejected. Should enrollment rejection get some sort of notice, like a pop up modal?

server/src/lib/model/db/migrations/20191024074951-challenge-related-schemas.ts
```diff
@@ -0,0 +1,77 @@
...
+    );
+
+    ALTER TABLE teams DROP PRIMARY KEY,
+              ADD COLUMN id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY, /* better performance and shorter sql statements */
+              CHANGE invite_token url_token VARCHAR(36) NOT NULL,
+              CHANGE team_name name VARCHAR(255),
+              ADD COLUMN challenge_id INT UNSIGNED NOT NULL,                  /* teams has to be associated with exact one challenge to make sense */
+              ADD FOREIGN KEY (challenge_id) REFERENCES challenges(id);
+
+    /* Records each enrollment (a user signing up for a team) across all challenges. */
```
can-007: Use enrollments instead of enroll is fine.
I was thinking about the table naming convention: a verb means a relation, and an noun mean a entity, But it doesn't show much benefits.

server/src/lib/model/user-client.ts
```diff
@@ -245,17 +252,56 @@ const UserClient = {
...
   ): Promise<boolean> {
-    return await db.query(
-      'UPDATE user_clients SET challenge_team = ? WHERE email = ? AND has_login AND challenge_team IS NULL',
-      [challenge_team, email]
-    );
+    if (email && challenge && team) {
+      let client_id = await Promise.all([UserClient.findClientId(email)]);
+      if (client_id) {
+        // Generate a random, anonymized, and unique invite ID for enrollments.
+        // HACK(riley/can): Can we move this entirely into the DB?
```
can-007: mysql has a function named uuid(), just what we need. Haven't successfully run this new code yet.

rileyjshaw: Oh wow 😂 amazing!!! `UUID()` is a bit long for an invite code, but maybe we can use `UUID_SHORT()` and convert it from decimal => base 36 on the server?

Practically, `UUID_SHORT()` is close to just using `AUTO INCREMENT`, but I feel fine about that. Thanks for calling my attention to all this!!

server/src/lib/model/db/migrations/20191024074951-challenge-related-schemas.ts
rileyjshaw: Ah, I understand – sorry for missing that.

server/src/lib/model/db/migrations/20191024074951-challenge-related-schemas.ts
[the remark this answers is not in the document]
rileyjshaw: That's fine! `invited_by` will probably be `NULL` for most participants. It's just an optional field to track if they signed up with the url `?challenge=pilot&team=<team_url_token>&invite=<someone_elses_enrollment_url_token>`.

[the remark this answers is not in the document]
rileyjshaw: `url_token` is not optional, though. We need to fill it for each enrollment. I like your idea from below of using `UUID()`!

server/src/lib/model/user-client.ts
```diff
@@ -262,43 +261,25 @@ const UserClient = {
...
-            inviteToken
-          );
-          if (!res || !res.length) break;
-        }
-
+        // if invite is null, the invitation comes from team; OTHERWISE from user
+        // for user invitation, invite_by is the same value of url_token from another row
+        // for team invitation, invite_by is the same value of url_token of the same row
+        const [[{ team_id, challenge_id, uuid }]] = await db.query(
+          `SELECT t.id AS team_id, t.challenge_id, UUID() AS uuid FROM teams t WHERE t.url_token=?`,
```
rileyjshaw: I still think `UUID()` output is pretty long for a public invite ID. Are you sure we can't use `UUID_SHORT()`?

An easy win is `REPLACE(UUID(),'-','')`. If we have time, we can save another 10 characters by base 58 encoding the strings on the server before sending them down, eg:

```
const base58Alphabet = '123456789abcdefghijkmnopqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ';

function encodeInviteId(uuid, chars) {
	const base = chars.length;
	let decimalUuid = parseInt(uuid, 16);
	let inviteId = '';
	while (decimalUuid) {
		inviteId += chars[decimalUuid % base];
		decimalUuid = Math.floor(decimalUuid / base);
	}
	return inviteId.padStart(22, chars[0]);
}

function decodeInviteId(inviteId, chars) {
	// ...
}

// Example usage: encodeInviteId(uuid, base58Alphabet);
```

Let's not prioritize this. It's a nice to have but not on the critical path.

server/src/lib/model/db/migrations/20191024074951-challenge-related-schemas.ts
```diff
@@ -10,33 +10,36 @@ export const up = async function(db: any): Promise<any> {
 
       name VARCHAR(255),
       url_token VARCHAR(36) NOT NULL,
-      start_date DATETIME NOT NULL
+      start_date DATETIME NOT NULL,
+
+      UNIQUE (url_token)
```
rileyjshaw: Nice

server/src/lib/model/db/migrations/20191024074951-challenge-related-schemas.ts
```diff
@@ -10,33 +10,36 @@ export const up = async function(db: any): Promise<any> {
...
     );
 
     ALTER TABLE teams DROP PRIMARY KEY,
               ADD COLUMN id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY, /* better performance and shorter sql statements */
-              CHANGE invite_token url_token VARCHAR(36) NOT NULL,
+              CHANGE invite_token url_token VARCHAR(36) UNIQUE NOT NULL,
```
rileyjshaw: This won't necessarily be unique across competitions. For instance, I think the following should be two different, but valid, teams:

- `?challenge=pilot&team=sap`
- `?challenge=other&team=sap`

Maybe we can add `UNIQUE(challenge_id, url_token)` instead?

server/src/lib/model/user-client.ts
```diff
@@ -262,43 +261,25 @@ const UserClient = {
...
-        let inviteToken;
-        while (true) {
-          inviteToken = (await hash(`${Math.random()}`)).slice(0, 12);
-          const res = await db.query(
-            `SELECT * FROM enrollments WHERE url_token=?`,
-            inviteToken
-          );
-          if (!res || !res.length) break;
-        }
-
```
rileyjshaw: Yay! Thanks for figuring out how to get rid of this 🎉

server/src/lib/model/user-client.ts
```diff
@@ -262,43 +261,25 @@ const UserClient = {
...
+        // if invite is null, the invitation comes from team; OTHERWISE from user
+        // for user invitation, invite_by is the same value of url_token from another row
+        // for team invitation, invite_by is the same value of url_token of the same row
```
rileyjshaw: If invite is null, we can keep `invited_by` as `NULL` in the DB.

server/src/lib/model/user-client.ts
```diff
@@ -262,43 +261,25 @@ const UserClient = {
...
-            null,
-            @challengeId,
-            (SELECT id FROM teams WHERE challenge_id=@challengeId AND url_token=?),
-            ?,
-            ?,
-            ?
-          )`,
-          [challenge, team, client_id, invite, inviteToken]
+          INSERT INTO enroll (challenge_id, team_id, client_id, invited_by, url_token)
+          VALUES (${challenge_id}, ${team_id}, '${client_id}', ?, '${uuid}')
```
rileyjshaw: Why are some of these surrounded by `''` and some aren't?

rileyjshaw: Much simpler, nice!

server/src/lib/model/user-client.ts
```diff
@@ -262,43 +261,25 @@ const UserClient = {
...
+          `,
+          [invite || uuid]
```
rileyjshaw: I think we want `invite || null` here.

server/src/lib/model/db/migrations/20191024074951-challenge-related-schemas.ts
can-007: good point!

server/src/lib/model/user-client.ts
can-007: I put most of the values inside the sql string, only invite_by as prepared parameters.
string parameters are surrounded by ' ', numbers are not.
I see it already creates confusion. I will make it unified.

server/src/lib/model/user-client.ts
can-007: MySQL have conv() that can do the base encoding stuff, but mysql can only base 32 encode the string. 

Beside, parseInt() in js won't support numbers as big as UUID which is 128 bits long, 53 bits at most, then the precision is lost. 

Remove '-' from UUID, and base 64 encoding the string (plus 0,- and ~) will save 1/3 of the total length. I think the performance won't change much for that. It is a great idea, but I think it is better to leave it for future. I can leave a comments like `[FUTURE]`

can-007: As for `UUID_SHORT()`, run `select uuid_short()` any times in a mysql shell, different values will be yield. However, enroll two users, they will get the same short uuid, thus creating a _duplicate key exception_, and failing the second enrollment.

That is what I observed in my local dev environment (win-10, mysql-5.6.40 and mysql 8.0.17), not sure if it is the same on your environment.

can-007:
> Awesome! Requesting a few small changes, but this looks great. You mentioned wanting to test this further, so let me know when it's ready to go.
> 
> Thanks 🎉

I made the changes, and build the mysql from source tree the same as the staging server. It worked without syntax error. I think it will also work on the staging server.

server/src/lib/model/user-client.ts
```diff
@@ -265,19 +265,23 @@ const UserClient = {
...
+        // for team invitation, invite_by is null
+        // [FUTURE] UUID is too long, maybe consider to optimize it by removing '-' and base64 encoding it.
+        const [[{ team_id, challenge_id, enrollment_token }]] = await db.query(
+          `SELECT t.id AS team_id, t.challenge_id, UUID() AS enrollment_token FROM teams t
+            LEFT JOIN challenges c ON t.challenge_id = c.id
+            WHERE t.url_token=? AND c.url_token = ?`,
+          [team, challenge]
         );
+
+        // INSERT IGNORE will hide exceptions such as duplicate foreign key violation
```
rileyjshaw: There isn't an `INSERT IGNORE` in the statement below - should there be?

server/src/lib/model/db/migrations/20191024074951-challenge-related-schemas.ts
```diff
@@ -17,10 +17,11 @@ export const up = async function(db: any): Promise<any> {
 
     ALTER TABLE teams DROP PRIMARY KEY,
               ADD COLUMN id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY, /* better performance and shorter sql statements */
-              CHANGE invite_token url_token VARCHAR(36) UNIQUE NOT NULL,
+              CHANGE invite_token url_token VARCHAR(36) NOT NULL,
               CHANGE team_name name VARCHAR(255),
               ADD COLUMN challenge_id INT UNSIGNED NOT NULL,                  /* teams has to be associated with exact one challenge to make sense */
-              ADD FOREIGN KEY (challenge_id) REFERENCES challenges(id);
+              ADD FOREIGN KEY (challenge_id) REFERENCES challenges(id),
+              ADD UNIQUE INDEX unique_team (url_token, challenge_id);
```
rileyjshaw: Nice.

server/src/lib/model/user-client.ts
rileyjshaw: Thanks for the clarification, and for adding the `[FUTURE]` comment! I'm happy with this solution for now.

server/src/lib/model/user-client.ts
rileyjshaw: Ah, makes sense. That's why I shouldn't do reviews on Friday evening. Thanks 😅

server/src/lib/model/user-client.ts
```diff
@@ -245,17 +251,42 @@ const UserClient = {
...
-      'UPDATE user_clients SET challenge_team = ? WHERE email = ? AND has_login AND challenge_team IS NULL',
-      [challenge_team, email]
-    );
+    if (email && challenge && team) {
+      const client_id = (await Promise.all([
+        UserClient.findClientId(email),
+      ]))[0];
+      if (client_id) {
+        // for user invitation, invite_by is the same value of url_token from another row
+        // for team invitation, invite_by is null
```
rileyjshaw: The language of "team invitation" confused me here. The URL will either look like:

`?challenge=pilot&team=sap&invite=<uuid>` or,

`?challenge=pilot&team=sap`

So I think it's either a user invitation or no user invitation. Adding the "team invitation" comment sounds to me like teams have their own invite codes. I think you can just delete that second line.

---

Also, nitpick but there's a typo in the comment – it should be `invited_by`.

rileyjshaw:
Awesome! I just went through my old comments and made sure everything's been resolved. There are two outstanding (tiny) changes requested, but I don't want to hold you up. I'll merge this in and fix the comments upstream.

So happy we're landing this! Great work.

server/src/lib/model/user-client.ts
rileyjshaw: (Fixed upstream)

PR #2260, opened 2019-10-29
Upgrade notification pill for Open Voice Challenge

catherine917:
@rileyjshaw Could you give me more requirements about the notification pills up-gradation? Now I add a type named `achievements` and set the background as blue. I don't know whether or not it is enough for the up-gradation.

### DIFF
--- a/web/src/components/notification-pill/notification-pill.css
+++ b/web/src/components/notification-pill/notification-pill.css
@@ -12,7 +12,6 @@
     opacity: 1;
     transition: opacity linear var(--transition-duration);
     animation: fade-in var(--transition-duration-slow);
-
     &.success {
         background: var(--valid-green);
     }
@@ -23,6 +22,14 @@
 
     &.achievements {
         background: var(--blue);
+        line-height: 1.5;
+        border-radius: 29px 0 0 29px;
+        padding: 10px 20px;
+        img {
+            width: 14px;
+            height: 14px;
+            margin-right: 15px;
+        }
     }
 
     &:first-child {
--- a/web/src/components/notification-pill/notification-pill.css
+++ b/web/src/components/notification-pill/notification-pill.css
@@ -20,7 +20,7 @@
         background: var(--red);
     }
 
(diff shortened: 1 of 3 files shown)


web/src/components/notification-pill/notification-pill.tsx
```diff
@@ -28,6 +28,9 @@ export default function NotificationPill({
       }
       style={{ opacity: show ? 1 : 0 }}
       onTransitionEnd={() => removeNotification(notification.id)}>
+      {notification.kind === 'pill' && notification.type === 'achievement' && (
+        <img src={require('../pages/dashboard/awards/star.svg')} alt="star" />
```
rileyjshaw: Since the star icon doesn't have semantic meaning, we can skip the `alt` tag. From https://www.w3.org/WAI/tutorials/images/decorative: 

```
<img src={require('../pages/dashboard/awards/star.svg')} alt="" role="presentation" />
```

web/src/components/notification-pill/notification-pill.css
```diff
@@ -21,6 +20,18 @@
         background: var(--red);
     }
 
+    &.achievement {
+        background: var(--blue);
+        line-height: 1.5;
+        border-radius: 29px 0 0 29px;
```
rileyjshaw: Since the "half-pill" slides in from the right, @mbransn probably had a different animation in mind than `fade-in`.

Maybe we should have a `.achievement` class to make the toast bigger and support an `img`, and an additional `.dashboard` class to make it a half-pill with a different animation? What do you think?

web/src/components/notification-pill/notification-pill.tsx
catherine917: Sure

web/src/components/notification-pill/notification-pill.css
catherine917: Yeah, I will adjust the styles associated with the challenge dashboard.

PR #2261, opened 2019-10-29
Challenge board backend

can-007:
1) fix a bug: saveAccount() will crash if the body has no enrollment, e.g, change visibility from leaderboard;
2) change visible column to support one more visibility option;
3) add backend APIs in a separate ts file with a separate route, and provide mock data (because the calculation will take more time thus will block the frontend);

@rileyjshaw , the challenge board frontend has been waiting for workable backend APIs, and I need more time on the exact SQLs underlying the APIs. So I provide the API interfaces first with some mock data. But I don't know if it is appropriate.


server/src/lib/model/user-client.ts
```diff
@@ -191,12 +191,14 @@ const UserClient = {
...
-        data.enrollment.invite
-      ),
+      data &&
+        data.enrollment &&
+        this.enrollRegisteredUser(
+          email,
+          data.enrollment.challenge,
+          data.enrollment.team,
+          data.enrollment.invite
+        ),
```
rileyjshaw: This seems like a more pressing fix, so I'm going to pull it into its own commit.

rileyjshaw:
https://github.com/mozilla/voice-web/commit/2f59870385a4fe460ed2cebeb08c71f76d32b8dd

server/src/lib/model/user-client.ts
rileyjshaw: https://github.com/mozilla/voice-web/commit/2f59870385a4fe460ed2cebeb08c71f76d32b8dd

rileyjshaw:
> @rileyjshaw , the challenge board frontend has been waiting for workable backend APIs, and I need more time on the exact SQLs underlying the APIs. So I provide the API interfaces first with some mock data. But I don't know if it is appropriate.

Thank you, that's a great approach! I've got some meetings coming up but will look into this ASAP.

challenge-mock-data.json
```diff
@@ -0,0 +1,421 @@
...
+          "accuracy": 9.18
+        }
+      ]
+    }
+  },
+  "topTeams": [
+    {
+      "position": 1,
+      "name": "SAP1",
+      "logo": "base641...",
```
rileyjshaw: Not a blocker obviously, but I want to make sure we're on the same page. I was thinking these would eventually be external URLs from the `logo_url` column of our database (or alternatively, some constants file on the front-end), not base64 encoded. Is that right?

server/src/lib/model/db/migrations/20191029113543-change-visible-use-clients.ts
```diff
@@ -0,0 +1,11 @@
+export const up = async function(db: any): Promise<any> {
+  return db.runSql(
+    `
+    ALTER TABLE user_clients MODIFY visible TINYINT(2) DEFAULT 0;
```
rileyjshaw: This seems okay to me, since `visible` already expects a type of `0 | 1`. Note that you'll need to add `| 2` to `/common/user-clients.ts`.

Can you write a bit more about what each value will represent?

server/src/lib/challenge.ts
```diff
@@ -0,0 +1,104 @@
...
+    router.get(
+      '/:challenge/:locale/contributors/:type',
+      this.getTopContributors
+    );
+
+    return router;
+  }
+
+  getPoints = async (
+    { client_id, params: { challeng_token } }: Request,
```
rileyjshaw: Can we update the spelling here `challeng_token` => `challenge_token`? Where is "challeng" coming from?

challenge-mock-data.json
```diff
@@ -0,0 +1,421 @@
...
+        "position": 10,
+        "name": "catherine00",
+        "logo": "base6400...",
+        "points": 12300,
+        "approved": 1031,
+        "accuracy": 12
+      }
+    ]
+  }
+}
```
rileyjshaw: lgtm

server/src/lib/model/db/migrations/20191029113543-change-visible-use-clients.ts
rileyjshaw: Ah, nevermind. I just saw it's updated in https://github.com/mozilla/voice-web/pull/2256/.

challenge-mock-data.json
can-007: Yes, that is also my thought.

server/src/lib/challenge.ts
can-007: A typo, my fault, should be challenge_token. Just not sure if
front end have that value.

server/src/lib/model/db/migrations/20191029113543-change-visible-use-clients.ts
can-007: A little comments would be good though, I will add that.

rileyjshaw:
> A typo, my fault, should be challenge_token. Just not sure if front end have that value.

Got it. I'm going to land this so you're not blocked, please fix it upstream!

PR #2263, opened 2019-10-31
Clarify contribution policy for localization content

phirework:
Clarifies policies on PRs to address confusion raised in #2251


rileyjshaw:
Great idea, thanks!

PR #2266, opened 2019-11-01
Add pilot challenge team info to the constants file

rileyjshaw:
As we start hooking the front- and back-end together, it's important to
have stricter checks in place that we're sending data in the correct
format. This file should help with that.

Eventually, all the team string literals we use in our challenge logic
should be replaced with values from this file.

Note: the Mozilla and IBM svgs that I added are placeholders until we
get updated assets from design.

cc. @phirework, @catherine917, @can-007


web/src/components/pages/dashboard/challenge/weekly-challenge.tsx
```diff
@@ -86,7 +87,7 @@ export default function WeeklyChallenge({ isNarrow }: { isNarrow?: boolean }) {
               week={currentWeek}
               isNarrow={isNarrow}
               avatarUrl={account.avatar_url}
-              challengeTeam={'sap'} //account.challenge_team}
+              logoUrl={challengeTeams['sap'].logoUrl}
```
rileyjshaw: Eventually these will look something like `challengeTeams[account.enroll.team].logoUrl`.

rileyjshaw:
Looks like Jest is getting caught on:

```
SyntaxError: Unexpected token <

[1] 

[1]       46 |     readableName: 'IBM',

[1]       47 |     token: 'ibm',

[1]     > 48 |     logoUrl: require('../web/src/components/pages/dashboard/challenge/images/ibm.svg'),

[1]          |              ^
```

Looking into it..

rileyjshaw:
I updated this to fix the failing test, and added the following to the commit message:

> We decided to add a front-end specific constants file, since our webpack config is inconsistent between /common and /web so broke on the `require` calls used for logo URLs.

web/src/components/pages/dashboard/challenge/weekly-challenge.tsx
```diff
@@ -131,7 +132,7 @@ export default function WeeklyChallenge({ isNarrow }: { isNarrow?: boolean }) {
                       key={index}
                       isNarrow={isNarrow}
                       avatarUrl={account.avatar_url}
-                      challengeTeam={'sap'} //account.challenge_team}
+                      logoUrl={challengeLogoUrls['sap']}
```
phirework: Are all these logos right now pointing to SAP because we're waiting on the other two companies? Or is this just a placeholder?

rileyjshaw: Yep!

> Note: the Mozilla and IBM svgs that I added are placeholders until we get updated assets from design.

rileyjshaw:
Awesome, thanks for reviewing, @phirework

PR #2268, opened 2019-11-04
challenge back-end: replace mock data with real data from db

can-007:
Description:
1) Reuse leaderboard.ts for its cursor and position functionality, as well as caching utility;
2) Return challenge token when front-end requests for the current account/user to be used in challenge related API calls;
_As a consequence, it will try to enroll the current account when save the account's data._
3) To avoid crash, check the client_id, enroll_id and challenge_id before enrolling registered users;
4) Change the interfaces between front-end and back-end for simplicity;

TODO:

- [x]  For now, the _Top Team_ leaderboard only shows the team rankings for current week, and it will appear at different column such as _w1, w2 OR w3_.
The design is to show rankings at all 3 columns in the 3rd week.
- [x]  Check and move the constants in to common/challenge.ts.
- [x]  Close this PR, fix the conflicts, squash all the commits into one and submit a new PR.

Notes:
I admit that this PR is a mess.
There are some refactoring to be done, however, this PR could be used for the review of the main functionalities.


can-007:
@rileyjshaw , please check the PR, especially the SQLs when you have time.

server/src/lib/challenge.ts
```diff
@@ -31,74 +28,82 @@ export default class Challenge {
...
+      challenge
     );
-    let data = getMockData();
-
-    response.json(data.weeklyChallenge);
+    const weeklyProgress = {
+      week: progress.week || null,
+      user: {
+        speak: progress.clip_count || null,
+        speak_total: progress.week == 2 ? 100 : 200,
```
rileyjshaw: nitpick: I'd prefer to see `===` here.

server/src/lib/challenge.ts
```diff
@@ -31,74 +28,82 @@ export default class Challenge {
...
-      `[DEBUG] Challenge.getWeeklyProgress - client_id:${client_id}, challenge_token:${challeng_token}`
+    const progress = await this.model.db.getWeeklyProgress(
+      client_id,
+      challenge
     );
-    let data = getMockData();
-
-    response.json(data.weeklyChallenge);
+    const weeklyProgress = {
+      week: progress.week || null,
```
rileyjshaw: Do we need these `|| null`s here? In what case will these values be falsey?

server/src/lib/challenge.ts
```diff
@@ -31,74 +28,82 @@ export default class Challenge {
...
-    );
-    let data = getMockData();
-
-    response.json(data.challengePoint);
+    response.json(await this.model.db.getUserPoints(client_id, challenge));
   };
 
   getWeeklyProgress = async (
-    { client_id, params: { challeng_token } }: Request,
+    { client_id, params: { challenge } }: Request,
```
rileyjshaw: Thanks for fixing the typo here!

server/src/lib/challenge.ts
```diff
@@ -31,74 +28,82 @@ export default class Challenge {
...
   };
 
   getTopMembers = async (
-    { client_id, params: { challeng_token, locale, type } }: Request,
+    { client_id, params: { challenge, locale, type }, query }: Request,
     response: Response
   ) => {
-    console.log(
-      `[DEBUG] Challenge.getTopMembers - client_id:${client_id}, locale: ${locale}, challenge_token:${challeng_token}, type:${type}`
+    const atype = type == 'vote' ? 'vote' : 'clip';
```
rileyjshaw: `===`

rileyjshaw: I'm not familiar with the potential values of `type` – will it ever be something other than `"clip"` or `"vote"`?

server/src/lib/model/db.ts
```diff
@@ -586,4 +586,71 @@ export default class DB {
       );
     }
   }
+
+  async getUserPoints(client_id: string, challenge: string) {
+    const [[row]] = await this.mysql.query(
+      `
+      SELECT (bonus_points + clip_point + vote_point) AS user
```
rileyjshaw: nitpick: can we pluralize all of these? ie. `bonus_points + clip_points + vote_points`

server/src/lib/model/leaderboard.ts
```diff
@@ -121,16 +278,75 @@ export const getFullVoteLeaderboard = lazyCache(
...
+      team_only,
+    })).map((row, i) => ({
+      position: i,
+      ...row,
+    }));
+  },
+  CACHE_TIME_MS
+);
+
+export const getTopTeamLeaderboard = lazyCache(
```
rileyjshaw: Thanks for caching all of these!

server/src/lib/model/db.ts
```diff
@@ -586,4 +586,71 @@ export default class DB {
...
+      FROM (
+          SELECT bonus_points, clip_point, COUNT(votes.id) AS vote_point
+          FROM (
+              SELECT challenger.client_id, start_date, end_date, bonus_points, COUNT(clips.id) AS clip_point
+              FROM (
+                  SELECT user_clients.client_id,
+                      challenges.start_date AS start_date,
+                      TIMESTAMPADD(WEEK, 3, challenges.start_date) AS end_date,
```
rileyjshaw: Clever!

server/src/lib/model/db.ts
```diff
@@ -586,4 +586,71 @@ export default class DB {
...
+              GROUP BY challenger.client_id, start_date, end_date, bonus_points
+          ) speaker
+          LEFT JOIN votes ON speaker.client_id = votes.client_id AND votes.created_at BETWEEN start_date AND end_date
+          GROUP BY speaker.client_id, speaker.bonus_points, speaker.clip_point
+      ) voter
+      `,
+      [client_id, challenge]
+    );
+    return row;
+  }
```
rileyjshaw: Wow, this is a huge query. Nice work! I couldn't find any issues with it, but we should definitely test these functions out before the challenge start date.

We should also start to think about performance, and if we want to index anything here...

Lgtm!

server/src/lib/model/db.ts
```diff
@@ -586,4 +586,71 @@ export default class DB {
...
+      `
+      SELECT speaker.client_id, start_date, end_date, week, colleague_count, clip_count, COUNT(votes.id) AS vote_count
+      FROM (
+          SELECT user.client_id, start_date, end_date, week, colleague_count, COUNT(clips.id) AS clip_count
+          FROM (
+              SELECT user_clients.client_id,
+                  TIMESTAMPDIFF(WEEK, start_date, NOW()) + 1 AS week,
+                  TIMESTAMPADD(WEEK, TIMESTAMPDIFF(WEEK, start_date, NOW()), start_date) AS start_date,
+                  TIMESTAMPADD(WEEK, TIMESTAMPDIFF(WEEK, start_date, NOW()) + 1, start_date) AS end_date,
+                  COUNT(colleagues.id) AS colleague_count
```
rileyjshaw: Love it! If any of this ends up being too intensive for our DB at scale we can simplify and pass in a static `current_week` value, but this looks great for now.

server/src/lib/model/leaderboard.ts
```diff
@@ -15,6 +15,47 @@ const bucket = new Bucket(model, s3);
...
+  SELECT team_id FROM enroll e LEFT JOIN challenges c ON e.challenge_id = c.id WHERE e.client_id = ? AND c.url_token = ?
+)`;
+const general_condition = `
+(
+  (visible = 0 AND user_clients.client_id = ?)
+  OR (visible = 1)
+  OR (visible = 2 AND teams.id = (
+      SELECT team_id FROM enroll e LEFT JOIN challenges c ON e.challenge_id = c.id WHERE e.client_id = ? AND c.url_token = ?
+  ))
+)`;
```
rileyjshaw: Not a blocker by any means, but it might be nice to combine some of this:
```
const self_subcondition = '(visible = 0 AND user_clients.client_id = ?)';
const teammate_subcondition =
	'(SELECT team_id FROM enroll e LEFT JOIN challenges c ON e.challenge_id = c.id WHERE e.client_id = ? AND c.url_token = ?)'
const team_only_condition =
	`(visible OR ${self_subcondition}) AND enroll.team_id = ${teammate_subcondition}`;
const general_condition = `(
  ${self_subcondition}
  OR (visible = 1)
  OR (visible = 2 AND teams.id = ${teammate_subcondition})
)`;
```

server/src/lib/model/leaderboard.ts
```diff
@@ -97,6 +138,122 @@ async function getVoteLeaderboard(locale?: string): Promise<any[]> {
...
 
+// NOTE: The 4 top-related SQLs
+// 1) think the huge SQL as a process of adding aggregated attributes to user_clients rows:
+// first add bonus, then add clips as well as the votes for the clips, at last do calculation.
+// 2) only calculate values within the range of the challenge time span for bonus and clips, (maybe not votes).
+// 3) be careful in deciding where to put the filtering conditions: in LEFT JOIN or in WHERE, big differences.
+// 4) surely it can be optimized in that:
+//        participant in sql are ALMOST identical;
+//        topContributor and topMember sqls are different only in the WHERE clause.
+// 5) some of the conditions are probably over-thinking.
```
rileyjshaw: Thanks for thinking about optimization. I'm happy to start testing unoptomized and see where we need attention.

server/src/lib/challenge.ts
can-007: good point, there shouldn't be these `|| null` here

server/src/lib/challenge.ts
can-007: it is the type checking of typescript, I have to pass in literal clip or vote, I plan to make the parameters shorter and squeeze the following function call in a single line, but VS Code always re-format it in several lines. Maybe I should put the ternary expression in the function call.

server/src/lib/model/db.ts
can-007: Yeah, for consistency, surely.

server/src/lib/model/leaderboard.ts
can-007: Great, it is cleaner.

can-007:
close this PR as new PR https://github.com/mozilla/voice-web/pull/2274 that is cleaner and without conflicts are created.

PR #2273, opened 2019-11-06
Block Kibana access from anyone outside of ADMIN_EMAILS

rileyjshaw:
Our AWS-hosted Kibana instance cannot support multiple user role
settings, eg. Admin and Contributor. We would eventually like to run our
own self-hosted instance, but for now this plugs the security hole
introduced by allowing anyone full admin / write access to Kibana.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -78,7 +78,7 @@ export default class Server {
 
     app.use(authRouter);
     app.use('/_plugin/kibana', authMiddleware, (request, response, next) => {
-      const target = getConfig().KIBANA_URL;
+      const { KIBANA_URL: target, ADMIN_EMAILS } = getConfig();
       if (!target) {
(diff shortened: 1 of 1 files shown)


phirework:
Code looks good to me, do we need to wait for the pending travis CI to finish running / can we trigger it manually?

rileyjshaw:
> Shouldn’t the JSON.parse () and getConfig() only happen once at server startup instead of on every single request ?

If you look elsewhere in this file, `getConfig()` is already called on every request. The function response is memoized, so it's a pretty negligible call.
 
It's cached on startup elsewhere in the app (eg. `auth_router.ts`) so it's possible, but I think refactoring the file is out of scope for this commit.

To prevent any added calls I amended this to pull `ADMIN_EMAILS` from an existing call.

PR #2275, opened 2019-11-06
Add unique row key to ChallengeList render to get rid of warning

phirework:
Was getting a warning on duplicate list keys on challenge-list.tsx, presumably because the mapped items/row positions are not unique across all FetchRows. This is just a super quick fix - I'm not familiar with this part of the codebase, lemme know if there's another way you'd prefer to handle it @catherine917


rileyjshaw:
Thanks for noticing this! After discussing in person, we're gonna fix it in another way once some changes to `challenge-list.tsx` land.

PR #2278, opened 2019-11-07
Add HTTP referer string to challenge enrollments

rileyjshaw:
When we analyze our signup data, we might want to know where users came
from when they initially signed up. Tracking the HTTP `referer` header
on `enroll` entries is a lightweight way to do so.

### DIFF
--- a/server/src/auth-router.ts
+++ b/server/src/auth-router.ts
@@ -97,11 +97,16 @@ if (DOMAIN) {
 router.get(
   CALLBACK_URL,
   passport.authenticate('auth0', { failureRedirect: '/login' }),
-  async ({ user, query: { state }, session }: Request, response: Response) => {
+  async (request: Request, response: Response) => {
+    const {
+      user,
+      query: { state },
+      session,
+    } = request;
     const { locale, old_user, old_email, redirect, enrollment } = JSON.parse(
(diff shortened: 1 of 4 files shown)


rileyjshaw:
Tagging @can-007 for visibility.

can-007:
This is exactly what I thought how this could be done.
My other thoughts would be:
1) should we keep the entire headers, just in case something might missing? 
2) should we extract only the host.domain part of the referee header?
3) we are not recording other information such as IP, browser type & version, in app’s webview or just browser, right?

rileyjshaw:
> This is exactly what I thought how this could be done.
> My other thoughts would be:
> 
>     1. should we keep the entire headers, just in case something might missing?
> 
Good question! I think we only care about the `referer` for this task. We're connected to Google Analytics for more in-depth analysis, so this is only to check where people come from initially.
>     2. should we extract only the host.domain part of the referee header?
> 
That's a good point. I think it's worth keeping the whole string, since we don't know exactly how we'll use it yet. We can process strings later if we want, and the whole string won't take too much space.
>     3. we are not recording other information such as IP, browser type & version, in app’s webview or just browser, right?
Correct! We can use Google Analytics for that, and I want to reduce complexity for this specific task.

PR #2287, opened 2019-11-09
Update locales

MichaelKohler:
@nukeador r?


locales/native-names.json
```diff
@@ -8,6 +8,7 @@
   "as": "অসমীয়া",
   "ast": "Asturianu",
   "az": "Azərbaycanca",
+  "be": "be",
```
MichaelKohler: looks like there is no be translation yet, and therefore there is no translated native name

locales/native-names.json
```diff
@@ -83,6 +87,7 @@
   "sq": "Shqip",
   "sr": "Српски",
   "sv-SE": "Svenska",
+  "sw": "sw",
```
MichaelKohler: looks like there is no be translation yet, and therefore there is no translated native name

nukeador:
Thanks!

PR #2295, opened 2019-11-12
Jz/challenge/add challenge date query params

phirework:
Adds URL query parameter processing so that testers can manually specify `challenge_start` and `challenge_end` dates for the sake of testing empty states. 

* Default dates are Nov 18 to Dec 18 midnight UTC time
* If you specify a start date but no end date, the challenge should run from the custom start date to Dec 18
* If you specify an end date but no start date, the challenge should run from Nov 18 to the custom end date
* If an invalid date is provided (e.g. `challenge_start=test`), it will be disregarded and treated as though the parameter did not exist
* If either of the values supplied create a scenario where the end is not after the start, the default dates will be used

This code does no auth checking for, e.g. making sure the person logged in is a Mozillian or anything like that, and should be removed before launch.

### DIFF
--- a/docker/local-docker-config.json
+++ b/docker/local-docker-config.json
@@ -6,5 +6,11 @@
     "accessKeyId": "local-identity",
     "secretAccessKey": "local-credential",
     "s3ForcePathStyle": true
-  }
+  },
+  "AUTH0": {
+    "DOMAIN": "phire.auth0.com",
+    "CLIENT_ID": "qbAdutLVRV0rgvghVkYDrr55CRWR4saD",
+    "CLIENT_SECRET": "dCs7ESt1NZuCvWH6zkGfaQICEmKxjcRFs1fwUMDxj1NP9dLRAQSIux8iaGGzrs1O"
+  },
+  "IMPORT_SENTENCES": false
 }
--- a/docker/local-docker-config.json
+++ b/docker/local-docker-config.json
@@ -6,11 +6,5 @@
     "accessKeyId": "local-identity",
     "secretAccessKey": "local-credential",
     "s3ForcePathStyle": true
-  },
-  "AUTH0": {
-    "DOMAIN": "phire.auth0.com",
-    "CLIENT_ID": "qbAdutLVRV0rgvghVkYDrr55CRWR4saD",
-    "CLIENT_SECRET": "dCs7ESt1NZuCvWH6zkGfaQICEmKxjcRFs1fwUMDxj1NP9dLRAQSIux8iaGGzrs1O"
-  },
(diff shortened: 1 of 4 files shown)


docker/local-docker-config.json
```diff
@@ -6,11 +6,5 @@
...
     "secretAccessKey": "local-credential",
     "s3ForcePathStyle": true
-  },
-  "AUTH0": {
-    "DOMAIN": "phire.auth0.com",
-    "CLIENT_ID": "qbAdutLVRV0rgvghVkYDrr55CRWR4saD",
-    "CLIENT_SECRET": "dCs7ESt1NZuCvWH6zkGfaQICEmKxjcRFs1fwUMDxj1NP9dLRAQSIux8iaGGzrs1O"
-  },
-  "IMPORT_SENTENCES": false
+  }
```
rileyjshaw: 😅 love this commit name

web/package.json
```diff
@@ -49,6 +49,7 @@
     "postcss-preset-env": "^6.7.0",
     "preload-webpack-plugin": "^3.0.0-beta.3",
     "promise-decode-audio-data": "^0.4.1",
+    "query-string": "^6.8.3",
```
rileyjshaw: I'm weary of adding a new dependency for this, since we're going to strip it out of the code in a few days.

web/src/components/pages/dashboard/challenge/constants.ts
```diff
@@ -26,7 +27,31 @@ export const isBeforeChallenge = (challenge: ChallengeDuration) => {
...
+
+  let challengeDates: ChallengeDuration = defaultDates;
+
+  if (params.challenge_start && isValidDate(new Date(params.challenge_start))) {
+    challengeDates.start = new Date(params.challenge_start);
+  }
+
+  if (params.challenge_end && isValidDate(new Date(params.challenge_end))) {
+    challengeDates.end = new Date(params.challenge_end);
+  }
```
rileyjshaw: Let's discuss this when you're free. My intuition is that having a single `date=` param will allow us to test against the challenge dates easier, and move us closer to being able to test specific weeks.

web/src/components/pages/dashboard/challenge/constants.ts
```diff
@@ -26,7 +27,31 @@ export const isBeforeChallenge = (challenge: ChallengeDuration) => {
...
+
+  return challengeDates.end > challengeDates.start
+    ? challengeDates
+    : defaultDates;
+};
```
rileyjshaw: Should we allow folks to set a single param? Right now it looks like both `challenge_start` and `challenge_end` need to exist for anything to change.

PR #2303, opened 2019-11-13
earn three-day-streak bonus every time a 3-day streak is reached

can-007:
earnBonus('three_day_streak'...) is a BUG before, it will return sooner than it should.

@rileyjshaw , suppose earn the bonus for every 3-day streak, not 5-day streak.

### DIFF
--- a/server/src/lib/model/goals.ts
+++ b/server/src/lib/model/goals.ts
@@ -349,16 +349,16 @@ export async function checkGoalsAfterContribution(
         VALUES (?,     ?,    ?,         ?,         NOW()     )`,
       [threshold, type, client_id, localeId]
     );
-
-    // reached a 3-day streak, and try to earn the three_day_streak bonus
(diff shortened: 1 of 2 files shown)


server/src/lib/model/goals.ts
```diff
@@ -349,16 +349,16 @@ export async function checkGoalsAfterContribution(
...
+  // reached a 3-day streak, and try to earn the three_day_streak bonus
+  let earned = false;
+  if (challenge && streak_days === 3) {
+    earned = await earnBonus('three_day_streak', [
+      client_id,
+      client_id,
+      challenge,
+    ]);
   }
+  return earned;
```
rileyjshaw: The achievement requirement is:

> "Set a goal and complete a three day streak"

This looks good to me, but I'm going to start a discussion in tonight's meeting to make sure this is the behaviour we're looking for.

PR #2317, opened 2019-11-14
Fix: Team progress ranking display should depend on the current week

catherine917:
For now, the data structure for the top team from the back-end is like below:
``` json
[{
 "position":0,
"id":13,
"name":"SAP",
"w1_points":"100",
"w1":"1",
"w2_points":"100",
"w2":"1",
"w3_points":"100",
"w3":"1",
"nextRank":2,
"avatarClipUrl":null,
"clientHash":"***",
"you":false
}]
```
`w1,w2,w3` is the weekly ranking of a team separately. But for the front-end,  I need to show the ranking of the current week

### DIFF
--- a/web/src/components/pages/dashboard/challenge/challenge-list.tsx
+++ b/web/src/components/pages/dashboard/challenge/challenge-list.tsx
@@ -189,7 +189,7 @@ class ChallengeList extends React.Component<Props, State> {
               </div>
             </div>
             <div className="week" title="Week">
-              {row.w1_points && week === 0 ? (
+              {row.w1_points && week >= 0 ? (
                 <>
                   <PointsIcon
                     className={row.w1 <= 3 ? `star-points-${row.w1}` : ''}
@@ -201,7 +201,7 @@ class ChallengeList extends React.Component<Props, State> {
               )}
             </div>
             <div className="week" title="Week">
-              {row.w2_points && week === 1 ? (
+              {row.w2_points && week >= 1 ? (
                 <>
                   <PointsIcon
                     className={row.w2 <= 3 ? `star-points-${row.w2}` : ''}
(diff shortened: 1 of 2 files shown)


rileyjshaw:
@catherine917 can you please link to the issue this addresses? It will be helpful for linking / closing it.

As for the PR: I think we want to be able to show multiple weeks, as shown in the mockups:

image

this has also made me realize that there's another state that I think we're missing:

image

I believe that second state is what we should see during the challenge.

can-007:
Top team is made of team bonus + all team member’s points?
Currently, only team bonus is calculated.

catherine917:
@rileyjshaw Now we don't have an issue to reflect this bug. And I think the new mockup looks like below:
Screen Shot 2019-11-15 at 9 25 46 AM

rileyjshaw:
Oops, sorry for pasting an old screenshot here. You can ignore the first screenshot, sorry for the confusion. The second screenshot is correct.

I think we want to show the places each team was in on each completed week. So in Week 2, we should show the Week 1 places as well as the Week 2 places.

We may need to add something to the database to track this 😬 do you mind taking a first attempt at it?

catherine917:
Understand. I will take a look at it and modify the code a little bit

can-007:
@rileyjshaw , the backend returns data for all three weeks. 
Backend returns `w1 `and `w1_points` by calculating the _team bonus_ earned in first week;
Backend returns `w2` and `w2_points `by calculating the _team bonus_ earned in the first two week.
Backend returns `w3 `and `w3_points `by calculating the _team bonus_ earned in the first three week.
So that the frontend could choose how to show them based on current week.

I think it is a mistake to calculate the _team bonus_ in this way, because the **first week's** _team bonus_ maybe earned in the **second week** since it is done manually.
And it is a big mistake that backend did not calculate the team member points in Top Team request. I am thinking how to calculate that.

rileyjshaw:
Nice! Thanks for switching `===` for `<=`. Looks good now!

PR #2320, opened 2019-11-14
earn streak bonus once only per user

can-007:
limiting new conditions for three-day-streak bonus: each user only earn it once, when contributing for 3 consecutive days after setting a custom_goal;


can-007:
@rileyjshaw , mainly two places:
1) added one condition to `LEFT JOIN streaks`:
`AND TIMESTAMPDIFF(DAY, streaks.started_at, streaks.last_activity_at) >= 3`.
2) move `earnBonus('three-day-streak')` out of `checkGoalsAfterContribution()`.
I hoped the streak_days in `checkGoalsAfterContribution() ` could save a lot of `earnBonus()` calls. But it was not right, because the `three-day-streak` actually means contributing for 3 consecutive days, not the `3-day-streak `of `reached_goals`.

rileyjshaw:
Thanks Can! Looking into this now.

server/src/lib/model/achievements.ts
```diff
@@ -27,27 +27,39 @@ const bonus_condition_sql = {
...
+  // 3) already set a custom goal during the challenge (after challenge start_date and before the end_date)
+  // 4) has contributed for 3 consecutive days (which is what 3-day-streak means) after setting the custom goal
+  //
+  // [NOTE]
+  // 1) challenges, enroll, streaks and earn are LEFT JOINed with at most one row, but the custom_goals may involve multiple rows.
+  // Thus, there is a GROUP BY clause.
+  // 2) a complex case may show the requirement well:
+  //    A streak began at day-1 and continued for 9 days;
+  //    At day 2 and day 3, two custom_goals were set separately;
+  //    Then at day 5, the user earned the bonus and wouldn't earn the bonus again.
```
rileyjshaw: Thanks, this is really illustrative.

server/src/lib/model/achievements.ts
```diff
@@ -27,27 +27,39 @@ const bonus_condition_sql = {
...
-    LEFT JOIN custom_goals ON custom_goals.client_id = enroll.client_id
-        AND custom_goals.created_at BETWEEN start_date AND TIMESTAMPADD(WEEK, 3, start_date)
-    LEFT JOIN streaks ON streaks.client_id = enroll.client_id
-        AND streaks.started_at BETWEEN start_date AND TIMESTAMPADD(DAY, 18, start_date)  # 18 days = 3 weeks - 3 days.
-        AND TIMESTAMPDIFF(DAY, streaks.started_at, streaks.last_activity_at) >= 3
-    LEFT JOIN earn ON enroll.client_id = earn.client_id
-        AND earn.achievement_id = achievements.id
-        AND earned_at BETWEEN TIMESTAMPADD(DAY, -3, streaks.last_activity_at) AND streaks.last_activity_at
-    WHERE challenges.url_token = ?
+    SELECT MAX(win_bonus) AS win_bonus, bonus_winner
```
rileyjshaw: Probably a silly question, but why is this in a `MAX()`?

server/src/lib/model/achievements.ts
```diff
@@ -27,27 +27,39 @@ const bonus_condition_sql = {
...
+        LEFT JOIN streaks ON streaks.client_id = enroll.client_id
+            AND streaks.started_at BETWEEN start_date AND TIMESTAMPADD(DAY, 18, start_date)  # 18 days = 3 weeks - 3 days.
+            AND TIMESTAMPDIFF(DAY, streaks.started_at, streaks.last_activity_at) >= 3
+            AND TIMESTAMPDIFF(DAY, custom_goals.created_at, streaks.last_activity_at) >= 3
+        LEFT JOIN earn ON enroll.client_id = earn.client_id
+            AND earn.achievement_id = achievements.id
+            AND earned_at BETWEEN TIMESTAMPADD(DAY, -3, streaks.last_activity_at) AND streaks.last_activity_at
+        WHERE challenges.url_token = ?
+    ) bonus
+    GROUP BY bonus_winner
```
rileyjshaw: 😅 yet another epic query. Looks good to me!

server/src/lib/model/achievements.ts
can-007: If user creates 2 or more custom_goals, there will be 2 or more rows selected. To make it return at most 1 row, use group by with aggregation function: MAX(). Some row may have win_bonus being 1 (true), some 0 (false). MAX() here is to select the biggest win_bonus, if it’s 1, allow the bonus insertion; otherwise disallow it.

PR #2331, opened 2019-11-14
Fix #2311: Possible to get multiple invite_send achievements

### LINKED ISSUE #2311: Possible to get multiple invite_send achievements

can-007:
When DB is outside the challenge time frame, there are problems, possible to get multiple invite_send achievements is one of them.

This PR could provide some suggestions for the case that no challenge is currently going on:
- protect the `hasEarnedBonus()` of achievements.ts from destructing properties of null;
- disallow to get the `invite_send `and `invite_contribute_same_session `  achievements;

@rileyjshaw , I test it on my local environment not the docker one, it could work.
We had brief discussion regarding what it should look like when there is no challenge currently, I only remembered a suggestion to disallow the enrollment.
@phirework , FYI.

I can think of the 4 situations above for now, any more?


phirework:
The redirect bug has been fixed on master, so you should be able to test it locally if you merge master back in. Looking at code now.

can-007:
Not completely fixed this one, the account is null, should check this one first.
Because the first time user ever visited the common-voice, there won't be any information about any user, so the account is null.

phirework:
@mbransn Can you confirm whether it's desired behaviour for users to be prevented from enrolling if there's no ongoing challenge? Don't we want users to be able to pre-register for a challenge before it starts?

rileyjshaw:
> Don't we want users to be able to pre-register for a challenge before it starts?

Can confirm that we definitely want this.

can-007:
Oh, that is different from what I thought. OK.
But, what could these pre-enrolled user do? Can they win bonus or see the challenge board?

rileyjshaw:
> But, what could these pre-enrolled user do? Can they win bonus or see the challenge board?

@can-007 no, they would just be signed up and waiting for the challenge. The only difference is that they would have a Challenge tab on the dashboard (which would show the pre-challenge state)

can-007:
@rileyjshaw , I see now, the UI is controlled by the constants. Nice.

mbransn:
+1 to everything Riley said.

rileyjshaw:
@can-007 what is the status on this? Does it require updates based on the above conversation, or should I review the code as-is?

can-007:
Please review the code as-is after a proper rest. I updated the branch based on the above conversation.

can-007:
A little bit more detail: this could prevent achieving multiple `invite_send `and `invite_contribute_same_session `achievements when now no challenges are going on.

server/src/lib/model/achievements.ts
```diff
@@ -73,9 +74,9 @@ const bonus_condition_sql = {
     GROUP BY user_clients.client_id
     `,
   // Arguments: [client_id, client_id, challenge]
-  // Check if user has already earned the bonus
+  // Check if user has already earned the bonus, and if challenge is on-going now
   invite_send: `
-    SELECT earn.id IS NULL AS win_bonus, ? AS bonus_winner
+    SELECT (earn.id IS NULL) AND (NOW() BETWEEN start_date AND TIMESTAMPADD(WEEK, 3, start_date)) AS win_bonus, ? AS bonus_winner
```
rileyjshaw: This should still be achievable before the challenge. We have a pre-challenge state that includes the invite modal.

server/src/lib/model/achievements.ts
```diff
@@ -84,9 +85,9 @@ const bonus_condition_sql = {
     WHERE challenges.url_token = ?
     `,
   // Arguments: [client_id, client_id, challenge]
-  // Check if user has already won the bonus
+  // Check if user has already won the bonus, and if challenge is on-going now
   invite_contribute_same_session: `
-    SELECT earn.id IS NULL AS win_bonus, ? AS bonus_winner
+    SELECT (earn.id IS NULL) AND (NOW() BETWEEN start_date AND TIMESTAMPADD(WEEK, 3, start_date)) AS win_bonus, ? AS bonus_winner
```
rileyjshaw: Same here - this should be available pre-challenge.

server/src/lib/model/achievements.ts
```diff
@@ -119,7 +120,8 @@ export const hasEarnedBonus = async (
   client_id: string,
   challenge: ChallengeToken
 ) => {
-  const [[{ earned }]] = await db.query(
+  // do not destructure to [[{earned}]], just in case null values returned, in scenarios like challenge=null
+  const [[res]] = await db.query(
```
rileyjshaw: This is fixed in https://github.com/mozilla/voice-web/pull/2373, can be safely removed.

phirework: omg I'm so annoyed that this comment was sitting in a pull request all along 😅

can-007: Yeah.

server/src/lib/model/achievements.ts
can-007: Done

server/src/lib/model/achievements.ts
can-007: Good point.

PR #2335, opened 2019-11-15
Fix #2336: The Top Team leaderboard does not take into account of team member points

### LINKED ISSUE #2336: The Top Team leaderboard does not take into account of team member points
I contributed and earned some bonus, but the points did not reflect on the Top Team leaderboard.

can-007:
change the meaning of Top Teams: user clips + user votes + user bonus + team bonus;

**Explanation**:
Top Team return json like the following:
```
[
  {
    "position": 0, "id": 13, "name": "SAP",
    "w1_points": "100", "w1": "1",
    "w2_points": "100", "w2": "1",
    "w3_points": "100", "w3": "1",
    "nextRank": 2, "avatarClipUrl": null, "clientHash": "***", "you": false
  }
]
```
The json **meaning**:
- `w1`, `w2`, `w3` is the team rankings in the first, second, third week respectively.
    `w1_point`, `w2_point`, `w3_point `is the _team_points_ in the first, second, third week respectively.
- `w2 `take into account of the `w1_point` and `w2_point`.
    `w3 `take into account of the `w1_point`, `w2_point `and `w3_point`.
- `w2_point `include the `w1_point `
   `w3_point `include `w2_point `and `w1_point`.
    _team_points_ = _team_bonus_ + _all_member_bonus_ + _all_member_points_.

@rileyjshaw, @phirework  , this is a possible way to re-implement the Top Team leaderboard, taking into account of team member points. You can see that the SQL is quite terrifying.
I test it locally, but the number of different cases are just too large. I only tested clip points, individual bonus, team bonus earned in different weeks and the earning could be reflected on the points and rankings.

The alternative is add more tables in database to store these points, as you said, which would be much easier to write lean SQL to do the job, but it require quite some efforts to maintain such tables and it would be harder to test the functionality locally.

### DIFF
--- a/server/src/lib/model/leaderboard.ts
+++ b/server/src/lib/model/leaderboard.ts
@@ -183,7 +183,14 @@ async function getTopListeners({
   return rows;
 }
 
-// this SQL is in low quality, but it does the job: showing all teams' rankings across 3 weeks.
+// show all teams' rankings across 3 weeks.
+// NOTE: explain the meaning of the returned result of the terrifying SQL.
+// 1) w1, w2, w3 is the team rankings in the first, second, third week respectively.
+//    w1_point, w2_point, w3_point is the _team_points_ in the first, second, third week respectively.
+// 2) w2 take into account of the w1_point and w2_point.
+//    w3 take into account of the w1_point, w2_point and w3_point.
+// 3) w2_point include the w1_point and w3_point include w2_point and w1_point.
+// _team_points_ = team_bonus + all_member_bonus + all_member_points.
 // [FUTURE] there are maybe several ways to optimize the SQL here
 //          and when new version of DB is available, ranking() and window function could be used
 async function getTopTeams(challenge: ChallengeToken): Promise<any[]> {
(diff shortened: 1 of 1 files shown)


phirework:
I would suggest that this is a case where the small performance hit of doing 3 separate queries is worth the codebase being more easily tested and maintained in the future. These are pretty complex data points we're trying to get, there's no reason it HAS to return all the formatted data in one query. Alternatively, I would get the raw data for the entire duration and do week-based calculations in JS instead of SQL. Thoughts @rileyjshaw?

rileyjshaw:
> I would suggest that this is a case where the small performance hit of doing 3 separate queries is worth the codebase being more easily tested and maintained in the future.

Absolutely agree. We can play around with performance improvements down the road with caching and optimization, but for now clarity and maintainability are more important.

> Alternatively, I would get the raw data for the entire duration and do week-based calculations in JS instead of SQL. Thoughts @rileyjshaw?

100%. In the future, I think we need to move in this direction anyway; doing all of our date logic in SQL isn't scaleable, and it's becoming increasingly difficult to make changes.

can-007:
It makes sense.
I am just offering a quick suggestion.

rileyjshaw:
Thanks for the hard work on this Can, but I think we want the Overall Challenge Top Team board to only reflect the weekly challenge bonuses, like so:

image

@mbransn can you confirm?

rileyjshaw:
Maybe we can repurpose this for the Team Progress board, as shown in the screenshot below?

image

rileyjshaw:
To further clarify, the weekly challenges shown in this board:

image

will be calculated and awarded manually at the end of each week.

mbransn:
> To further clarify, the weekly challenges shown in this board:
> 
> image
> 
> will be calculated and awarded manually at the end of each week.

+1 and the calculations in this mock up are actually wrong (apologies) -- SAP would have 300 total; Mozilla would have 220 total; IBM would have 200 total -- based on the manual tally of the points awarded for each ranking week over week.

can-007:
This PR doesn't serve its original purpose any more.
It only shows how long the SQL could be and the comments show how such long SQL should be handled.
Thus, close this PR.

Issue #2328, opened 2019-11-14
Cannot create a new account in common voice from the invite link 


viorelaioia:
**STR:** 
1. Navigate to a challenge invite link (e.g: https://voice.allizom.org/?challenge=pilot&team=ibm&invite=7c2be817-0477-11ea-bb30-067a2205f3e0)
2. Accept terms and conditions and click "Join the IBM team" from the modal
3. Enter an email address never used before in the Common Voice
4. Click "Enter" and send login link
5. Login to your email server and click the login link from the email

**Expected:**
User is redirected to the profile page to create a common voice profile

**Actual:**
Error message is returned for a few seconds, then user is redirected to login page. Browser console shows no errors.  

![sign up to team-2019-11-14_16 32 23(1)](https://user-images.githubusercontent.com/5293100/68867639-dac49e80-06fe-11ea-94e8-71ff451fe7e5.gif)

**Note:**
Following steps 1-2, then using the google login with an account not used in common voice before returns login page with errors in browser console.
See below.
![image_2019_11_14T14_25_18_005Z](https://user-images.githubusercontent.com/5293100/68867901-41e25300-06ff-11ea-8a0d-489fa8aaf00a.png)

phirework:
@can-007 @catherine917 I've reverted the main commits that I think could be affecting this and it's still persisting. Can you take a look at the pieces of `api.ts` that are throwing up the error and see what's up? 

This is reproducible on my local using docker, let me know if you run into issues with that. 

cc @rileyjshaw

can-007:
I tried the allizom.org, using the linked provided above. But the allizom.org could redirect to the Profile page correctly.
I think the root cause is the same as https://github.com/mozilla/voice-web/issues/2324.

rleitan:
The issue is still reproducible. I can see the Build profile page but when I click  Save button the error is displayed

phirework:
Confirming that @rleitan's description of the problem is still reproducible. You have to click save on the profile page redirect to get the error.

can-007:
That is strange.
@phirework , I have a theory that before the challenge board is loaded, the front-end sends out a request for account. The account contains user's enrollment information. And before the account is returned, challenge board is loaded, and send requests for challenge information like points, current week, etc. But these requests need the enrollment information which is not available at that time.
Thus, crashes the front-end.

phirework:
@rleitan This should be fixed on stage! Crossing all my fingers :D

rileyjshaw:
[...]

PR #2342, opened 2019-11-15
Check if user exists before making any API calls for challenge data

phirework:
Thanks to @can-007 for support 👍

### DIFF
--- a/web/src/components/pages/dashboard/dashboard.tsx
+++ b/web/src/components/pages/dashboard/dashboard.tsx
@@ -224,7 +224,7 @@ const ChallengeBar = ({ isNarrow, setShowInviteModal }: ChallengeBarProps) => {
   const [points, setAllPoints] = useState({ user: 0, team: 0 });
 
   useEffect(() => {
-    api.fetchChallengePoints().then(setAllPoints);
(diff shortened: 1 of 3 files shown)


phirework:
I'm hoping this entirely fixes #2328 but I guess we'll know in <looks at watch> 20-40 minutes haha. Agree on extracting pattern into its own utility, it's all over the place at this point. Will add to my list.

web/src/components/pages/dashboard/dashboard.tsx
```diff
@@ -224,7 +224,7 @@ const ChallengeBar = ({ isNarrow, setShowInviteModal }: ChallengeBarProps) => {
   const [points, setAllPoints] = useState({ user: 0, team: 0 });
 
   useEffect(() => {
-    api.fetchChallengePoints().then(setAllPoints);
+    api.fetchChallengePoints().then(setAllPoints); // TODO: it shouldn't even try to fetch any challenge data in useEffect if not enrolled
   }, []);
```
rileyjshaw: @phirework we'll see what happens on staging, but this line gives me pause. Passing `[]` into useEffect means it's only going to run once per mount. If the first attempt fails because the enrollment isn't there yet, it won't try again when `user` changes.

We may need to figure out _why_ `user` is updating asynchronously and hook into that flow.

Issue #2344, opened 2019-11-15
Challenge dashboard should be default dashboard for those enrolled in an active challenge


rleitan:
[Precondition]:
Join the voice challenge

[STR]:
1.Go to challenge dashboard
2.Click the Listen/Speak button
3.Click the back arrow

[Expected result]:
The page is redirected to https://voice.allizom.org/en/dashboard/stats 

[Actual result]:
The page should be redirected back to the Challenge dashboard
https://drive.google.com/file/d/1qmu4wwke0aVpk6l-9fPQ95qK29cBsrv4/view?usp=sharing

PR #2350, opened 2019-11-15
Hide /challenge route for users who aren't enrolled

rileyjshaw:
Also sets the default tab to /challenge for enrolled users, which should fix https://github.com/mozilla/voice-web/issues/2344.

### DIFF
--- a/web/src/components/pages/dashboard/challenge/constants.ts
+++ b/web/src/components/pages/dashboard/challenge/constants.ts
@@ -1,4 +1,5 @@
 import { ChallengeDuration, ChallengeTeamToken } from 'common/challenge';
+import { UserClient } from 'common/user-clients';
 import { isProduction } from '../../../../utility';
 
 export const challengeLogos: {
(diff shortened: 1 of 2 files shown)


web/src/components/pages/dashboard/dashboard.tsx
```diff
@@ -349,7 +355,7 @@ export default function Dashboard() {
                     render={() => (
                       <Redirect
                         to={toLocaleRoute(
-                          URLS.DASHBOARD + '/' + dashboardLocale + URLS.STATS
+                          URLS.DASHBOARD + '/' + dashboardLocale + defaultPage
```
phirework: Awesome.

web/src/components/pages/dashboard/challenge/constants.ts
```diff
@@ -43,3 +44,6 @@ export const pilotDates: ChallengeDuration = {
   start: new Date(!isProduction() ? '2019-11-10' : '2019-11-18'),
   end: new Date(!isProduction() ? '2019-11-30' : '2019-12-08'),
 };
+
+export const isEnrolled = (account: UserClient) =>
```
phirework: This will need to go into `api.ts` as well, though I imagine that'll get cleaned up as we investigate the login bug.

PR #2351, opened 2019-11-15
RFC: Switch enrollment condition

rileyjshaw:
I think we might be seeing some bugs from this enrollment flow, but I
can't test thoroughly enough on my local branch to know for sure.

My thoughts:

If the enrollment fails, we should not add user enrollment info or
redirect to a challenge URL. We're seeing confusing states where someone's
challenge enrollment fails, but they still see the Team Only visibility
option on their profile page.

I'm hoping this fixes some of our other enrollment woes.

@can-007, you were the initial author of this section of the code - I'd
love to hear your thoughts / comments.

### DIFF
--- a/server/src/auth-router.ts
+++ b/server/src/auth-router.ts
@@ -119,34 +119,32 @@ router.get(
         session.passport.user = old_user;
       }
       response.redirect('/profile/settings?success=' + success.toString());
-    } else if (enrollment && enrollment.challenge && enrollment.team) {
-      if (
-        !(await UserClient.enrollRegisteredUser(
-          user.emails[0].value,
-          enrollment.challenge,
-          enrollment.team,
-          enrollment.invite,
-          request.header('Referer')
-        ))
-      ) {
-        // if the user is unregistered, pass enrollment to frontend
-        user.enrollment = enrollment;
-      } else {
-        // if the user is already registered, now he/she should be enrolled
-        // [TODO] there should be an elegant way to get the client_id here
-        const client_id = await UserClient.findClientId(user.emails[0].value);
-        await earnBonus('sign_up_first_three_days', [
-          enrollment.challenge,
-          client_id,
-        ]);
-        await earnBonus('invite_signup', [
-          client_id,
-          enrollment.invite,
-          enrollment.invite,
-          enrollment.challenge,
-        ]);
-      }
-
+    } else if (
+      enrollment &&
+      enrollment.challenge &&
+      enrollment.team &&
+      (await UserClient.enrollRegisteredUser(
+        user.emails[0].value,
+        enrollment.challenge,
+        enrollment.team,
+        enrollment.invite,
+        request.header('Referer')
+      ))
+    ) {
+      // if the user is already registered, now he/she should be enrolled
+      // [TODO] there should be an elegant way to get the client_id here
+      const client_id = await UserClient.findClientId(user.emails[0].value);
+      user.enrollment = enrollment; // TODO: Do we need this?
+      await earnBonus('sign_up_first_three_days', [
+        enrollment.challenge,
+        client_id,
+      ]);
+      await earnBonus('invite_signup', [
+        client_id,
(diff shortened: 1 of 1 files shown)


server/src/auth-router.ts
```diff
@@ -119,34 +119,32 @@ router.get(
...
+        enrollment.challenge,
+        enrollment.team,
+        enrollment.invite,
+        request.header('Referer')
+      ))
+    ) {
+      // if the user is already registered, now he/she should be enrolled
+      // [TODO] there should be an elegant way to get the client_id here
+      const client_id = await UserClient.findClientId(user.emails[0].value);
+      user.enrollment = enrollment; // TODO: Do we need this?
```
phirework: Seconding Q of whether we need this @can-007

can-007: I think we need this.
This enroll function here is for enrolling registered user only.
For brand new user, the enrollment happens when saving the profile. And the enrollment information will be needed then.

I think this PR will prevent brand new user from enrolling. He/she has to enroll again to get enrolled here.

can-007:
I would like to share my understanding of the signup flow when I first read the code.

New user passes the SSO authentication to get here. But the user is not register yet. When user click the save button in the Profile page for the first time, his/her email gets saved, it means he/she is registered.

How could the backend know the user’s enrollment info when user save his/her profile? Only place to pass that info is here.

can-007:
Maybe we can use the client_id to determine whether or not to add enrollment info or redirection url.

client_id here is based on email. Null client_id means it is brand new user, otherwise a registered user.

phirework:
No, that's incorrect. All active users on the site are assigned a client_id - there are tens of thousands of records in the database that have a client_id but no email. You can test this yourself locally: 

* Go into your SQL database and make a note of how many users are in user_clients
* Open up a browser that has never seen `http://localhost:9000` OR clear all your history and session data for a test browser, then go onto the site and click around a bit
* Go back into your SQL database and make a note of how many users are in user_clients - you should have at least one more
* Try to register a normal user without enrolling
* At the "build profile" step, the number of user_clients remains unchanged
* If you agree to the privacy policy and hit save, your email address is attached to the new token from step 3, and the number of user_clients still remains unchanged

So it's perfectly possible to attach enrollment data to just a user_client at this step. In the database, the key constraints are all client_id anyway, not email.

can-007:
Oh, perhaps I didn’t understood it correctly, I thought enrollment only made sense to user_clients with authenticated emails.

So, if it’s Ok to enroll any user_client (with email or not), I think you are right, we don’t need to pass the enrollment info around, we should delete that line.

And we should delete the enroll function from saveAccount() in UserClient of backend, making this callback function the only place to enroll users.

phirework:
We ARE only enrolling user clients with authenticated emails. The act of
registering a full profile associates an email address with a client_id.
They should enroll in teams at the same time as they save their email
address.

Are you saying that the point of this line is to give the enrolment data to
UserClient.saveAccount()? We were confused because the comment said that
the data was being passed to the *frontend*.

can-007:
yes, that is my point indeed.
There is a data flow of the enrollment: callback of backend -> login.ts of frontend -> profile of frontend -> saveAccount() of frontend -> saveAccount() of backend.

can-007:
And the client_id HERE is queried from the database by checking the email.

rileyjshaw:
Thanks for the discussion, all! I have a much better understanding of this code now.

Based on my understanding, it sounds like I should close this commit and leave everything as-is. Apart from commenting & rearranging for clarity, can anyone think of any immediate improvements we need to make here before the challenge launches?

can-007:
I only have one, but not here, it’s about https://github.com/mozilla/voice-web/pull/2335, the Top Team leaderboard.

Currently the leaderboard only calculates the team bonus.  The PR will take team members’ points into consideration.

rileyjshaw:
I've added some comments to the outstanding PRs. Closing this one, thanks again for the discussion!

PR #2352, opened 2019-11-16
Move the challenge redirect on new-profile-build inside of async

phirework:


web/src/components/pages/profile/info/info.tsx
```diff
@@ -172,16 +172,17 @@ function ProfilePage({
         if (!(user.account && user.account.basket_token) && sendEmails) {
           await api.subscribeToNewsletter(user.userClients[0].email);
         }
+
         saveAccount(data);
```
rileyjshaw: This isn’t behind an `await` - should we wait for this to complete before redirecting? Or is it just saving data client-side?

phirework: My read of these pieces is that nothing is actually being done with the response `saveAccount` on the frontend, it does a whole bunch of other work asynchronously at `stores/user.ts` but nothing as far as I can tell that we need to wait for. If it was safe to set `isSaving` to false and pop up a notification without an `await`, I would say the redirect should be too. I could be wrong! I guess it wouldn't hurt to add an await, other than making users wait a bit longer to register.

Also, when I was debugging the flow here, the reason the users weren't being logged in properly initially is because they were hitting the redirect before they hit the `addUploads` action, so nothing inside there was even being called.

phirework: Actually, in local testing I'm getting a "you're getting redirected even though there may be unsubmitted data" so I'm gonna throw an await in there.

PR #2362, opened 2019-11-18
Improve challenge referer tracking

rileyjshaw:
TIL [`document.referrer`](https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer) is standard across all major browsers, and reports the same value we would see on the server. This reduces the headache of sending it from the server, to the client, back to the server, that we were discussing earlier.

This change means we'll get the original inbound referer, and will hopefully catch more sources like Twitter, Facebook, etc. Fingers crossed!

Note: Referrer? Referer? Typo [theirs](https://tools.ietf.org/html/rfc1945#section-10.13), not mine :)

### DIFF
--- a/common/challenge.ts
+++ b/common/challenge.ts
@@ -35,7 +35,8 @@ export type ChallengeTeamToken = 'ibm' | 'mozilla' | 'sap';
 export type Enrollment = {
   challenge: ChallengeToken;
   team: ChallengeTeamToken;
-  invite: string;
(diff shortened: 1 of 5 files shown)


rileyjshaw:
Note: this is going to be difficult to test until it's on staging, since localhost runs on HTTP and thus rarely has `document.referrer` set. I'm going to land this now, but please review with comments when you get a chance @phirework.

PR #2363, opened 2019-11-18
Fix: Achievements display on challenge dashboard

catherine917:
Now the achievements on challenge dashboard look like below: 
I have modified the styles in PR https://github.com/mozilla/voice-web/pull/2272. But the styles have been overwritten. So remove it again. 
Annotation 2019-11-18 141926


rileyjshaw:
Great, thanks! Tested and confirmed that dashboard achievements still look correct when they appear.

PR #2373, opened 2019-11-18
Fix type error on achievements sql

phirework:
This is a tiny change for a big bug, so here's my logic as to what's happening: 

* The clips are still being stored in the db and uploaded to S3
* If you comment out the upload code locally (153-215 in `server/src/lib/clip.ts`) you can fake submitting clips without actually uploading
* A faked submission gave me the following error: 

`web        | [BE] TypeError: Cannot destructure property `earned` of 'undefined' or 'null'.`
`web        | [BE]     at Object.exports.hasEarnedBonus (/code/server/src/lib/model/achievements.ts:122:26)`
`web        | [BE]     at <anonymous>`
`web        | [BE]     at process._tickDomainCallback (internal/process/next_tick.js:229:7)`

Fake submissions work now on my local. The intermittentness of the error, and particularly why Lindsay and I were able to submit clips, makes sense in context of the error only showing up if earned achievements are null, whereas both Lindsay and I have done enough testing that we probably have achievements everywhere. Let me know if this makes sense to you.

### DIFF
--- a/server/src/lib/model/achievements.ts
+++ b/server/src/lib/model/achievements.ts
@@ -118,7 +118,7 @@ export const hasEarnedBonus = async (
   type: AchievementType,
   client_id: string,
   challenge: ChallengeToken
-) => {
+): Promise<boolean> => {
   const earned = await db.query(
     `
       SELECT earn.client_id IS NOT NULL AS earned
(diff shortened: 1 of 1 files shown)


rileyjshaw:
Great solve!! Thanks for walking me through the process, solid debugging of a yucky issue 💯

rileyjshaw:
(Waiting for Travis before hitting the big green button)

PR #2377, opened 2019-11-18
Remove some bug vectors from profile's async submit / redirect flow

rileyjshaw:
re: comments on https://github.com/mozilla/voice-web/commit/9e984e021debff09efe33ddece045b4898e096c3, a few of these `await` calls weren't actually `await`ing anything.

I'm going to look into cleaning up & simplifying the profile save behaviour. Before diving in, I wanted to restructure the code to indicate what's actually happening right now.

### DIFF
--- a/web/src/components/layout/content.tsx
+++ b/web/src/components/layout/content.tsx
@@ -76,7 +76,7 @@ export default function Content() {
             path={toLocaleRoute(URLS.DASHBOARD)}
             component={DashboardPage}
           />
-          {[URLS.STATS, URLS.GOALS, URLS.AWARDS, URLS.CHALLENGE].map(path => (
+          {[URLS.CHALLENGE, URLS.STATS, URLS.GOALS, URLS.AWARDS].map(path => (
             <Route
               key={path}
               exact
--- a/web/src/components/layout/layout.tsx
+++ b/web/src/components/layout/layout.tsx
@@ -16,7 +16,7 @@ import {
   isStaging,
   replacePathLocale,
 } from '../../utility';
-import { LocaleLink, LocaleNavLink } from '../locale-helpers';
+import { LocaleLink, LocaleNavLink, useLocale } from '../locale-helpers';
 import {
   CogIcon,
   DashboardIcon,
--- a/web/src/components/pages/dashboard/dashboard.tsx
+++ b/web/src/components/pages/dashboard/dashboard.tsx
@@ -198,10 +198,10 @@ function DashboardContent({
   dashboardLocale,
 }: {
   Page:
+    | typeof ChallengePage
     | typeof StatsPage
     | typeof GoalsPage
-    | typeof AwardsPage
-    | typeof ChallengePage;
+    | typeof AwardsPage;
   dashboardLocale: string;
 }) {
   const api = useAPI();
@@ -264,6 +264,7 @@ export default function Dashboard() {
   ];
   let defaultPage = URLS.STATS;
   if (isChallengeEnrolled) {
+    // @ts-ignore
     pages.unshift({ subPath: URLS.CHALLENGE, Page: ChallengePage });
     defaultPage = URLS.CHALLENGE;
   }
--- a/web/src/components/pages/profile/info/info.tsx
+++ b/web/src/components/pages/profile/info/info.tsx
@@ -5,7 +5,7 @@ import {
 } from 'fluent-react/compat';
 import * as React from 'react';
 import { useCallback, useEffect, useState } from 'react';
-import { RouteComponentProps, withRouter } from 'react-router';
+import { Redirect, RouteComponentProps, withRouter } from 'react-router';
 import { useAction, useAPI } from '../../../../hooks/store-hooks';
 import { NATIVE_NAMES } from '../../../../services/localization';
 import { trackProfile } from '../../../../services/tracker';
@@ -25,6 +25,7 @@ import {
   LabeledInput,
   LabeledSelect,
 } from '../../../ui/ui';
+import { isEnrolled } from '../../dashboard/challenge/constants';
 
 import './info.css';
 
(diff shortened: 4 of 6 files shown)


phirework:
This doesn't seem to be redirecting for me locally. Tried with http://localhost:9000/en?challenge=pilot&team=sap and creating a new passwordless login.

rileyjshaw:
Thanks for the feedback, @phirework!

For me, the redirect triggers but a message appears when trying to leave:

"`This page is asking you to confirm that you want to leave - data you have entered may not be saved.`"

The culprit is an `onBeforeUnload` handler in `app.tsx` that blocks navigating from a page if there are any pending `Upload`s. We're trying to redirect from within an `addUploads` resolver... which should always fail! I think this was working initially due to a combination of `await` blocking the redirect, and some race condition.

I want to refactor things so that the redirect happens _after_ the `addUploads` handler has resolved. An annoying detail: all of this happens in a weird `componentWillReceiveProps` handler in `app.tsx`, and there's no Promise or Event I can easily hook into from `info.tsx` to tell us when `uploads[]` is empty.

I just made some pretty major changes, and pushed them here. I'm about to hop on the bus; once I'm home and I've eaten, I'll do a better write-up. For now, all I'll say is that it seems to work well for me, and removes a lot of potential bug vectors. This is ready for re-testing now, so please let me know if it works for you!

web/src/components/layout/content.tsx
```diff
@@ -76,7 +76,7 @@ export default function Content() {
             path={toLocaleRoute(URLS.DASHBOARD)}
             component={DashboardPage}
           />
-          {[URLS.STATS, URLS.GOALS, URLS.AWARDS, URLS.CHALLENGE].map(path => (
+          {[URLS.CHALLENGE, URLS.STATS, URLS.GOALS, URLS.AWARDS].map(path => (
```
rileyjshaw: This is entirely unrelated; just swapped it for consistency with other files.

web/src/components/pages/dashboard/dashboard.tsx
```diff
@@ -264,6 +264,7 @@ export default function Dashboard() {
   ];
   let defaultPage = URLS.STATS;
   if (isChallengeEnrolled) {
+    // @ts-ignore
```
rileyjshaw: I truly can't be bothered. Seeing the following errors:

```
web        | [FE] ERROR in [at-loader] ./src/components/pages/dashboard/dashboard.tsx:267:46
web        | [FE]     TS2322: Type 'ComponentClass<Pick<Props & RouteComponentProps<any, StaticContext, any>, "dashboardLocale" | "allGoals">, any> & WithRouterStatics<(props: Props & RouteComponentProps<any, StaticContext, any>) => Element>' is not assignable to type '({ allGoals, dashboardLocale }: Props) => Element'.
web        | [FE]   Type 'ComponentClass<Pick<Props & RouteComponentProps<any, StaticContext, any>, "dashboardLocale" | "allGoals">, any> & WithRouterStatics<(props: Props & RouteComponentProps<any, StaticContext, any>) => Element>' provides no match for the signature '({ allGoals, dashboardLocale }: Props): Element'.
```

It works as expected, but doesn't like that we've tacked the new React Router props onto `ChallengePage`. This is a problem for another day.

rileyjshaw:
I'm hoping that this fixes https://github.com/mozilla/voice-web/issues/2330, since it's actually giving our async functions time to resolve.

The `await`s we initially added were themselves handled in an asynchronous function, so there was no guarantee that they would complete before the redirect happened. By navigating client-side, the API requests now have all the time in the world to complete.

I wrote https://github.com/mozilla/voice-web/pull/2380 as a follow-on diff to test a few more email tweaks, but this change is the most promising IMO.

web/src/components/layout/layout.tsx
```diff
@@ -16,7 +16,7 @@ import {
   isStaging,
   replacePathLocale,
 } from '../../utility';
-import { LocaleLink, LocaleNavLink } from '../locale-helpers';
+import { LocaleLink, LocaleNavLink, useLocale } from '../locale-helpers';
```
phirework: This is the only change in this file, is this still being used?

web/src/components/pages/dashboard/dashboard.tsx
phirework: I'm good with it if you are.

web/src/components/pages/profile/info/info.tsx
```diff
@@ -193,6 +184,20 @@ function ProfilePage({
     return null;
   }
 
+  if (!isSaving && isSubmitted && isEnrolled(user.account)) {
+    return (
+      <Redirect
```
phirework: Did you run into any issues related to "can't perform state update on unmounted components" when you were working on this piece? `Redirect` worked so inconsistently for me. (This seems fine here, just wanted to ask!)

web/src/components/pages/profile/info/info.tsx
```diff
@@ -193,6 +184,20 @@ function ProfilePage({
...
+        to={{
+          pathname: toLocaleRoute(URLS.DASHBOARD + URLS.CHALLENGE),
+          state: {
+            showOnboardingModal: true,
+            earlyEnroll: window.location.search.includes('achievement=1'),
+          },
```
phirework: I like this so much better than just endlessly passing url params around, yay!

phirework:
I'm still not getting a redirect, but looking at my error console it looks like it may be because I borked something in my S3 setup that's interfering with this. Going to land this so that I can test this on staging. Great work, this is much cleaner.

web/src/components/pages/profile/info/info.tsx
rileyjshaw: Good question - I didn't notice any more than usual, but... that page was already throwing a few similar warnings, so I can't be sure.

web/src/components/pages/profile/info/info.tsx
rileyjshaw: 👍

web/src/components/layout/layout.tsx
rileyjshaw: Ah, good catch! Will remove it, thank you.

rileyjshaw: https://github.com/mozilla/voice-web/commit/9df4a5f58693c34da7a123a4b3c92dd1e95db330

(We should make a lint rule for this once the dust settles)

PR #2378, opened 2019-11-18
Tighten query for enrolling registered users

phirework:
Another simple fix! Rationale: 

* there are zero duplicate emails within the realm of `has_login = true`
* the `findClientId` method from Gregor uses `has_login` as a uniqueness check
  * this method was last updated in the same commit that created the `has_login` field
* `has_login` is set to true anytime `saveAccount` is called

¯\_(ツ)_/¯


rileyjshaw:
Hell ya, I love this commit!!!

PR #2382, opened 2019-11-19
Fix #2311: Possible to get multiple invite_send achievements (NEW PR)

### LINKED ISSUE #2311: Possible to get multiple invite_send achievements

can-007:
Before the challenge:
- prevent multiple `invite_send ` bonus
- prevent multiple `invite_contribute_same_session ` bonus as well;

After the challenge:
- prevent `invite_send `and `invite_signup `bonus;
- multiple `invite_contribute_same_session `bonus can't be avoid with this PR;

Code change is mainly about adding ` AND (NOW() < TIMESTAMPADD(WEEK, 3, start_date))` in SELECT clauses to determine earn the bonus or not.

@rileyjshaw , @phirework , this should fix https://github.com/mozilla/voice-web/issues/2311.
I have created this PR to replace the other PR https://github.com/mozilla/voice-web/pull/2331. 
Because the original branch of that PR has conflicts that can't be solved locally, git reported errors like "`Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`", seems like git-hook problem, but it doesn't matter.

### DIFF
--- a/server/src/lib/model/achievements.ts
+++ b/server/src/lib/model/achievements.ts
@@ -1,6 +1,7 @@
 import { ChallengeToken, AchievementType } from 'common/challenge';
 import { getMySQLInstance } from './db/mysql';
 
+// SQLs here is the last guards before inserting into earn table
 const bonus_condition_sql = {
   // Arguments: [challenge, client_id]
   // No need to check if user has already won the bonus because it must happend after a successful enrollment.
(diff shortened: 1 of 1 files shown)


rileyjshaw:
This looks great, thanks.

> `multiple invite_contribute_same_session` bonus can't be avoid with this PR;

Are you able to make another PR to solve this issue?

can-007:
> Are you able to make another PR to solve this issue?

Me and @catherine917 were thinking about add another piggy-backed value to help to solve this issue. The value could be a `shouldShowToast: bool `to indicate whether or not to process this toast in UI. Or it could be a db time `dbTime: Timestamp with Timezone` letting the front-end to decide what to do.

Do you have any better ideas?

rileyjshaw:
@can-007 OH, I see now. Only one achievement is ever actually stored, but the achievement notification shows repeatedly on the front-end. Is that correct?

Assuming my understanding is correct, I think sending an extra flag is okay. At some point, it might make sense to send down a list of all awarded achievements on the user account object (eg. if we ever wanted to make a /profile/awards page). But for now, an extra flag sounds fine to me.

PR #2385, opened 2019-11-19
calculate the team invites more precisely:

can-007:
1) signed up within the first week;
2) signed up without invited_by;
3) including the current user if it should;

Currently, the number of team invites doesn't include current user.
@rileyjshaw , I made the bold statement in the stand-up that the current code base is correct about the team invites.
I realized that I was wrong after rechecking.


rileyjshaw:
Thanks @can-007! I pushed a slight update to match the specific challenge requirements of Week 1.

PR #2392, opened 2019-11-20
Fix #2386: No points earned for first Listen contribution, until all 5 are submitted

### LINKED ISSUE #2386: No points earned for first Listen contribution, until all 5 are submitted
[STR]:
1.Sign-up for the Voice Challenge using an invite link
2.From https://voice.mozilla.org/en/dashboard/challenge, click the Listen button
3.Contribute with one clip (listening)

[Expected result]:
A message with 50 points earned for the first contribution should be displayed

[Actual result]:
No message displayed, no points earned

can-007:
show toast after the `async` `vote()` is returned, otherwise toast will never show up, this is done in the following steps:

1) extract the toast functionality into a separate `async `method of the same _tsx_ file,
2) use the `promise.all().then() `to make sure that the achievement attributes piggy-backed in vote request already update the `props `before showing toasts.

This is PR should resolve issue: https://github.com/mozilla/voice-web/issues/2386.

@rileyjshaw , @phirework  please take a look at it, and help to identify whether it would obviously break other functionalities.

### DIFF
--- a/web/src/components/pages/contribution/listen/listen.tsx
+++ b/web/src/components/pages/contribution/listen/listen.tsx
@@ -134,19 +134,18 @@ class ListenPage extends React.Component<Props, State> {
     trackListening('listen', this.props.locale);
   };
 
-  private vote = (isValid: boolean) => {
-    const { clips } = this.state;
+  private showAchievementToast = async () => {
+    if (!(this && this.props)) {
+      return;
+    }
     const {
-      firstContribute,
-      hasAchieved,
       addAchievement,
       api,
+      firstContribute,
       firstStreak,
+      hasAchieved,
     } = this.props;
-    const clipIndex = this.getClipIndex();
 
-    this.stop();
-    this.props.vote(isValid, this.state.clips[this.getClipIndex()].id);
     sessionStorage.setItem('hasContributed', 'true');
     if (firstContribute) {
       addAchievement(
@@ -173,6 +172,16 @@ class ListenPage extends React.Component<Props, State> {
       // Each user can only get once.
       api.setInviteContributeAchievement();
     }
+  };
+
+  private vote = (isValid: boolean) => {
+    const { clips } = this.state;
+    const clipIndex = this.getClipIndex();
+
+    this.stop();
+    Promise.all([
+      this.props.vote(isValid, this.state.clips[this.getClipIndex()].id),
+    ]).then(this.showAchievementToast);
     this.setState({
       hasPlayed: false,
       hasPlayedSome: false,
--- a/web/src/components/pages/contribution/listen/listen.tsx
+++ b/web/src/components/pages/contribution/listen/listen.tsx
@@ -134,18 +134,19 @@ class ListenPage extends React.Component<Props, State> {
     trackListening('listen', this.props.locale);
   };
 
-  private showAchievementToast = () => {
-    if (!(this && this.props)) {
-      return;
-    }
+  private vote = async (isValid: boolean) => {
+    const { clips } = this.state;
+    const clipIndex = this.getClipIndex();
+
+    this.stop();
+    await this.props.vote(isValid, this.state.clips[this.getClipIndex()].id);
     const {
+      firstContribute,
+      hasAchieved,
       addAchievement,
       api,
-      firstContribute,
       firstStreak,
-      hasAchieved,
     } = this.props;
-
     sessionStorage.setItem('hasContributed', 'true');
     if (firstContribute) {
       addAchievement(
@@ -172,16 +173,6 @@ class ListenPage extends React.Component<Props, State> {
       // Each user can only get once.
       api.setInviteContributeAchievement();
     }
-  };
-
-  private vote = (isValid: boolean) => {
-    const { clips } = this.state;
-    const clipIndex = this.getClipIndex();
-
-    this.stop();
-    Promise.all([
-      this.props.vote(isValid, this.state.clips[this.getClipIndex()].id),
-    ]).then(this.showAchievementToast);
     this.setState({
       hasPlayed: false,
       hasPlayedSome: false,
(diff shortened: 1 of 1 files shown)


web/src/components/pages/contribution/listen/listen.tsx
```diff
@@ -173,6 +172,16 @@ class ListenPage extends React.Component<Props, State> {
...
+  };
+
+  private vote = (isValid: boolean) => {
+    const { clips } = this.state;
+    const clipIndex = this.getClipIndex();
+
+    this.stop();
+    Promise.all([
+      this.props.vote(isValid, this.state.clips[this.getClipIndex()].id),
+    ]).then(this.showAchievementToast);
```
rileyjshaw: It looks like there's only one promise in here; is this equivalent to:

```
this.props.vote(isValid, this.state.clips[this.getClipIndex()].id).then(this.showAchievementToast);
```

?

web/src/components/pages/contribution/listen/listen.tsx
```diff
@@ -134,19 +134,18 @@ class ListenPage extends React.Component<Props, State> {
     trackListening('listen', this.props.locale);
   };
 
-  private vote = (isValid: boolean) => {
-    const { clips } = this.state;
+  private showAchievementToast = async () => {
```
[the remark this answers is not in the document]
rileyjshaw: ++

web/src/components/pages/contribution/listen/listen.tsx
can-007: Yes, only typescript will complain about strange syntax error. It treat the then() as somewhat a property.

web/src/components/pages/contribution/listen/listen.tsx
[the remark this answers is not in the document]
can-007: Good point, I will remove the async soon.

web/src/components/pages/contribution/listen/listen.tsx
rileyjshaw: I'm looking into this now; I see what you're referring to:

```
Property 'then' does not exist on type '(dispatch: Dispatch<LoadAction | RefillCacheAction | RemoveClipAction | AchievementAction | UpdateAction | TallyRecordingAction | TallyVerificationAction>, getState: () => StateTree) => Promise<...>'.
```

It seems that `this.props.vote` returns an async function, not a Promise. I think the `Promise.all()` might be obscuring an actual error.

can-007: feel free to change/refactor anything.
I don't know that much about React.

rileyjshaw: From https://react-redux.js.org/5.x/api/connect#docsNav:

> `[mapDispatchToProps(dispatch, [ownProps]): dispatchProps]` (Object or Function): If an object is passed, each function inside it is assumed to be a Redux action creator. An object with the same function names, **but with every action creator wrapped into a dispatch call so they may be invoked directly,** will be merged into the component’s props.

The types we're using for these action creators are:

```typescript
interface PropsFromDispatch {
  removeClip: typeof Clips.actions.remove;
  vote: typeof Clips.actions.vote;
  addAchievement: typeof Notifications.actions.addAchievement;
}
```

They seem to be referencing the _original_ functions, not the wrapped functions that can be invoked directly. That explains the error.

My suggestion would be to rewrite the types pf `PropsFromDispatch` as needed, eg:

```typescript
interface PropsFromDispatch {
  removeClip: (clipId: string) => Promise<any>;
  vote: (isValid: boolean, clipId?: string) => Promise<any>;
  addAchievement: (
    score: number,
    text: string,
    type: NotificationType
  ) => Promise<any>;
}
```

@Gregoor you've probably encountered this problem heaps before. Any guidance?

can-007: @rileyjshaw , the code can run, I think it's more of a programming style, a guideline thing. To rewrite the types of `PropsFromDispatch `, we need to change  many places, causing nonresistant interface style of different `PropsFromDispatch `.
Maybe get a coding guideline first, stating things like: prefer `Promise `to `async`. So that the contributors could know the direction.

Gregoor: I can't seem to recall how we handled async functions in the past. My expectation would've been that redux(-thunk) respects an action's async-ness when `connect`ing it. But I might be expecting something from those libs that's maybe not even possible to express in TS.
Anyway, when types stand in your way I also vote for what Riley said: override the types to make it fit (and reconsider things more fundamentally when it becomes a pattern).

can-007: OK, but how about the minimum modifications that I just pushed here: https://github.com/mozilla/voice-web/pull/2392/commits/fdf5d551c433ba2f7d3a98e4fe9615a339235ab5?

It is small change, it's working and it doesn't have introduce Promise here.

web/src/components/pages/contribution/listen/listen.tsx
```diff
@@ -134,19 +134,19 @@ class ListenPage extends React.Component<Props, State> {
...
+
+    this.stop();
+    await this.props.vote(isValid, this.state.clips[this.getClipIndex()].id);
     const {
       firstContribute,
       hasAchieved,
       addAchievement,
       api,
       firstStreak,
     } = this.props;
```
rileyjshaw: Let's move this back to where it was before so that the diff is simpler. As far as I can tell only two lines actually changed (adding async / await).

can-007: please don‘t. This line won’t change after this.props gets updated. For example, firstContribute will always be false.

It has to be where vote() gets called and returned.

can-007: Forgot to mention, this.props.vote() updates attributes such as firstContribute, hasAchieved...

I had subsequent commits to change these attributes’ names to something like showFirstContributeToast, hasEarnedxxxBonus etc. Also fix the bug about multiple same_session toasts AFTER challenge’s end date.
I will put them in another PR tomorrow.

rileyjshaw: Oh! Surprising - thanks for the extra info. With that extra information I'd say this is good to go.

PR #2395, opened 2019-11-20
Update locales

MichaelKohler:


MichaelKohler:
of course I forgot to update the locales with my last export.. at some point I'll get used to it -.-

cc @irvin @gion-andri 

@nukeador

PR #2396, opened 2019-11-20
Set default value for importing sentences

phirework:
Small quibble of a commit, I would argue that in local dev you're much more likely to turn this off than need it on and I would like the default to be `false` just so I don't keep accidentally maxing out my system resources :P


rileyjshaw:
Omg thank you.

PR #2400, opened 2019-11-20
Tweak enrollment flow for logged-in users

### LINKED ISSUE #2356: Clicking on enrollment link while logged in should not require a second login

phirework:
Fixes #2356. Added account check to `welcome-modal.tsx` that enrolls and redirects directly.

### DIFF
--- a/web/src/components/welcome-modal/welcome-modal.tsx
+++ b/web/src/components/welcome-modal/welcome-modal.tsx
@@ -5,6 +5,9 @@ import BalanceText from 'react-balance-text';
 import Modal, { ModalProps } from '../modal/modal';
 import { Button, Checkbox } from '../ui/ui';
 import { trackChallenge } from '../../services/tracker';
+import { useAccount, useAction } from '../../hooks/store-hooks';
+import { User } from '../../stores/user';
+import { Enrollment } from '../../../../common/challenge';
 import {
   ChallengeTeamToken,
   challengeTeams,
@@ -22,7 +25,29 @@ export interface WelcomeModalProps extends ModalProps {
 export default ({ challengeToken, teamToken, ...props }: WelcomeModalProps) => {
   const readableTeamName = challengeTeams[teamToken].readableName;
   const [hasAgreed, setHasAgreed] = useState<boolean>(false);
+  const account = useAccount();
+  const saveAccount = useAction(User.actions.saveAccount);
+
   useEffect(() => trackChallenge('modal-welcome'), []);
+
+  const parseEnrollment = (queryString: string): Enrollment => {
+    if (queryString[0] === '?') queryString = queryString.substring(1);
+    const queries = queryString.split('&').reduce(
+      (acc, curr) => {
+        const keyVal = curr.split('=');
+        if (keyVal.length > 0) acc[keyVal[0]] = keyVal[1];
+        return acc;
+      },
+      {} as { [key: string]: string }
+    );
+
+    return {
+      challenge: queries.challenge,
+      team: queries.team,
+      referer: queries.referer,
+      invite: queries.invite,
+    } as Enrollment;
+  };
 
   return (
     <Modal {...props} innerClassName="welcome-modal">
@@ -63,11 +88,19 @@ export default ({ challengeToken, teamToken, ...props }: WelcomeModalProps) => {
           const enrollmentDetails = window.location.search;
           // `enrollmentDetails` should always exist here, but in case it
           // doesn't we abort the login flow.
+
           if (enrollmentDetails) {
-            const { referrer } = document;
-            window.location.href = `/login${enrollmentDetails}${
-              referrer ? `&referer=${referrer}` : ''
-            }`;
+            if (account) {
+              const enrollObject = parseEnrollment(enrollmentDetails);
+              saveAccount({ enrollment: enrollObject }).then(() => {
+                window.location.href = `/dashboard/challenge?challenge=${enrollObject.challenge}&achievement=1`;
+              });
+            } else {
+              const { referrer } = document;
+              window.location.href = `/login${enrollmentDetails}${
+                referrer ? `&referer=${referrer}` : ''
+              }`;
+            }
           } else {
             window.location.reload();
           }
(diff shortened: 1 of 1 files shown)


web/src/components/welcome-modal/welcome-modal.tsx
```diff
@@ -22,8 +25,28 @@ export interface WelcomeModalProps extends ModalProps {
...
+      queries[pair[1]] = pair[2];
+    }
+
+    return {
+      challenge: queries.challenge,
+      team: queries.team,
+      referer: queries.referer,
+      invite: queries.invite,
+    } as Enrollment;
+  };
```
phirework: This whole method is messy and I don't love it, very much open to suggestions for improvements. Without using an external library for accessing query params, this was my best shot. Also, the casting at `return` seems pretty redundant but that was the best way I could think of to avoid a meaningless `:Object` return type.

web/src/components/welcome-modal/welcome-modal.tsx
```diff
@@ -31,15 +31,13 @@ export default ({ challengeToken, teamToken, ...props }: WelcomeModalProps) => {
...
-      },
-      {} as { [key: string]: string }
-    );
+    const regex = new RegExp(/([a-z]+)=([a-z]+)/, 'gm');
+    const queries = {} as { [key: string]: string };
+    let pair: Array<string> = [];
+
+    while ((pair = regex.exec(queryString)) !== null) {
+      queries[pair[1]] = pair[2];
+    }
```
phirework: I extremely do not understand why my brain did not jump to regex immediately; I shall blame it on fatigue 😅

phirework:
Merging into

phirework:
Merging into master so we can test on stage - please still feel free to provide comments!

web/src/components/welcome-modal/welcome-modal.tsx
```diff
@@ -63,11 +86,19 @@ export default ({ challengeToken, teamToken, ...props }: WelcomeModalProps) => {
...
           if (enrollmentDetails) {
-            const { referrer } = document;
-            window.location.href = `/login${enrollmentDetails}${
-              referrer ? `&referer=${referrer}` : ''
-            }`;
+            if (account) {
+              const enrollObject = parseEnrollment(enrollmentDetails);
+              saveAccount({ enrollment: enrollObject }).then(() => {
+                window.location.href = `/dashboard/challenge?challenge=${enrollObject.challenge}&achievement=1`;
+              });
```
rileyjshaw: I think we'll still want to save the `referer` here.

web/src/components/welcome-modal/welcome-modal.tsx
```diff
@@ -63,11 +86,19 @@ export default ({ challengeToken, teamToken, ...props }: WelcomeModalProps) => {
...
+
           if (enrollmentDetails) {
-            const { referrer } = document;
-            window.location.href = `/login${enrollmentDetails}${
-              referrer ? `&referer=${referrer}` : ''
-            }`;
+            if (account) {
+              const enrollObject = parseEnrollment(enrollmentDetails);
+              saveAccount({ enrollment: enrollObject }).then(() => {
+                window.location.href = `/dashboard/challenge?challenge=${enrollObject.challenge}&achievement=1`;
```
rileyjshaw: Can we handle this through a `<Redirect>` component instead?

web/src/components/welcome-modal/welcome-modal.tsx
```diff
@@ -22,8 +25,28 @@ export interface WelcomeModalProps extends ModalProps {
...
+    let pair: Array<string> = [];
+
+    while ((pair = regex.exec(queryString)) !== null) {
+      queries[pair[1]] = pair[2];
+    }
+
+    return {
+      challenge: queries.challenge,
+      team: queries.team,
+      referer: queries.referer,
```
rileyjshaw: `referer` is pulled from `document.referrer`, not from a query param.

web/src/components/welcome-modal/welcome-modal.tsx
```diff
@@ -22,8 +25,28 @@ export interface WelcomeModalProps extends ModalProps {
 export default ({ challengeToken, teamToken, ...props }: WelcomeModalProps) => {
   const readableTeamName = challengeTeams[teamToken].readableName;
   const [hasAgreed, setHasAgreed] = useState<boolean>(false);
+  const account = useAccount();
+  const saveAccount = useAction(User.actions.saveAccount);
+
   useEffect(() => trackChallenge('modal-welcome'), []);
 
+  const parseEnrollment = (queryString: string): Enrollment => {
+    const regex = new RegExp(/([a-z]+)=([a-z]+)/, 'gm');
```
rileyjshaw: I don't think `[a-z]+` will match our invite codes, which look like: `e3cdefb8-0af0-11ea-b89e-06dc45d2f642`.

There's a chance that invite codes (or team names) will contain capitals or other characters in the future, so perhaps `[^=]` is more durable?

Also re: `'gm'`, does this need to be multiline?

web/src/components/welcome-modal/welcome-modal.tsx
rileyjshaw: You could always use this piece of garbage we coded together last week 😂 https://gist.github.com/rileyjshaw/32d7455e63c129d6064d406870a427c2

...but I like this, it's clear. Check the comments above though, I think it's not working for invite codes currently.

Also re: the casting at `return`, if we wanted to do some safety checks we could do something like:

```
if (challengeTokens.includes(queries.challenge) && challengeTeamTokens.includes(queries.team)) {
    return queries as Enrollment;
} else return null;
```

phirework: The thing I was trying to guard against is if `challengeTokens` has invalid properties beyond the four that `Enrollment` defines for some reason - it's probably overly cautious and I think (I could be wrong?) it would only generate a warning, not an error, so this is probably fine for prod.

rileyjshaw: Ah, makes sense. In that case, it looks like we have access to a `pick` function:

```javascript
const pick = require('lodash.pick');
// ...
if (challengeTokens.includes(queries.challenge) && challengeTeamTokens.includes(queries.team)) {
    return pick(queries, 'challenge', 'team', 'invite') as Enrollment;
} else return null;
```

PR #2401, opened 2019-11-20
Remove iOS banner until we know what we're doing with the app

phirework:
Have deliberately left the CSS in case we need to restore this in the future, and have also tagged this as `tags/remove-iOS-banner` for easy referencing.


rileyjshaw:
Our first tag! Thanks, this looks good to me.

PR #2407, opened 2019-11-22
Better login redirect flow

phirework:
Based on PR comments on #2400 (thanks!!) 
* improved regex
* Moved redirect logic into separate method for readability 
* Correct referer value this time
* Incidentally solves (or...improves, I guess) the intermittent dashboard reload issue

Started a new PR instead of reopening that old one since it was already merged, didn't want to deal with conflicts or messiness.

### DIFF
--- a/web/src/components/welcome-modal/welcome-modal.tsx
+++ b/web/src/components/welcome-modal/welcome-modal.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import { useEffect, useState } from 'react';
-import { Link } from 'react-router-dom';
 import { Redirect } from 'react-router';
+import { Link } from 'react-router-dom';
+import { pick } from 'lodash';
 import BalanceText from 'react-balance-text';
 import Modal, { ModalProps } from '../modal/modal';
 import { Button, Checkbox } from '../ui/ui';
@@ -18,7 +19,6 @@ import {
   challengeTeamTokens,
 } from 'common/challenge';
 import URLS from '../../urls';
-const pick = require('lodash.pick');
 
 import './welcome-modal.css';
 
@@ -33,6 +33,7 @@ export default ({ challengeToken, teamToken, ...props }: WelcomeModalProps) => {
   const account = useAccount();
   const saveAccount = useAction(User.actions.saveAccount);
   const [locale, toLocaleRoute] = useLocale();
+  const [redirectChallenge, setRedirectChallenge] = useState();
 
   useEffect(() => trackChallenge('modal-welcome'), []);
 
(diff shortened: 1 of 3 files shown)


web/src/components/welcome-modal/welcome-modal.tsx
```diff
@@ -30,21 +33,44 @@ export default ({ challengeToken, teamToken, ...props }: WelcomeModalProps) => {
 
   useEffect(() => trackChallenge('modal-welcome'), []);
 
-  const parseEnrollment = (queryString: string): Enrollment => {
-    const regex = new RegExp(/([a-z]+)=([a-z]+)/, 'gm');
+  const parseEnrollment = (
+    queryString: string,
+    referer?: string
+  ): Enrollment => {
+    const regex = new RegExp(/([\w\-]+)=([\w\-]+)/, 'g');
```
phirework: The `m` was a leftover value from when I was testing the regex so it's been taken out. This pattern works for something like `?key_name=value-PROPERTY-123` which I think covers all of our current use cases.

web/src/components/welcome-modal/welcome-modal.tsx
```diff
@@ -30,21 +33,44 @@ export default ({ challengeToken, teamToken, ...props }: WelcomeModalProps) => {
...
+
+    if (enrollmentDetails) {
+      if (account) {
+        const enrollObject = parseEnrollment(enrollmentDetails, referrer);
+
+        saveAccount({ enrollment: enrollObject }).then(() => {
+          window.location.href = `/dashboard/challenge?challenge=${enrollObject.challenge}&achievement=1${referrerString}`;
+        });
+      } else {
+        window.location.href = `/login${enrollmentDetails}${referrerString}`;
```
phirework: I tried a bunch of different ways to get these two working with `<Redirect />` and while I could get logged-in users to redirect, I could not get the logged out version to send people to `/login` for the life of me. I'm fine with this the way it is, but if you know how to make `<Redirect />` behave I'm all ears!!

rileyjshaw: Oops, I should have been more specific! I don't think `/login` is handled by React Router, so it needs a redirect as above. But logged-in users should be able to use `<Redirect />` (as you noticed), since dashboard routes are handled on the front-end. I think we should use `<Redirect />` wherever we can.

---

We _could_ handle external routes within our router by following a [pattern like this](https://jessewaites.com/Redirecting-To-An-External-Link-In-A-React-Application), but it's a matter of taste at that point.

phirework: Very relieved that I wasn't just missing something totally glaring about how `<Redirect />` works 😅

phirework:
Added an extra quick fix to address #2416, since it felt semantically related.

web/src/components/welcome-modal/welcome-modal.tsx
```diff
@@ -27,24 +32,52 @@ export default ({ challengeToken, teamToken, ...props }: WelcomeModalProps) => {
...
+    if (enrollmentDetails) {
+      if (account) {
+        const enrollObject = parseEnrollment(enrollmentDetails, referrer);
+
+        saveAccount({ enrollment: enrollObject }).then(() => {
+          const redirect = {
+            pathname: toLocaleRoute(URLS.DASHBOARD + '/' + URLS.CHALLENGE),
+            search: `?challenge=${enrollObject.challenge}&achievement=1${referrerString}`,
+          };
+          return <Redirect to="{redirect}" />;
```
rileyjshaw: I don't think returning `<Redirect...` from here will actually render it, which may cause a problem with various React lifecycle methods. I think we'll want to replace this code block with a `setChallengeRedirect(true)` call, and handle that state directly in the render method.

rileyjshaw: The quotes here mean we're redirecting to the literal `./{redirect}`.

Also: I think we decided last week to start adding the `push` attribute to these. If you think it's best to not use it in this case I probably agree, just wanted to call it out in case.

phirework:
Thank YOU @rileyjshaw for the detailed reviews! Super appreciate your attention to detail 😊

PR #2408, opened 2019-11-22
Remove emails from config-helper.ts, add kibana_admins to confd template

phirework:
Have already updated the `ADMIN_EMAILS` and `KIBANA_ADMINS` values via consul on both prod and stage, this just removes the PII and relies entirely on that KV store.


phirework:
Heads up to @The-smooth-operator for future Kibana admin purposes.

PR #2414, opened 2019-11-23
Add condition for checking file type

cantasci:
Throws an error on web when the user has canceled the image file selection. After that, the user can't select a new avatar again. I've added a condition for fixing this problem.

Scenario:
- Open the avatar page.
- Click the browse button. 
- Select an image file.
- Click the browse button, again.
- Press ESC or click the cancel.


rileyjshaw:
Looks great, thanks for the contribution!

PR #2415, opened 2019-11-24
[web] UI: Fix alignment of registration section

Kriyszig:
Registration section aligned left for niche screen size especially for devices like Kindle Fire HDX and certain other tablets

#### Previously:
Imgur
Imgur

#### After the changes:
Imgur
Imgur

Currently there are no issues opened related t this. I thought it would be good to have consistent styling throughout for even the niche screen sizes.
In case this is an unnecessary change, please let me know and I'll close the PR. In case the changes need refinement, drop a comment below and I'll address them asap

### DIFF
--- a/web/src/components/register-section/register-section.css
+++ b/web/src/components/register-section/register-section.css
@@ -13,6 +13,7 @@
 
         @media (--md-down) {
             flex-direction: column;
+            align-items: center;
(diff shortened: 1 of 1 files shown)


phirework:
This looks great, thanks!

PR #2424, opened 2019-11-25
Alter DB schema to cascade client_id deletions

phirework:
In preparation for deleting a whole bunch of users and automating this process in the future, this adds `ON DELETE CASCADE` constraints to all tables that refer to `user_clients.client_id` as a foreign key, as well as any other foreign keys that may be affected by a client deletion (for example, `awards.custom_goal_id` now also cascade on `custom_goals.id` deletion). 

On `clips` I just removed the foreign key constraint entirely, because we give users the option of removing their profiles while retaining their clips. If we are keeping their clips on S3 I thought it made the most sense for those clips to also still remain in the DB. Let me know if you have any thoughts otherwise.


rileyjshaw:
Awesome! I'm really glad we're removing some of the manual friction from account deletions. This looks good to me, but I'll wait for @Gregoor to sign off on it in case there's something of historical significance that I'm missing.

(Also, probably a stupid question but all this `DROP`ing  isn't going to delete anything important, right?)

phirework:
Thanks you two! And @rileyjshaw nope deleting the constraint doesn't touch the data in the column, thankfully 😅

PR #2406, opened 2019-11-22
Fix for Italian original sentence dataset

Mte90:
@Sav22999 waiting for a check by him

I opened the pr in the wrong repo (we usually do before on our fork) but anyway the changes are very few, based on https://github.com/mozillaitalia/voice-web/issues/35

PR #2428, opened 2019-11-26
Removed sentences from italian after #2406

Mte90:
I removed all the sentences that we fixed in #2406 so this can be approved.


nukeador:
This PR still contains edits, not only removals.

 +5 −63

Mte90:
Changes done, seems that the last line seems a new one but probably because I removed the last empty line.

nukeador:
Thanks!

PR #2437, opened 2019-11-28
prevent multiple invite_contribute_same_session AFTER challenge

can-007:
- prevent multiple `invite_contribute_same_session ` toasts AFTER challenge by checking challenge time under different timezone;
- change `firstInvite `to `showInviteSendToast`;
- change `hasAchieved `to `hasEarnedSessionToast`;
- change `firstContribute `to `showFirstContributionToast`;
- change `firstStreak `to `showFirstStreakToast`;
- set the default value of `challengeEnded `to true;

@rileyjshaw , as mentioned in PR: https://github.com/mozilla/voice-web/pull/2382, this PR will not show any `invite_contribute_same_session `bonus once challenge has ended.

### DIFF
--- a/server/src/lib/model/db.ts
+++ b/server/src/lib/model/db.ts
@@ -690,6 +690,8 @@ export default class DB {
       [challenge]
     );
     if (row) {
+      // row.start_date_utc is utc time (timezone offset is 0);
+      // startDateServer = start_date_utc + backend timezone offset (in minutes);
       const startDateServer = new Date(row.start_date_utc);
       const startDateUtc = new Date(
         startDateServer.valueOf() -
--- a/server/src/lib/model/db.ts
+++ b/server/src/lib/model/db.ts
@@ -4,6 +4,7 @@ import Mysql, { getMySQLInstance } from './db/mysql';
 import Schema from './db/schema';
 import ClipTable, { DBClipWithVoters } from './db/tables/clip-table';
 import VoteTable from './db/tables/vote-table';
+import { ChallengeToken } from 'common/challenge';
 
 // When getting new sentences/clips we need to fetch a larger pool and shuffle it to make it less
 // likely that different users requesting at the same time get the same data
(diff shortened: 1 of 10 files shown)


web/src/components/pages/contribution/listen/listen.tsx
```diff
@@ -140,29 +141,37 @@ class ListenPage extends React.Component<Props, State> {
 
     this.stop();
     await this.props.vote(isValid, this.state.clips[this.getClipIndex()].id);
+    // this.props.vote() will update props values like challengeEnded, showFirstStreakBonus etc.
+    // So the following line must be run AFTER this.props.vote() has finished synchronously.
```
rileyjshaw: Thanks for adding the comment!

server/src/lib/model/db.ts
```diff
@@ -678,4 +679,28 @@ export default class DB {
...
+      // row.start_date_utc is utc time (timezone offset is 0);
+      // startDateServer = start_date_utc + backend timezone offset (in minutes);
+      const startDateServer = new Date(row.start_date_utc);
+      const startDateUtc = new Date(
+        startDateServer.valueOf() -
+          startDateServer.getTimezoneOffset() * 60 * 1000
+      );
+      hasChallengeEnded =
+        new Date().valueOf() >
+        startDateUtc.valueOf() + 21 * 24 * 60 * 60 * 1000;
```
rileyjshaw: I think this would be clearer (and more efficient) if we put the line:

```
const THREE_WEEKS = 3 * 7 * 24 * 60 * 60 * 1000;
```

at the top of the file, then used that constant here.

server/src/lib/model/db.ts
```diff
@@ -678,4 +679,28 @@ export default class DB {
     );
     return row;
   }
+
+  async hasChallengeEnded(challenge: ChallengeToken) {
+    let hasChallengeEnded = true;
```
rileyjshaw: We probably should name these different things. Maybe the method could be called `getHasChallengeEnded()`?

server/src/lib/model/db.ts
```diff
@@ -678,4 +679,28 @@ export default class DB {
...
+    if (row) {
+      // row.start_date_utc is utc time (timezone offset is 0);
+      // startDateServer = start_date_utc + backend timezone offset (in minutes);
+      const startDateServer = new Date(row.start_date_utc);
+      const startDateUtc = new Date(
+        startDateServer.valueOf() -
+          startDateServer.getTimezoneOffset() * 60 * 1000
+      );
+      hasChallengeEnded =
+        new Date().valueOf() >
```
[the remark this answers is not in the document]
can-007: Yes, why I didn't come up with this!

server/src/lib/model/db.ts
```diff
@@ -678,4 +679,28 @@ export default class DB {
...
+      [challenge]
+    );
+    if (row) {
+      // row.start_date_utc is utc time (timezone offset is 0);
+      // startDateServer = start_date_utc + backend timezone offset (in minutes);
+      const startDateServer = new Date(row.start_date_utc);
+      const startDateUtc = new Date(
+        startDateServer.valueOf() -
+          startDateServer.getTimezoneOffset() * 60 * 1000
+      );
```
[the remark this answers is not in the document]
can-007: Nice!

server/src/lib/model/db.ts
can-007: OK

can-007: maybe the variable could be called challengeEnded.

PR #2438, opened 2019-11-29
refactor hasChallengeEnded:

can-007:
- use Date.now() instead of new Date().valueOf() to get millisecond of now;
- use `time zone designator`, plus 'Z' in the end, to initialize date to UTC time;
- extract millisecond computation of 3 weeks into one constant;
- local variable hasChallengeEnded is the same as the function name, change to challengeEnded;

@rileyjshaw , really good comments there ...


rileyjshaw:
Awesome, thank you for making these changes!

PR #2441, opened 2019-11-29
Upgrade TypeScript

rileyjshaw:
This commit ensures that we're on the latest version of TypeScript. More
importantly, it ensures we're using the same version of TypeScript across
the application.

Also… now we can finally use [Optional Chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)! 🎉

Test plan:

- `yarn install`
- `rm -rf server/js web/dist`
- `yarn build`
- `yarn prettier`
- `yarn test`

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "prettier": "^1.18.2",
     "request": "^2.88.0",
     "request-promise-native": "^1.0.7",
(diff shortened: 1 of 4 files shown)


phirework:
HURRAY

PR #2445, opened 2019-12-02
Node script for profile deletion

phirework:
A pretty rough and basic node command line script for deleting user profiles. 

Expected behaviour: 

* Trigger by calling `node scripts/scrub-user.js [1+ email address, separated by spaces]` from root directory
* For each email address, command line will: 
  * Display a summary of the user's stats, with number of `client_id`s, number of clips, and date the first clip was submitted
  * Prompt for confirmation that this user profile should be deleted
  * If the user submitted clips, prompt for confirmation that the records for those clips should also be deleted
    * Add those `client_id`s to `deletionSummary` object for display later
    * If the first contribution from that user has been included in a past dataset, add those `client_ids` to `deletionSummary` object for display later
* Once all emails have been processed, print `deletionSummary `object with the `client_id`s to remove from S3, and any `client_id`s that we need to notify past dataset downloaders about

Right now this script doesn't automate deletion from S3, mostly because I don't have a personal instance set up with test files yet and I can't connect to the staging S3 locally. That'll be in the next iteration. I considered storing the deletion summary somewhere, but that seems like it goes against the spirit of removing user data on request, but I'm open to being persuaded otherwise.

### DIFF
--- a/scripts/scrub-user.js
+++ b/scripts/scrub-user.js
@@ -7,64 +7,72 @@ const readline = require('readline').createInterface({
 });
 
 let safeToDelete = true;
-let deletionSummary = {
+const deletionSummary = {
   emails: [],
-  s3_ids: [],
-  dataset_ids: [],
+  s3Ids: [],
+  datasetIds: [],
+};
+
+const isValidDate = dateStr => {
+  const dateObj = new Date(dateStr);
+  return dateObj instanceof Date && !isNaN(dateObj.getTime());
 };
 
 const parseResults = (results, email) => {
-  const LAST_DATASET = config.LAST_DATASET || new Date();
-  const clip_total = results.reduce((acc, curr) => acc + curr.clip_count, 0);
-  const client_ids = results.map(each => each.client_id);
+  const lastDatasetDate = isValidDate(config.LAST_DATASET_DATE)
+    ? new Date(config.LAST_DATASET_DATE)
+    : new Date();
+  const clipTotal = results.reduce((acc, curr) => acc + curr.clip_count, 0);
+  const clientIds = results.map(each => each.client_id);
 
   const summary = {
-    email: email,
+    email,
     length: results.length,
-    ids: client_ids,
-    clip_total: clip_total,
-    first_clip: null,
-    in_past_dataset: false,
+    ids: clientIds,
+    clipTotal: clipTotal,
+    firstClip: results[0] ? results[0].first_clip : null,
+    inPastDataset: results[0] && lastDatasetDate >= results[0].first_clip,
   };
-
-  if (summary.clip_total) {
-    summary.first_clip = results[0].first_clip;
-
-    if (LAST_DATASET >= new Date(summary.first_clip))
-      summary.in_past_dataset = true;
-  }
 
   return summary;
 };
 
-const processIteration = last => {
+const readlineLoop = (prompt, options) => {
+  readline.question(prompt, answer => {
+    const callback = options[answer.toLowerCase()];
+    if (callback) callback();
+    else readlineLoop(prompt, options);
+  });
+};
+
+const processIteration = isLast => {
   console.log(`=======================================`);
   safeToDelete = true;
 
-  if (last) {
-    printSummary(deletionSummary);
+  if (isLast) {
+    printSummary();
     readline.close();
     process.exit(1);
   }
 };
 
-const printSummary = summary => {
+const printSummary = () => {
   console.log('Deletion summary: \n');
   if (deletionSummary.emails.length)
     console.log(
       `Removed user information related to ${deletionSummary.emails.join(', ')}`
     );
 
-  if (deletionSummary.s3_ids.length)
+  if (deletionSummary.s3Ids.length)
     console.log(
-      `Please manually remove the following folders from S3: ${deletionSummary.s3_ids.join(
+      `Please manually remove the following folders from S3: ${deletionSummary.s3Ids.join(
         ', '
       )}.`
     );
 
-  if (deletionSummary.dataset_ids.length)
+  if (deletionSummary.datasetIds.length)
     console.log(
-      `The following client_ids were included in a past dataset: ${deletionSummary.dataset_ids.join(
+      `The following client_ids were included in a past dataset: ${deletionSummary.datasetIds.join(
         ', '
       )}.`
     );
@@ -74,8 +82,8 @@ const printUserStats = stats => {
   console.log(`
     Email: ${stats.email}
     Associated client_ids: ${stats.length}
-    Total clips contributed: ${stats.clip_total}
-    First contribution: ${stats.first_clip}
+    Total clips contributed: ${stats.clipTotal}
+    First contribution: ${stats.firstClip}
     `);
 };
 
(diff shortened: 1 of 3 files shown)


scripts/scrub-user.js
```diff
@@ -0,0 +1,195 @@
...
+            `There are no user accounts associated with the email address ${clean_email}.`
+          );
+          processIteration(isLast);
+        }
+      }
+    );
+  } else {
+    setTimeout(() => {
+      deleteUser(connection, email, isLast);
+    }, 200);
```
phirework: I don't love this piece, but without this the script would print all user stats at once, and accept all prompts at once, which still works but is a real sucky experience prone to error. Super open to hearing a different way of handling this! 

Alternatively, could make this script handle only one email address at a time, but that's pretty inefficient.

scripts/scrub-user.js
```diff
@@ -0,0 +1,195 @@
...
+    summary.first_clip = results[0].first_clip;
+
+    if (LAST_DATASET >= new Date(summary.first_clip))
+      summary.in_past_dataset = true;
+  }
+
+  return summary;
+};
+
+const processIteration = last => {
```
rileyjshaw: nitpick: this is called `isLast` everywhere else.

scripts/scrub-user.js
```diff
@@ -0,0 +1,195 @@
...
+
+      deletionSummary.emails.push(userStats.email);
+
+      if (userStats.clip_total) {
+        readline.question(
+          `\nWould you also like to delete the ${userStats.clip_total} associated clips? \nThis CANNOT be reversed [Y/N] `,
+          answer => {
+            if (answer.toLowerCase() === 'y')
+              deleteClipRecords(connection, userStats, isLast);
+            else processIteration(isLast);
```
rileyjshaw: If you wanted to account for typos, you could create a simple loop like:

```javascript
function readlineLoop (prompt, options) {
  readline.question(prompt, answer => {
    const callback = options[answer.toLowerCase()];
    if (callback) callback();
    else readlineLoop(prompt, options);
  });
}
```

then call it with:

```javascript
readlineLoop(`\nWould you also like to delete the ${userStats.clip_total} associated clips? \nThis CANNOT be reversed [Y/N] `, {
  y: () => deleteClipRecords(connection, userStats, isLast),
  n: () => processIteration(isLast),
});
```

scripts/scrub-user.js
rileyjshaw: > Alternatively, could make this script handle only one email address at a time, but that's pretty inefficient.

Is that not what's happening already? `emails.map` (or `emails.forEach`, as recommended below) will process each email sequentially. When the first email runs through `deleteUser`, it will mutate `safeToDelete` and block all other emails until it calls `processIteration` and releases the lock. Soon after, one of the `setTimeout`s will resolve, lock the `safeToDelete` variable, and start the process over again.

To me, it seems like something like:

```
for (const [i, email] of emails.entries()) {
  await deleteUser(connection, email, i === emails.length - 1);
}
```

should clear things up. Note, this will involve converting the functions above to `async`.

scripts/scrub-user.js
```diff
@@ -0,0 +1,195 @@
+const mysql = require('mysql');
+const config = require('../config.json');
```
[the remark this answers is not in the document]
phirework: It IS, but if it didn't exist the rest of the site also wouldn't function (well the homepage would but you wouldn't be able to actually do anything). Having said that though, after the move to Kubernetes we'll be using env variables rather than config.json for most things so I'm going to refactor this pre-emptively anyway.

scripts/scrub-user.js
```diff
@@ -0,0 +1,195 @@
...
+
+const readline = require('readline').createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+let safeToDelete = true;
+let deletionSummary = {
+  emails: [],
+  s3_ids: [],
```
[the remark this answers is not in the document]
phirework: Nope, just PHP brain rearing its ugly head. Gonna change all the ones that aren't set in a SQL query. Thanks for the flag!

scripts/scrub-user.js
```diff
@@ -0,0 +1,195 @@
...
+
+const deleteClipRecords = (connection, userStats, isLast) => {
+  const idStrings = userStats.ids.map(id => `"${id}"`);
+  connection.query(
+    `DELETE FROM clips WHERE client_id IN (${idStrings})`,
+    (clip_error, clip_results, clip_fields) => {
+      if (clip_error) throw clip_error;
+
+      if (clip_results.affectedRows) {
+        deletionSummary.s3_ids = deletionSummary.s3_ids.concat(userStats.ids);
```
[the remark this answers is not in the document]
phirework: No, it's actually possible for one email to have multiple client_ids associated with it, from before email addresses were used as unique account identifiers, so `userStats.ids` is an array and not a value. I'm not sure whether those redundant client_ids have any activities or clips associated with it but it seems safer to treat them like valid client_ids.

scripts/scrub-user.js
phirework: Oooh this is neat and elegant, thank you!

scripts/scrub-user.js
phirework: Sorry I meant "one at a time" as in "you'd have to keep typing `node script email`, instead of being able to do `node script emailA emailB emailC`". I tried implementing this with  async/await, but async/await is [pretty messy](https://zellwk.com/blog/async-await-in-loops/) inside loops. Couldn't get it to work in a `for...of` loop (as in it would print all `readline` prompts at once), it definitely isn't expected to work in a `forEach` loop with a callback, and with asyncs inside `map` I'd need to treat it as a `Promises.all`. Overall I did feel like `setTimeout` was cleaner and easier to follow than turning all callbacks up the chain into `asyncs` and then dealing with the Promise array. Totally open to being persuaded otherwise though.

scripts/scrub-user.js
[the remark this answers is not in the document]
rileyjshaw: Ohh, got it. Thanks for clarifying! `concat` is so weird like that 😓

A self-documenting alternative would be:

```
deletionSummary.s3_ids.push(...userStats.ids);
```

...but it's just a matter of style, I'm no-longer concerned about functionality here 👍

scripts/scrub-user.js
rileyjshaw: > I meant "one at a time" as in "you'd have to keep typing `node script email`

Ah okay, got it.

> I tried implementing this with async/await, but async/await is pretty messy

[Yeah, that's the classic thing with asynchronous code](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/). Once you do one, you've gotta do 'em all…

> Couldn't get it to work in a for...of loop (as in it would print all readline prompts at once)

It sounds like part of the call chain wasn't converted to `async`, so the `await` was resolving immediately. `for...of` _should_ be able to work smoothly with `async`. However,

> it definitely isn't expected to work in a `forEach` loop with a callback, and with asyncs inside `map` I'd need to treat it as a `Promises.all`.

Yep, it definitely won't work in a `forEach`. But for the same reason, it won't work in `map` either! The separate calls from `map` don't have an easy way to communicate, so there's no (good) way to say "wait for the previous iteration". `Promise.all` helps if all the calls can run at once, but the next step needs to wait until everything has resolved. Here, we need to schedule one after another – hence `for...of`.

> Overall I did feel like `setTimeout` was cleaner and easier to follow than turning all callbacks up the chain into `asyncs`

To each their own! If you want to pair on this sometime this week, it could be a fun chance for us to work together 😊 this seems to work as-is, though, so I don't wanna block you any longer.

Thanks for your patience and for the discussion! Clicking the big green button in 3... 2... 1...

PR #2452, opened 2019-12-05
async ALL the things

phirework:
Thanks again for all your help 🙏

### DIFF
--- a/scripts/scrub-user.js
+++ b/scripts/scrub-user.js
@@ -1,5 +1,6 @@
 require('dotenv').config();
 const mysql = require('mysql');
+const { promisify } = require('util');
 
 const dbConfig = {
   host: process.env.MYSQLHOST || 'localhost',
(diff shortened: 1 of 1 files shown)


scripts/scrub-user.js
```diff
@@ -29,32 +35,27 @@ const parseResults = (results, email) => {
...
-    const callback = options[answer.toLowerCase()];
-    if (callback) callback();
-    else readlineLoop(prompt, options);
-  });
+const readlineLoop = async (prompt, options) => {
+  const answer = await promptAsync(prompt);
+  const callback = options[answer.toLowerCase()];
+
+  if (callback) await callback();
+  else await readlineLoop(prompt, options);
```
rileyjshaw: Not important, but I wonder if this would be cleaner if we returned the promise instead:

```
// untested, might break something 🙃
return callback ? callback() : readlineLoop(prompt, options);
```

scripts/scrub-user.js
```diff
@@ -87,115 +88,113 @@ const printSummary = () => {
...
-          );
-          processIteration(isLast);
-        }
-      }
+      `
     );
-  } else {
-    setTimeout(() => {
-      deleteUser(connection, email, isLast);
-    }, 200);
```
rileyjshaw: yesssss

Issue #2459, opened 2019-12-07
Clicking yes or no does not do anything: i.user.account is null


rugk:
## STRs

1. Go to https://voice.mozilla.org/de/listen
2. Click "Yes" (or "No")

What happens: Nothing.

What should happen: Show me next sentence.

Console:
```
16:05:51.435 Content Security Policy: Die Einstellungen der Seite haben das Laden einer Ressource auf inline blockiert ("script-src"). 2 utils.js:35:9
16:05:51.448 do not track header set, disabling analytics listen:65:17
16:05:51.457 Content Security Policy: Die Einstellungen der Seite haben das Laden einer Ressource auf inline blockiert ("script-src"). listen:107:1
16:05:51.512 Intl.PluralRules already exists, and has NOT been replaced by this polyfill polyfill.js:92:12
16:05:51.512 To force, set a global ClobberIntlPluralRules = true polyfill.js:93:12
16:05:51.623 Error: Cannot load extension settings keepassxc-browser.js:828:17
16:06:04.582 TypeError: i.user.account is null
clips.ts:127:10
16:06:13.296 Error: Cannot load extension settings
```

## Additional information

When I once reloaded the site in this state, I got an empty white page "an error occured", and could create a Senetry error report.

Firefox 71

popey:
Confirmed broken here in chrome and firefox 70.

dabinat:
I’m seeing similar issues on my iPhone (iOS 13.2.3). Instead of being instant, tapping Yes/No takes about 2 seconds for the next sentence to appear. And sometimes it does nothing at all no matter how many times I tap the buttons.

Also, sometimes tapping Play cuts off the recording because it appears to interpret it as two taps.

phirework:
Thanks for the report, folks. Can you still reproduce the issue?

popey:
@phirework all good new in Firefox 70. Thanks.

dabinat:
@phirework This issue appears to have returned. The buttons are not doing anything again.

phirework:
@dabinat Thanks for the flag! Trying to track it down in our error logs, were you voting on french clips by any chance?

dabinat:
@phirework No, I only validate English. If it helps track it down, I’m on an iPhone 8 running iOS 13.2.3, using Firefox as my browser and my IP is located in Los Angeles, California. I was logged into my account at the time (dabinat).

phirework:
Awesome, thanks for the extra detail.

[...]

PR #2460, opened 2019-12-08
Revert PR #2225

phirework:
This addresses a new spike in Sentry error logs after the deployment to Prod on Dec 6th, and I think is related to bug #2459: 
* https://sentry.io/organizations/mozilla-q6/issues/1376071679/?query=is%3Aunresolved&sort=freq&statsPeriod=7d
* https://sentry.io/organizations/mozilla-q6/issues/1376100317/?query=is%3Aunresolved&sort=freq&statsPeriod=7d

Looking through the history of `speak.tsx` it looks like the `onReset` and `onSubmit` callbacks were intentionally passed as functions rather than references. It would be worth revisiting the purpose of the original PR to avoid re-rendering the contribute button unnecessarily, but as there is a December snippet campaign launching midnight UTC of Dec 10 I wanted to make sure core site functions were working first.

### DIFF
--- a/web/src/components/pages/contribution/speak/speak.tsx
+++ b/web/src/components/pages/contribution/speak/speak.tsx
@@ -597,9 +597,9 @@ class SpeakPage extends React.Component<Props, State> {
           isFirstSubmit={user.recordTally === 0}
           isPlaying={this.isRecording}
           isSubmitted={isSubmitted}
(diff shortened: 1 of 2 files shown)


rileyjshaw:
Awesome, thanks for noticing and fixing this!

PR #2467, opened 2019-12-10
removed incorect/foreign sentences

stefangrotz:
Hi, these are some incorrect sentences I found in the past month.

### DIFF
--- a/server/data/eo/sentence-collector.txt
+++ b/server/data/eo/sentence-collector.txt
@@ -457,8 +457,6 @@ Atentu Teo, vi malsekigos viajn ŝtrumpojn.
 Atinginte la jam senhoman ferdekon, Leonardo sidiĝis en ŝirmita angulo sur salikan seĝegon.
 Atinginte la supron de la altaĵo, la kunularo haltis momenton.
 Aĉjo, mi akiris novan ĉevalon, rigardu!
-Aŭbyn ekzistas malamikaj rilatoj jam de la tempo de la krucmilitiroj, kiel oni diras.
-Aŭbyn reciproke forte enamiĝis, furioze.
 Aŭdinte tion, la Muso turnis sin malrapide kaj renaĝis al ŝi.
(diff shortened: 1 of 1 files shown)


MichaelKohler:
Can you give me some examples on why these are wrong?

Then I would remove those from the sentence collector, and then you could add the corrected sentences (if it makes sense) through https://common-voice.github.io/sentence-collector. As context: if we do only remove it here, it will be added again with the next export, so we need to remove it in the Sentence Collector database first.

stefangrotz:
These sentences use "Si" instead of "Ŝi"

> Si kapjesis kaj rigardis min malĝoje.
> Si ne ekhavos mian ĉevalon, li diris arogante.
> Si pensas pri tiu misteraĵo.

These sentences contain non-Esperanto words, mostly spanish:

> José treege fieros pri tia honoro.
> Tion mi treege bedaŭras, señor, ĉar vi nun devos pacienci ĝis morgaŭ matene.


These sentences contain letters that are not part of the esperanto Alphabet: (y)

> Aŭbyn ekzistas malamikaj rilatoj jam de la tempo de la krucmilitiroj, kiel oni diras.
> Aŭbyn reciproke forte enamiĝis, furioze.

This one is gramatically wrong (it is a sentences said by Friday from robinson cruso)

> Jes, jes, vi iri en du boatoj.

For the future: Can everyone remove them from the sentence collector or can only you do it?

MichaelKohler:
Thanks for the explanation. I'll remove those within the next few days.

For now it's me who runs the deletions, to keep accountability in check.

MichaelKohler:
I've removed the sentences from Sentence Collector and verified that these deletions will be part of the next export. That will probably happen next Sunday.

@stefangrotz thanks for filing this, if you want to, you could upload corrected versions of those sentences at https://common-voice.github.io/sentence-collector/# :)

@nukeador this PR can be closed

PR #2472, opened 2019-12-12
Revert #2386 to see if that fixes 502 gateway issues

phirework:


server/src/lib/lazy-cache.ts
```diff
@@ -60,7 +60,7 @@ function redisCache<T, S>(
     return new Promise(async resolve => {
       const lock = await redlock.lock(
         key + '-lock',
-        1000 * 60 * 60 * 3 /*3 minutes*/
+        1000 * 60 * 3 /*3 minutes*/
```
rileyjshaw: 😓 jeez

PR #2473, opened 2019-12-12
Corrections of Sentence Collector sentences to replace all weird occu…

MichaelKohler:
…rrences

Fix for https://github.com/Common-Voice/sentence-collector/issues/269 ..


MichaelKohler:
I'll need to ask a few question to some people before we come say this is how it should look like.

PR #2474, opened 2019-12-12
WIP: Alter redisCache to refresh in background

phirework:
This won't do anything about the fact that the SQL queries themselves are super slow, but it should improve the user experience and minimize the times when the leaderboard appears as empty.

Basically my theory is that if the user hits a leaderboard page and the cache is expired, it's better to show them the cached result anyway and then renew the cache in the background so that the next user gets the fresh one, instead of making the user wait forever. Super WIP though, so very much open to thoughts/discussions.

### DIFF
--- a/server/src/lib/lazy-cache.ts
+++ b/server/src/lib/lazy-cache.ts
@@ -45,8 +45,6 @@ function renewRedisCache<T, S>(
 ): Fn<T, S> {
   return async (...args) => {
     return new Promise(async resolve => {
-      let renewCache = true;
-
       const lock = await redlock.lock(
         key + '-lock',
         1000 * 60 * 3 /*3 minutes*/
@@ -71,14 +69,12 @@ function redisCache<T, S>(
     const result = await redis.get(key);
 
     let value: any;
-    let renewCache = true;
+
     if (result) {
       const cached = JSON.parse(result);
-      value = cached.value;
-      renewCache = isExpired(cached.at, timeMs);
-
-      if (renewCache) renewRedisCache(key, f, timeMs)(...args);
-      return value;
+      if (isExpired(cached.at, timeMs))
+        renewRedisCache(key, f, timeMs)(...args);
+      return cached.value;
     }
 
     return new Promise(async resolve => {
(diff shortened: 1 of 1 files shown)


server/src/lib/lazy-cache.ts
```diff
@@ -38,6 +38,29 @@ function memoryCache<T, S>(f: Fn<T, S>, timeMs: number): Fn<T, S> {
...
 }
 
+function renewRedisCache<T, S>(
+  key: string,
+  f: Fn<T, S>,
+  timeMs: number
+): Fn<T, S> {
+  return async (...args) => {
+    return new Promise(async resolve => {
+      let renewCache = true;
```
rileyjshaw: This doesn't appear to be used.

server/src/lib/lazy-cache.ts
```diff
@@ -53,29 +76,13 @@ function redisCache<T, S>(
...
-      }
-
-      if (renewCache) {
-        value = await f(...args);
-        await redis.set(key, JSON.stringify({ at: Date.now(), value }));
-        resolve(value);
-      }
-
-      await lock.unlock();
+      resolve(await renewRedisCache(key, f, timeMs)(...args));
```
rileyjshaw: I'll need to dive into the surrounding files to get more context on how / where this is being called, but I wonder why we're calling `renewRedisCache` in two places now?

phirework: So the old logic used to be "if cached value exists and is not expired, return cached value. Under all other circumstances, call a fresh query and cache it in Redis, and when you're done, return that new value". Only one call to update the cached value was needed because whether the value expired or never existed in the first place, you had to wait for it to come back before doing anything. 

With the new logic, if the cached value exists but has expired, we're no longer waiting for the update function to come back because we're sending back the expired value regardless, but the call to update the expired value has to happen before the `return`. So we can't combine it with the same `await renew` that we'd need to perform if the value didn't exist at all. 

Does that make sense? It's totally possible I'm overcomplicating it.

phirework:
See comment in #2476

PR #2475, opened 2019-12-12
remove all sentence collector json

MichaelKohler:
Writing the json files doesn't make much sense anymore as it only captures approved sentences. And these approved sentences are part of the txt files anyway.


MichaelKohler:
@nukeador r?

MichaelKohler:
As discussed with Nuke we will keep them for now

PR #2476, opened 2019-12-12
Change default retry count

phirework:
Hey @Gregoor - we're running into a bunch of issues with Redis just spinning itself into a frenzy trying to acquire a lock and literally `eval`ing thousands of times per minute. Is there any danger you can think of if we changed the retryCount from infinite to the default of ten (or some other finite value)?

### DIFF
--- a/server/src/lib/redis.ts
+++ b/server/src/lib/redis.ts
@@ -4,7 +4,7 @@ import { getConfig } from '../config-helper';
 
 export const redis = new Redis(getConfig().REDIS_URL);
 
-export const redlock = new Redlock([redis], { retryCount: -1 });
+export const redlock = new Redlock([redis]);
 
 export const useRedis = new Promise(resolve => {
   redis.on('ready', () => {


Gregoor:
Hm do we know why lock acquisition takes that many retries? I wonder whether we should fallback to the local cache in those cases instead?
If it does take a long time to connect to redis, every request for something from redis would wait around until it gets back or timeouts. So this could potentially OOM our web nodes out of existence (and thankfully new ones being spawned)?

Sorry more questions than answers 🙈

phirework:
@Gregoor I don't know for sure, but New Relic is showing that every call to a leaderboard page is associated with 5-10k redis evals (which are only called for the lock, unlock, and extend scripts). I suspect the leaderboard pages will need to be revamped entirely, but because the whole redis infrastructure is new to me I'm trying to figure out where the chokepoints are.

I do have another PR here that changes the logic of the `redisCache` slightly that will at least show people a cached version of the leaderboard and try refreshing it in the background, but I don't think that'll fix this problem, only how the user experiences it: https://github.com/mozilla/voice-web/pull/2474

Gregoor:
Yaiks, sorry that that blew up in your face and not in mine. Looking at the code I wonder why I put the lock acquisition before reading the value. My present-mind thinks a lock only needs to be acquired when an update to the value is needed 🤔 Just an idea.

phirework:
Oooh good flag on that, I'll take another look at that piece.

phirework:
Closing this one and #2474 for now - we have a new sandbox environment that we can test this featureset more holistically in and I will create a new PR for that.

PR #2481, opened 2019-12-14
Move generation of security headers to node

### LINKED ISSUE #2480: Move special HTTP headers handling in node
Right now, apache is doing some of that, and it doesn't make much sense.

Better if the node application does it all itself.

gozer:
Fixes #2480

Sneaking in:
 - X-Release-Version
 - X-Environment


phirework:
Looks good to me!

PR #2483, opened 2019-12-14
Handle HTTP => HTTPS redirection in prod

### LINKED ISSUE #2482: Handle http => https redirection on our own
Makes sense for production and one more thing Apache doesn't need to do for us

gozer:
Handle HTTP => HTTPS redirection in PROD

This is based on the X-Forwarded-Protocol header set by upstream reverse-proxies, like ELBs and the like

Fixes #2482
Depends On #2481 

Sneaking in:
- http-status-codes for human-readeable http status codes
- [travis] skip docker login when building a PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_script:
     fi
   - echo "Branch $SAFE_BRANCH_NAME"
   - echo "Tag $TRAVIS_TAG"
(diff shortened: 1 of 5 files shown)


phirework:
Love the human readable status codes! Thanks for removing more of the coupling from apache.

PR #2491, opened 2019-12-18
add Fullstory tracking code, add CSP hash

phirework:
Also added a `prettier-ignore` to the `index_template` so that it wouldn't try to reformat the inline script and thereby mess up the CSP hash.


phirework:
(The other modified CSP hash is to fix an existing inline-eval bug currently on prod, unrelated to Fullstory.)

PR #2494, opened 2019-12-20
Handle setting cache-control max-age expiry directives depending on content-type

### LINKED ISSUE #2493: Set content-type based Cache-Control max-age headers

gozer:
Note: The values in here are somewhat out pulled out of a hat, might need adjusting

Fixes #2493


phirework:
I would also remove the corresponding cache settings from `web.pp`. Otherwise this looks good to me, @rileyjshaw any thoughts?

MichaelKohler:
What's the status here?

server/src/server.ts
```diff
@@ -135,11 +148,18 @@ export default class Server {
     app.use('/api/v1', this.api.getRouter());
 
     const staticOptions = {
-      setHeaders: (response: express.Response) => {
+      maxage: MAX_AGE['default_ms'],
```
rileyjshaw: Nitpick: `MAX_AGE.default_ms`

server/src/server.ts
```diff
@@ -135,11 +148,18 @@ export default class Server {
...
+      setHeaders: (response: express.Response, path: string) => {
         // Basic Information
         response.set('X-Release-Version', RELEASE_VERSION);
         response.set('X-Environment', ENVIRONMENT);
 
+        // Type specific cache headers
+        var max_age;
+        if ((max_age = MAX_AGE[serveStatic.mime.lookup(path)])) {
+          response.setHeader('Cache-Control', 'public, max-age=' + max_age);
+        }
```
rileyjshaw: Nitpick:

```
const max_age = MAX_AGE[serveStatic.mime.lookup(path)];
if (max_age) {
    response.setHeader('Cache-Control', 'public, max-age=' + max_age);
}
```

rileyjshaw:
(Fixed the merge conflict)

phirework:
Thanks for fixing the merge conflict! This was part of some work we were doing on Apache back in December, when we were testing different configurations (esp with the timeout limit) to see if it made a difference in our rate of 502s. I've left this open for two main reasons: 

* I have no idea if there are different needs in K8s land and it was low priority enough that I didn't wan tot bug Alberto about it until everything else was done
* I always thought the standard cache control timeout for static things like images and fonts was way longer, like 30 days or more, and didn't have time to properly research the implications 😅

PR #2495, opened 2019-12-20
Run prettier 😬

rileyjshaw:
We updated prettier, and now it's time to face the consequences!

I guess this commit makes our code more ✨beautiiful ✨, but it also
opens us up to a bunch of merge conflicts. If you're not working on any
of these files, feel free to merge! But I totally understand if you'd
rather wait until something else has landed.

Mostly treating this commit as an FYI that we should do this eventually,
and next week might be a good time since code churn will be at 50% :)

To reproduce: `yarn prettier`


phirework:
This looks good to me!

PR #2505, opened 2019-12-27
removed > 13k sentences with non-german letters

stefangrotz:
I removed many sentences containing non-german letters, mainly cyrilic, arabic and chinese one:

I removed all lines containing:
* ð|ð|À|Á|Â|Ã|Å|Æ|Ç|È|É|Ê|Ë|Ì|Í|İ|Î|Ï|Ð|Ñ|Ò|Ó|Ô|Õ|Ø|Ù|Ú|Û|Û|Ý|Ž|à|á|â|ã|å|æ|ç|è|é|ê|ë|ì|í|î|ï|ð|ñ|ò|ó|ô|õ|ø|ù|ú|û|ý|þ|ÿ|ā|ă|ą|ć|ċ|č|ď|đ|ē|ĕ|ė|ę|ě|ğ|ġ|ģ|ħ|ĩ|ī|ĭ|į|ı|ķ|ĸ|ĺ|ļ|ľ|ŀ|ł|ń|ņ|ņ|ṫ|š|Ў|ḃ|ḋ|ḟ|ṁ|ṗ|ṡ|ẁ|ẃ|ẅ|ẛ|ỳ|α|β|Γ|γ|Δ|δ|ε|ζ|η|Θ|θ|ι|κ|Λ|λ|μ|ν|Ξ|ξ|Π|π|ρ|Σ|σ|ς|τ
* аАбБвВгГдДеЕёЁжЖзЗиИйЙкКлЛмМнНоОпПрРсСтТуУфФхХцЦчЧшШщЩъЪыЫьЬэЭюЮяЯ
* [ء-ي]
* I removed mostly single chinese letters with regex like "^. ".  or " . " (for some reason single chinese Letters haven't been removed by the extractor script)
* I inserted a few spaces on the End of existing sentences by accident during some manual deletions, that's why there are 8 additions. I would fix this but I don't know how I can find these lines in the diff in such a huge file.

Resulting in 13 568 deletions. I checked the first 1000 deletions manually and I am confident, that almost all of the deleted sentences would be considered as very hard or impossible to pronounce.


MichaelKohler:
I'll do a quick review once I am back on my laptop. I guess I'll also add these to the blacklist for the scraper to make sure we capture these in case we use it for anything else (until we also have a whitelist)

stefangrotz:
>  I'll do a quick review once I am back on my laptop. 
Thanks, but I will update this pull request in the next hour because...
> I guess I'll also add these to the blacklist for the scraper to make sure we capture these in case we use it for anything else (until we also have a whitelist)
Good Idea. Maybe this isn't the right place for this discussion, but the closest thing to a whitelist I got until now ist this unicode regex that excludes all non german letters. (doesn't work with sed though):
`[^\u0000-\u007BäöüßÄÖÜ„“‚‘’–] `

I tried this regex with the regex search in VSCode and I found a lot of things that should be deleted, including '{' and '}' . I will update this pull request soon.

stefangrotz:
Okay, in the second commit I deleted:

- […|ș|ế|ō|\{|\}|ů|»|ū|Ł|ş|«|ş|ǐ|ż|ő|ň|ự|ň|ž|ị|Ō|ŏ|Č|Š|ř|ś]
- the apprevation dt. (german)

The more I delete the more things I find that have to be deleted. I will stop for now, I think in the long run a whitelist will be much more realistic (maybe combindes with a whitelist for words like Škoda)

stefangrotz:
Yes, I wasn't sure about these two signs and one could also think about changing them with search and replace or accepting  them in general. There are at least four different types of quotation marks and this one appeared very often in sentences that contained foreign words.

stefangrotz:
Okay I updated the pull request, "…", "»" and "«" are part of the commit again.

phirework:
@stefangrotz @MichaelKohler is this good to go?

MichaelKohler:
Yes

stefangrotz:
Yes :)

Issue #2488, opened 2019-12-17
Foreign language clips appearing in English validation queue


dabinat:
I've been noticing foreign language clips appearing in the English validation queue recently. These contain characters that do not exist in English and would have been filtered out by Sentence Collector.

Additionally, the user speaks them fluently and without hesitation, which would probably not be the case if the user was expecting to speak a clip in English.

So for these reasons, I think this is not a case of a sentence in the wrong language being mistakenly included but that clips correctly recorded in other languages are somehow making it into the English validation queue.

![IMG_8075](https://user-images.githubusercontent.com/18251622/71032174-a8c1a600-20c9-11ea-9a56-6619a4b27571.PNG)

![IMG_7961](https://user-images.githubusercontent.com/18251622/71032189-b119e100-20c9-11ea-9da4-4a8724df014d.PNG)

stefangrotz:
I have seen similar things in German. I think most of these sentences can be avoided by using a whitelist for the alphabet of one language or a blacklist for letters like cyrilic, greek and non-english-latin letters. This blacklist should be part of both, the wiki extractor and the sentence collector.

phirework:
Yikes, super good flag, thanks you two!

phirework:
@stefangrotz @dabinat Circling back to this, do you have an approximate sense of when this started?

dabinat:
@phirework Not sure... maybe a week or so before I reported it.

But the clips should be fairly easy to find because many of them contain characters that don’t exist in English (see my screenshots above).

dabinat:
@phirework I was validating some clips and came across a foreign language one. I reported it and moved onto the next clip. The next clip was the exact same voice speaking a sentence in English.

This made me wonder if this has something to do with recording in one language and then switching and recording in another. Perhaps a cache or session variable somewhere still retains the previous language ID?

phirework:
@dabinat That's a great clue, thanks for that. We have a migration script that'll fix the ~700 clips that are currently mislabelled that'll land on production later this week / early next week, and I'll file a JIRA ticket to investigate the language setting mechanics more fully.

stefangrotz:
The same bug is happening in Esperanto in the last Days. It is a language where it is very likely that people also donate other languages.

What is the best thing to do when this happens? Ignore the sentence, klick no or report it?

MichaelKohler:
There is another mention of that here: https://discourse.mozilla.org/t/beim-anhoren-satze-in-anderer-sprache/55729

[...]

PR #2507, opened 2019-12-28
Add migration script for rewriting locale_ids

phirework:
In investigating issue #2488, discovered that there are ~650 clips whose `locale_id` do not match that of the original sentence. I'm not sure why this is happening because it seems pretty intermittent: the issue happened a BUNCH on May 9th 2018, then a handful of times on June 1st 2019, July 16 2019, Oct 18th 2019, Nov 2nd 2019, Nov 23rd 2019, and then steadily but not overwhelmingly since Dec 7th. My best guess based on timing is that this might be related to server load, since the dates in Oct/Nov coincide with big bugs we had to fix, and we've been fighting server performance issues all December but I haven't been able to find anything in the code that would point to that being a problem.

In the absence of any sort of useful debugging information, I want to update all of the offending clips so that the `locale_ids` DO match so that the users aren't affected, and keep an eye on this. Hopefully, the next time this happens again we might actually be able to correlate Kibana/Sentry error logs with the timing of the clip being saved.

### DIFF
--- a/server/src/lib/model/db/migrations/20191227183100-match-clip-sentence-locale.ts
+++ b/server/src/lib/model/db/migrations/20191227183100-match-clip-sentence-locale.ts
@@ -1,0 +1,14 @@
+export const up = async function(db: any): Promise<any> {
+  // Note: Manual backfill to follow.
+  return db.runSql(`
+    UPDATE clips
+    INNER JOIN sentences
+    	ON clips.original_sentence_id = sentences.id
+  	SET clips.locale_id = sentences.locale_id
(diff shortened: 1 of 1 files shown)


rileyjshaw:
😖 this is really strange. I don't understand how / why this would happen at increased server load, but the query itself looks fine!

phirework:
Super interesting additional debugging information from @dabinat here: https://github.com/mozilla/voice-web/issues/2488#issuecomment-570942520 I've filed a ticket for it: https://jira.mozilla.com/browse/OI-1405

PR #2519, opened 2020-01-06
UI[web]: Fix background size in dataset page

Kriyszig:
The background image set using CSS contained within the div instead of covering the entire background. This led to issues like:
* Displeasing UI
* The heading getting hidden in the white background
This is prominent on screens of tablet.

#### Heading
Before:
Imgur 
After:
Imgur

#### Newsletter Subscription
Before:
Imgur
After:
Imgur

**Edit**
Issue in focus: #2520


Kriyszig:
Thanks for the quick response @phirework :octocat:

PR #2524, opened 2020-01-07
Update locales

MichaelKohler:


locales/native-names.json
```diff
@@ -75,14 +76,15 @@
   "or": "ଓଡ଼ିଆ",
   "pa-IN": "ਪੰਜਾਬੀ",
   "pl": "polski",
-  "pt": "Português",
+  "pt": "pt",
```
MichaelKohler: that's a bit sad, where did we lose that? :(

phirework:
Hmm, what do you mean by lose? Should it be “pt”?

MichaelKohler:
Interesting, looks like the pt localization file still has pt-BR as key, where en has pt. So we can't map it correctly in the export script, and therefore do a fallback to "pt".

https://github.com/mozilla/voice-web/blob/master/web/locales/pt/messages.ftl#L99

I thought this was supposed to be synced by Pontoon as the English one is the base and is correct?

MichaelKohler:
Also, FWIW, it's probably best to have both

https://github.com/mozilla/voice-web/pull/2524
https://github.com/mozilla/voice-web/pull/2525

merged before this one.

phirework:
@MichaelKohler This is #2524, did you mean a different one?

@nukeador do you have any insight into the `en`/`pt` discrepancy? I'm not super familiar with that process

MichaelKohler:
> @MichaelKohler This is #2524, did you mean a different one?

I have absolutely no idea why I listed that one. I only meant the Sentences one, which you just merged. Nevermind me. Good catch!

MichaelKohler:
Oh, I meant https://github.com/mozilla/voice-web/pull/2525 as this one here includes pa-IN as contributeable. There we go :)

phirework:
Going to merge this in right now despite the PT issue, that's a Pontoon syncing thing that I'm not familiar with and I don't want it to hold up launching other languages while we investigate.

PR #2525, opened 2020-01-07
Add pa-IN

MichaelKohler:
pa-IN has collected and approved sentences, but those got never exported, as the Sentence Collector used "pa" and voice-web uses "pa-IN". I've now added that mapping in the Sentence Collector Export so we get those sentences too.


MichaelKohler:
@nukeador r?

nukeador:
What I'm thinking is that we should probably ask @peiying2 to use locale codes without country names, in this case "pa" following this standard

https://www.unicode.org/cldr/charts/latest/supplemental/languages_and_scripts.html

phirework:
@nukeador Are we waiting on a verdict from Peiying or can this be merged?

nukeador:
Let's go for this now. But we really need to review standardize our locale list to match with that unicode one, that will imply changes on the codes from pontoon that will have to be reflected on our side.

PR #2544, opened 2020-01-25
Correct spellings & remove symbols, abbreviations

Mohsin274:
Symbols and abbreviations are removed, some spellings are corrected, letter and words (like র) that are not used in this language are replaced with the correct ones.


MichaelKohler:
Hi

Sentences that are in the sentence-collector.txt file are coming from the Sentence Collector at https://common-voice.github.io/sentence-collector/ . This means that removing and correcting them here won't change them in the Sentence Collector database and would be overwritten with the next export.

It seems that you had to change 4618 sentences out of the existing 6142. That would mean the error rate is 75%, and all these still get reviewed positively by two persons in the Sentence Collector. Can you elaborate more on what the errors were? Does this also involve different dialects? It seems quite shocking to me that there would be that many errors.

There is currently no validation rules file for Arabic. So it's falling back to the English one, which for Arabic won't catch anything. This could explain why uses were not warned about sentences being wrong. There is an open issue here to tackle this: https://github.com/Common-Voice/sentence-collector/issues/281#issuecomment-575887802 . Would be great if you could help out there.

For the changes to the database, I would need the following - if we indeed need to fix all these:
* List of sentences to be removed, preferably in a text file - one sentence per line including all sentences that are wrong
* Corrected sentences should be re-submitted through the Sentence Collector at https://common-voice.github.io/sentence-collector/

Thanks
Michael

Mohsin274:
I've just removed the "।" symbol (which is used as a full stop) because it might cause issues. If it's possible to define it as punctuation marks then there's no need to remove it. 
Replaced র (Not used in Assamese) with ৰ. These sentences got validated because when this letter is combined with other letters (Eg: প+র and প+ৰ both become প্র) both appears to be correct but is not valid. So, it's difficult to figure out whether the correct letter is used. This is actually an issue with Google's Transliteration Keyboard that many people aren't aware of. 
Replaced া‌‌ে (it's basically two seperate vowel marks া and ে used together, which is again invalid although appears to be correct) with ো (it's a single vowel mark, the correct one).
And also replaced  য়, ঢ় and ড় (which are basically য‌়, ঢ‌়, ড‌়) with য়, ঢ়, ড় respectively. In this case, although it's not invalid to use them but most of the words contain the single letters (য়, ঢ়, ড়), so it's better to keep this uniform.

MichaelKohler:
> I've removed the "।" symbol (which is used as period) because it might cause issues. If it's possible to define it as punctuation marks then there's no need to remove it.

Can you elaborate on which issues those are?

> Replaced র (Not used in Assamese) with ৰ. These sentences got validated because when this letter is combined with other letters (Eg: প+র and প+ৰ both become প্র) both appears to be correct but is not valid. So, it's difficult to figure out whether the correct letter is used just by reading it. This is actually an issue with Google's Transliteration Keyboard that many people aren't aware of.
Replaced া‌‌ে (it's basically two seperate vowel marks া and ে used together, which is again invalid although appears to be correct) with ো (it's a single vowel mark, the correct one).
And also replaced য়, ঢ় and ড় (which are basically য‌়, ঢ‌়, ড‌়) with য়, ঢ়, ড় respectively. In this case, although it's not invalid to use them but most of the words contain the single letters (য়, ঢ়, ড়), so it's better to keep this uniform.

This sounds like something that could be added to the rules: https://github.com/Common-Voice/sentence-collector/blob/master/shared/validation/VALIDATION.md

As I said, generally we would need to remove all these sentences from the Sentence Collector database, and then resubmit all the corrected sentences again for review. Given that this is quite a few sentences, I would like to get a review from somebody else here as well. What I'm gonna need in the end is a text file with one wrong sentence per line to delete them from the Sentence Collector.

Mohsin274:
Ok then, I'll prepare the text file. But, for a change, what if I upload the corrected sentences in a separate file instead of uploading them through the sentence collector tool? Because the review tool is very inconvenient and moreover there are too many sentences to upload and review. So it'll be very frustrating and time consuming for me to review them one by one.

Mohsin274:
Regarding "।" symbol,
Here are some examples:
মই স্কুললৈ যাওঁ। (I go to school.)
মই গাওঁ। (I sing.)
As you can see, here "।" is used to mark the end of the sentences.
My question is - Is it okay to use this symbol? As I've observed, there's no "।" symbol at the end of the sentences in other indian languages. I understand why this may have been removed. Maybe it's considered to be a Letter rather than a symbol. So, if it's okay to use them, there's no need to remove the sentences having that symbol.

MichaelKohler:
> So, if it's okay to use them, there's no need to remove the sentences having that symbol.

I'm no expert, but I think it's not absolutely necessary to remove them if they are not wrong in terms of the language.

Mohsin274:
They are not wrong. As the Sentence collector tool doesn't allow any symbol but at the same time there's no validation rules for this language, so I was wondering it may cause problems later.

PR #2555, opened 2020-02-04
Enable backend monitoring using Sentry

johngian:


johngian:
@phirework thoughts? I tested it locally with a free sentry account and errors seem to be working fine.

phirework:
Do you want to track backend errors in the same project as the frontend ones? If yes, it'll probable be useful to pass along some parameter with each error report to make filtering between backend and frontend easier. If not, you'll need to file a Bugzilla ticket to create a new CV backend project in our #voice team in Sentry, I don't have permisison to administer that.

johngian:
We already have the reporting for frontend and ideally I would like to get sentry tracebacks for the backend as well. Sentry setup is different for React and NodeJS so they cant be on the same sentry project.

The config part is not blocking this PR, since if the config is missing then the backend just skips reporting. Also regarding access, I think I have permissions to create new projects.

phirework:
I know, my question was about whether you wanted a new Sentry project that's JUST for backend or whether you wanted to reuse the existing frontend tracking. I'm not concerned about whether Sentry is capable of tracking both (I know it can handle multiple languages in the same repo), more about separating noise from signal. If you're not concerned about too many emails from the frontend making it harder to notice critical backend reports then that's fine and this is good to go.

johngian:
I think i would prefer a separate project but i don't have strong opinions about it.

phirework:
Cool, I renamed the current instance to `common-voice-frontend` and created a new `common-voice-backend` project here: https://sentry.prod.mozaws.net/operations/common-voice-backend/. Feel free to merge when conflicts are clear!

PR #2563, opened 2020-02-13
Remove functions that save sentence text, vote text to S3

phirework:


rileyjshaw:
Nice - good riddance 😂

PR #2579, opened 2020-02-24
Fixing Copy Link Button

### LINKED ISSUE #2574: On Clicking Multiple times the "Copy link" button , we get multiple "Link copied" pop-up
**Describe the bug**
On clicking the share link button , we get multiple pop-up of Link Copied

**To Reproduce**
1. Go to https://voice.mozilla.org
2. Record a clip
3. Click the upload button
image
4. Click the Copy link multiple times
You will come across
Screenshot from 2020-02-23 23-27-25

** Expected behavior**
Pop up should come once irrespective of the number of times copy link is clicked

**Additional Behaviour**
Can I work on this bug ?

aryamanpuri:
Fixes #2574


aryamanpuri:
Please Review @phirework

web/src/components/share-buttons/share-buttons.tsx
```diff
@@ -21,13 +21,18 @@ interface Props extends LocalizationProps {
...
   const encodedShareText = encodeURIComponent(
     shareTextId
       ? getString(shareTextId, { link: SHARE_URL })
       : getString('share-text', { link: SHARE_URL })
   );
   const shareURLInputRef = useRef(null);
+  const onlinkedCopied = () => {
+    setCopyClicked(copyClicked + 1);
+    setTimeout(() => setCopyClicked(0), 5000);
+  };
```
rileyjshaw: I think that this state should be tied to the actual banner state. Adding a `setTimeout` may approximate the behaviour we're looking for, but it's disconnected from the notification system and could lead to confusing behaviour.

If `addNotification` returned event or timing information, I'd recommend using that. Unfortunately, it doesn't. Adding that functionality would add more complexity than it's worth.

Since you're trying to reduce the number of banners that are _displayed,_ I recommend taking a look at our notification rendering code at `web/src/components/app.tsx:239`.

aryamanpuri: @rileyjshaw It is only a problem with the number of banners, so we can reduce the timeout gap to decrease the confusing behavior.
Please help to solve the problem.
This changed code is fully functional

rileyjshaw: @aryamanpuri: as I mentioned in my previous review, the `setTimeout` option is not an acceptable solution. Rather than editing the `addNotification` logic, I think your best bet is to look into preventing exact duplicate notifications from rendering in `web/src/components/app.tsx:239`.

Please also see my [note on communication in your other PR](https://github.com/mozilla/voice-web/pull/2590#discussion_r387453083).

Thanks!

aryamanpuri: Can we use the `setTimeout` function in the app.tsx?
Possible solutions :
1. Whether it is `Pill` or `Banner`, setting a timeout of 3 seconds ( in this case) , will suspend the notification for re-rendering for 3 seconds.
2. Storing the current notification for three seconds, and check whether it is being rendered or not.
If it is being rendered, then no action will take place.
If it is not rendered, then the notification will be rendered and current notification will be stored in a state for three seconds which I guess would be a more complex case.

rileyjshaw: > Can we use the `setTimeout` function in the app.tsx?

For the reasons mentioned above, no solution using `setTimeout` will be acceptable. In our React app, we prefer to tend with asynchronous behaviour using event listeners or the React lifecycle. The callback should be tied to an event.

However, we shouldn't need any asynchronous behaviour for this. We can add a `.filter(…)` call in place of the `.slice()` on line 240 to filter out any duplicate notifications.

Adding such a function will require exhaustive testing. If you don't feel like following through with that method, no worries! It's the only option I'm currently comfortable with, though.

aryamanpuri: Agree with you, using a timeout function will become a complex case for the platform, and using a filter function will a good solution.
No No, I just was asking about whether to use the function or not in this problem .
> Adding such a function will require exhaustive testing. If you don't feel like following through with that method, no worries! It's the only option I'm currently comfortable with, though.

Looking forward to a much better solution

aryamanpuri:
Closing this for now

PR #2593, opened 2020-02-28
fix crash when process audio file: FFmpeg error: pipe:: Invalid data found when processing input

quanghd96:
When many user use this system same time, it is crashed. I found log:
uncaught exception Error: FFmpeg error: pipe:: Invalid data found when processing input at ChildProcess.

I fixed it by listen when process audio file error and add a flag to control it.

### DIFF
--- a/server/src/lib/clip.ts
+++ b/server/src/lib/clip.ts
@@ -103,23 +103,23 @@ export default class Clip {
     Basket.sync(client_id).catch(e => console.error(e));
     const ret = challenge
       ? {
-          glob: glob,
-          showFirstContributionToast: await earnBonus('first_contribution', [
-            challenge,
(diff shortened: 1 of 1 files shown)


phirework:
Hi @quanghd96, thanks for submitting this and welcome to the repo! Have you been able to reproduce the error locally to test that this is working as expected?

quanghd96:
You can test with body of this api not a valid mp3 file. After I fixed it, server returns error code 500:
image

phirework:
Hi! Apologies for this dropping off the radar. I'm going to close this, as it targets a branch we no longer use. Would you be interested in resubmitting this PR to `main`, as that is our current default branch? If not, I'm also happy to port the changes over manually.

Issue #2038, opened 2019-05-17
"Finish recording first" prompt should be automatically closed after TOS were agreed


rleitan:
[STR]:
1.Open new/uncached session
2.Navigate to https://voice.allizom.org/en/speak (without being logged in)
3.Record a clip then click "back arrow button" located in the upper left corner
4.Click "Submit clips" from "Finish recording first" prompt
5.Click "I agree" button

[Expected result]:
The clip should be uploaded and the "Finish recording first" prompt should be automatically closed

[Actual result]:
Clips uploaded message is displayed
"Finish recording first" prompt is still displayed

robertdudaa:
I I'd like to take up this issue. My first contribution here. :-)

aryamanpuri:
@Gregoor @phirework @MichaelKohler Is this still open to work upon ?

MichaelKohler:
@aryamanpuri yes, this is still an issue. Feel free to work on it :)

aryamanpuri:
@MichaelKohler @rleitan  I am not able to reproduce this bug .
Can you share a screen recording please ?

MichaelKohler:
[2020-02-26 18-38-23.2020-02-26 18_51_05.gif.zip](https://github.com/mozilla/voice-web/files/4257370/2020-02-26.18-38-23.2020-02-26.18_51_05.gif.zip)

Sorry had to zip it as it was too big for GitHub.. Note that this is without logging in.

aryamanpuri:
It would be fine to take a new state that would takes account of whether video is uploaded or not .
@MichaelKohler @phirework

aryamanpuri:
How about adding onConfirm() in the onAgree ?
```
`{showPrivacyModal && (
          <TermsModal
            onAgree={this.agreeToTerms}
            onDisagree={this.toggleDiscardModal}
          />
        )}`
```

moz-bozden:
Closing as stale...

PR #2595, opened 2020-02-28
Finish recording Prompt

aryamanpuri:
Adding `resetAndGoHome() ` to the Submit Recording Button which will close the Prompt when the Clip Recording is Uploaded.
This Pull request will Solve #2038

### DIFF
--- a/web/src/components/pages/contribution/speak/speak.tsx
+++ b/web/src/components/pages/contribution/speak/speak.tsx
@@ -16,16 +16,12 @@ import { User } from '../../../../stores/user';
 import API from '../../../../services/api';
 import { trackRecording, getTrackClass } from '../../../../services/tracker';
 import URLS from '../../../../urls';
-import {
-  localeConnector,
-  LocaleLink,
-  LocalePropsFromState,
-} from '../../../locale-helpers';
+import { localeConnector, LocalePropsFromState } from '../../../locale-helpers';
 import Modal, { ModalButtons } from '../../../modal/modal';
 import TermsModal from '../../../terms-modal';
 import { CheckIcon, FontIcon, MicIcon, StopIcon } from '../../../ui/icons';
 import { Button, TextButton } from '../../../ui/ui';
-import { getItunesURL, isFirefoxFocus, isNativeIOS } from '../../../../utility';
+import { isFirefoxFocus, isNativeIOS } from '../../../../utility';
 import ContributionPage, {
   ContributionPillProps,
   SET_COUNT,
--- a/web/src/components/pages/contribution/speak/speak.tsx
+++ b/web/src/components/pages/contribution/speak/speak.tsx
@@ -517,6 +517,7 @@ class SpeakPage extends React.Component<Props, State> {
                     className={getTrackClass('fs', 'exit-submit-clips')}
                     onClick={() => {
                       if (this.upload()) onConfirm();
+                      this.resetAndGoHome();
                     }}
                   />
                 </Localized>


aryamanpuri:
@phirework Any changes needed?

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -521,6 +517,7 @@ class SpeakPage extends React.Component<Props, State> {
                     className={getTrackClass('fs', 'exit-submit-clips')}
                     onClick={() => {
                       if (this.upload()) onConfirm();
+                      this.resetAndGoHome();
```
rileyjshaw: Thanks for finding a pre-existing method to solve the problem! Unfortunately, this will causes a few additional bugs. Your code sends users back to the homepage no matter what. If someone wanted to navigate elsewhere (eg. if they clicked the "Listen" button), or if the upload failed, they would still be directed to the homepage.

The actual bug that we want to solve here, I think, is that `onConfirm()` isn't being called when it should be. It could have something to do with the Terms & Conditions modal popping up?

aryamanpuri: I encountered this  too 
>  Your code sends users back to the homepage no matter what. If someone wanted to navigate elsewhere 

Since the `onConfirm` is working can be cross-checked from the `Exit & Delete clips` button.
So one thing is clear that the function is working. 
> The actual bug that we want to solve here, I think, is that `onConfirm()` isn't being called when it should be. It could have something to do with the Terms & Conditions modal popping up?

Maybe this is happening here, the if condition is checking the value of `this.upload()` on the click of `submit Clips` button which is declared initially false. So that may be the reason for the `onConfirm()` not be called.
Your views?

rileyjshaw: Could be! It's likely some sort of race condition like that. Let me know if you figure it out :)

aryamanpuri: Calling the `this.upload()` on click and waiting for it to get completed.
Then checking that the clip is uploaded or not will make sense?
This solution would be acceptable?

rileyjshaw: Yeah – `this.upload` should probably be asynchronous, honestly. This is the issue in the `upload` method:

```
    if (!hasAgreed && !(user.privacyAgreed || user.account)) {
      this.setState({ showPrivacyModal: true });
      return false;
    }
```
^ meaning `onConfirm` will never run if the user hasn't accepted the privacy agreement yet.

We should wait for the privacy modal to be accepted, then trigger the redirect. Ideally this would all be asynchronous (using callbacks, promises, or async/await). Practically, that might be a huge code change. So I'd also consider a fix that saves `onConfirm` to state until the privacy modal is accepted.

aryamanpuri:
Added `clipUpload` to the state which will take account of whether the clip is uploaded or not.
When the privacy model is accepted and the clip is uploaded, it will trigger the redirect and closes the navigation prompt. 
This works fine for me.

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -519,8 +520,8 @@ class SpeakPage extends React.Component<Props, State> {
                     outline
                     rounded
                     className={getTrackClass('fs', 'exit-submit-clips')}
-                    onClick={() => {
-                      if (this.upload()) onConfirm();
```
rileyjshaw: I'm confused by this change; we still need to call `onConfirm` eventually. Please re-review my previous feedback and let me know if you have questions.

aryamanpuri: This will automatically close the prompt when the clip is uploaded and the modal is accepted that is what is required and to be done by the function. By this, we don't have to call onConfirm to close the prompt.
If this is not a satisfactory change, I will try with your previous feedback.

web/src/components/pages/contribution/speak/speak.tsx
aryamanpuri: >  So I'd also consider a fix that saves `onConfirm` to state until the privacy modal is accepted.

That means, make another function(which will do the same work as of onConfirm), save into the state and call it after the privacy modal is accepted, that will close the navigation prompt. 
Correct me if I am wrong here.

web/src/components/pages/contribution/speak/speak.tsx
aryamanpuri: Screencast 2020-03-26 21_27_48

jessicarose:
Hello @aryamanpuri and let me start out by apologizing for the extreme amount of time that you've had to wait for a response to your really interesting and useful work on this PR. We saw some rather extreme changes to the project in 2020 and the team that began to review your work would have been unable to continue to do so, through no fault of their own.

I'm sorry as well that the code around this workflow has changed dramatically in the 4 years since your work on this, so I'm going to close this PR with my thanks and ongoing apologies. You put an incredible amount of effort into this and if you're interested in future contributions I can promise a much speedier review process in the future.

PR #2596, opened 2020-02-28
Update db schema with new table and populate it

phirework:
Doing these in phases so it's not just one unnecessarily big PR

### DIFF
--- a/server/src/lib/model/db/migrations/20200226094500-newsletter-sub-tables.ts
+++ b/server/src/lib/model/db/migrations/20200226094500-newsletter-sub-tables.ts
@@ -1,0 +1,31 @@
+export const up = async function(db: any): Promise<any> {
+  return db.runSql(
+    `
+    CREATE TABLE user_client_newsletter_prefs (
+      id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+
+      client_id CHAR(36) NOT NULL,
+      email VARCHAR(255) NOT NULL,
+      basket_token CHAR(36) NOT NULL,
+      first_contrib DATETIME,
+      goal_created DATETIME,
+      goal_reached DATETIME,
+      two_day_streak BOOLEAN DEFAULT FALSE,
+      last_active DATETIME,
+      
+      UNIQUE (client_id),
+      UNIQUE (basket_token),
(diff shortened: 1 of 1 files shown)


server/src/lib/model/db/migrations/20200226094500-newsletter-sub-tables.ts
```diff
@@ -0,0 +1,31 @@
+export const up = async function(db: any): Promise<any> {
+  return db.runSql(
+    `
+    CREATE TABLE user_client_newsletter_prefs (
+      id INT UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+
+      client_id CHAR(36) NOT NULL,
+      email VARCHAR(255) NOT NULL,
+      basket_token CHAR(36) NOT NULL,
```
rileyjshaw: Since we already store these in the `user_clients` table, do we need to duplicate them here? Or is the thought that we would remove them from `user_clients`?

phirework: So I definitely plan on removing `basket_token` from `user_clients` but I wanted to wait on that until I was 100% sure that all the refactored code using the new table structure still worked perfectly. (There are a few other deprecated columns I wouldn't mind getting rid of from that db that have since been moved into other tables too.) 

I was going back and forth on whether or not to store a redundant copy of the email in this table as well - I know it doesn't strictly follow normalization practices, but my thinking there is that this table interacts with an external system that identifies users by email and not by client ID, and moreover the number of people who currently have a `basket_token` is ~11k, out of 900k users in the `user_clients` table. If I could avoid having to join to that to find records in this table that seems like it's worth the redundancy. Your thoughts?

rileyjshaw: Okay cool! Totally makes sense, thanks for clarifying.

I do get a bit concerned about denormalizing email, since it's something I can imagine getting out of sync. But if it's important for performance, works for me! I'm going to accept the PR and leave it open in case you change your mind over the weekend. Feel free to land it on when you're back! 🎉

phirework: Added one more entry to #2597 that I would not have caught otherwise - I didn't even realize we let people change their email since it was all associated with SSO. Thank you!!

PR #2599, opened 2020-02-29
Fix #2592: Fixing padding issues on home page and navbar for --lg-only window size

### LINKED ISSUE #2592: Home page layout issues between 991px and 1200px
Between 991px and 1200px window width there are some minor display errors on the front page.

Screenshot 2020-02-27 at 00 09 58

* 991px and upwards to threshold: No padding left on the main description title 
* Exactly 992px: no profile button visible
* When logged out the login button text gets quite weird

Screenshot 2020-02-27 at 00 14 33

@mbransn what's the desired behavior here? How much padding should the text always have? And I guess the profile button should be visible with 992px as it is with 993px.

OsamaAdam98:
Fixed padding issues on home page for windows sizes below 992px where text didn't have any right or left padding.

After change: 

padding-fix

Also fixed weird login button wrapping issues at the same sizes.

After change: 

navbar


MichaelKohler:
@mbransn do you agree this looks better as well? :)

mbransn:
Wonderful, yes +1. My designer heart is swooning at this polish, thank you both!

rileyjshaw:
🎉 thanks @OsamaAdam98, this is great!

rileyjshaw:
Also, good catch on the media query definitions 🔎

PR #2600, opened 2020-03-02
Throttle basket requests

phirework:
This PR relies on both #2596 and #2597. If it's easier to just review this one and close the other two that's totally fine too! This adds a check to each `Basket.sync` that first retrieves the existing newsletter data from the newly created table.

* If it was last updated less than an hour ago, it doesn't send anything
* If it already has data in all of the email triggering columns, it sends a new `last_updated` value and nothing else
* Otherwise it does the full calculation as before

Left some comments in-line too, let me know if anything doesn't make sense.


server/src/lib/basket.ts
```diff
@@ -11,80 +11,154 @@ export const API_URL =
...
-    [client_id]
+    [currentUserStats.client_id]
   );
-  if (
-    !row ||
-    !row.basket_token ||
-    (!row.first_contribution_date && !row.goal_created_at)
-  ) {
-    return;
-  }
```
phirework: Took out this whole block because: 
* If the user doesn't exist in `user_client_newsletter_prefs` they don't get emails and this whole sync function ends, which is already handled on line 148
* Sync function gets triggered for new goals, recording a clip, and submitting a vote. By definition, `first_contribution_date` and `goal_created_at` will never be empty at the same time, because the act of creating one of those is what leads to this function

server/src/lib/basket.ts
```diff
@@ -11,80 +11,154 @@ export const API_URL =
...
+  const [[computed]] = await db.query(
     `
       SELECT
         email,
         basket_token,
         LEAST(
-          (SELECT MIN(clips.created_at) FROM clips WHERE client_id = user_clients.client_id),
-          (SELECT MIN(votes.created_at) FROM votes WHERE client_id = user_clients.client_id)
+          IFNULL((SELECT MIN(clips.created_at) FROM clips WHERE client_id = newsletter.client_id), NOW()),
+          IFNULL((SELECT MIN(votes.created_at) FROM votes WHERE client_id = newsletter.client_id), NOW())
```
phirework: It turns out `LEAST()` will return NULL if one of the values is NULL, which means that if a user has voted but not recorded or vice versa it'll look like their first_contribution hasn't happened yet.

phirework:
Also, thanks to our handy-dandy half-finished K8s setup, I've managed to deploy this to staging even though it's not merged into master yet :D (using staging bc dev/sandbox are currently still being used to test the env config work)

server/src/lib/basket.ts
```diff
@@ -11,80 +11,154 @@ export const API_URL =
...
+    email: currentUserStats.email,
+    last_active_date: toISO(currentUserStats.new_last_active),
+  };
+
+  sendRequest({
+    ...basketConfig,
+    form: data,
+  }).catch(err => console.error(err.message));
+
+  console.log('basket', JSON.stringify(data, null, 2));
```
rileyjshaw: Was this left in intentionally?

server/src/lib/basket.ts
```diff
@@ -11,80 +11,154 @@ export const API_URL =
...
+  return;
```
rileyjshaw: Nitpick: unnecessary line. Returning undefined is the default behaviour.

server/src/lib/basket.ts
```diff
@@ -11,80 +11,154 @@ export const API_URL =
...
+      HOUR(TIMEDIFF(NOW(), u.last_active)) AS hours_elapsed
+    FROM user_client_newsletter_prefs AS u
+    WHERE client_id = ?`,
+    [client_id]
+  );
+
+  return row;
+}
+
+function updateLastActive(currentUserStats: any) {
```
rileyjshaw: This is a pretty complex object; would be nice to have a type definition for it.

server/src/lib/basket.ts
rileyjshaw: Nice catch!

server/src/lib/basket.ts
```diff
@@ -11,80 +11,154 @@ export const API_URL =
...
   console.log('basket', JSON.stringify(data, null, 2));
 }
+
+export async function sync(client_id: string) {
+  await computeGoals(client_id);
+  const currUserStats = await getCurrentStatus(client_id);
+
+  // If no newsletter prefs exist, they don't get email triggers
+  // If last update was within the past hour, do not send another sync
+  if (!currUserStats || currUserStats.hours_elapsed === 0) {
```
rileyjshaw: I think this is fine but haven't tested yet – `hours_elapsed` will definitely be a number at this point, right? (as opposed to, eg, `"0"`)

server/src/lib/basket.ts
phirework: Hmm originally yes, because there was a `console.log` in the original sync function and I just mirrored it. But now that I look at it, we don't want to be logging email addresses persistently, so I will take it out. TY!

phirework:
Thanks so much for the quick turnaround and sorry for all the noise there! The last three were just poorly handled conflict resolves <facepalm>

PR #2601, opened 2020-03-02
Update about page’s "Get involved" copy

rileyjshaw:


web/locales/en/messages.ftl
```diff
@@ -478,6 +478,10 @@ about-get-involved-text-2020 = Want to help make Common Voice even better?
     Great! Get in touch via email or on <discourseLink>Discourse</discourseLink>,
     submit feedback through <githubLink>GitHub</githubLink>,
     or join us on <matrixLink>Matrix</matrixLink>.
+about-get-involved-text-2020-2 = Want to help make Common Voice even better?
+    Great! Get in touch via email or <discourseLink>Discourse</discourseLink>
```
MichaelKohler: given that "email" does not link anywhere, and the team seems to not like random emails, maybe we should remove that completely?

MichaelKohler: also, I guess `about-get-involved-text-2020` could be deleted now? (didn't verify that's it's not used anywhere else)

MichaelKohler: Oh, there is a button with a contact form below it. Nevermind me for that one.

rileyjshaw: Great points, thanks! I've updated the translation ID and removed the old ones. I've also sent out a question about removing the email portion, it makes sense to me.

mbransn: > given that "email" does not link anywhere, and the team seems to not like random emails, maybe we should remove that completely?

> Oh, there is a button with a contact form below it. Nevermind me for that one.

We could link the contact form to 'email', but wouldn't remove it completely at this point.

rileyjshaw: Unless there's a preference for making "email" clickable, I'm happy to leave as-is. That way email remains accessible, but other methods are prioritized.

mbransn: +1, this is why it was designed as such -- glad it comes across this way.

rileyjshaw:
Thanks all!

PR #2602, opened 2020-03-03
Write logs to a file inside the container.

The-smooth-operator:
This will allow us to ship them to a central location

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 FROM node:10-stretch
 
-ARG MAINTAINER="Philippe M. Chiasson <gozer@mozilla.com>"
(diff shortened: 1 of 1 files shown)


johngian:
Can we just use stdout/stderr to ship logs?

The-smooth-operator:
We cannot tail logs from stdout/stderr and ship them.

There are the 2 common approaches:
- Modify the application's logger code to allow shipping logs to a remote location.
- Write the logs to a file, and using a sidecar pod, tail those and ship them.

I decided to go for the second approach to save development time. We are already doing this for other applications. However, if you prefer to go for option 1, I'm more than happy because it's cleaner and uses (slightly) less resources.

PR #2606, opened 2020-03-04
Add SSM support for deployment secrets

johngian:


johngian:
Hey @phirework! I couldn't find a more sane way to have a blocking call to get the secrets on load. I thing having a lightweight entrypoint to run the server is a good paradigm

phirework:
This is a great solution, I like it! Looking forward to testing on dev/sandbox.

PR #2609, opened 2020-03-06
Disable SSM secrets for Nubis deployment

johngian:


johngian:
@phirework r?

Context: We don't want our nubis env (current prod) to fetch SSM secrets. It's only meant to be a feature for our k8s setup.

PR #2617, opened 2020-03-10
Multiple fixes

johngian:
* Undo revert regarding env/ssm setup
* Fix top-level package json `yarn start:prod` command


johngian:
The culprit for the server not starting was the fact that the top-level `yarn start:prod` command was using the old `server.js` and not `main.js`.

johngian:
I tested it on sandbox here: https://github.com/mozilla/voice-web/releases/tag/sandbox-v1.23-nemo-dev

johngian:
@phirework 
* Current k8s setup has the right env vars setup using ConfigMap and k8s secrets and sandbox works.
* Nubis is using `config.json` so it should be fine when it overrides the defaults.
* SSM is disabled by default

Issue #2614, opened 2020-03-08
Unexpected language change in the website


kdalkafoukis:
### Description

The language of the website changes after using auto-fill from chrome with english(uk) selected language.

- preselected language for the website **english**
- Browser **chrome**

### Steps to reproduce

1. Go to https://voice.mozilla.org/en/datasets

![image](https://user-images.githubusercontent.com/4775609/76154793-6dd47a80-60da-11ea-9595-8a2e73b43c13.png)

2. In the email text field choose an email with uk language

![image](https://user-images.githubusercontent.com/4775609/76154900-03bcd500-60dc-11ea-9625-2dc74f3d8af8.png)

3. the result is the language to automatically change to polish even though the selected language is english

![image](https://user-images.githubusercontent.com/4775609/76154807-a70cea80-60da-11ea-8842-49babbaffd6d.png)

My apologies if the issue is irrelevant with the website.

MichaelKohler:
I could reproduce this, very interesting. United Kingdom -> UK -> uk (Ukranian)? The same also happens with a German (de) address for example

MichaelKohler:
@mbransn I assume that is not intended, right?

mbransn:
Nope, not intended. Thanks for flagging @kdalkafoukis -- were you able to download the language you intended?

And to confirm, the website is triggered to localize to the language tied to the auto-complete email address (EN -> UK) ; the dataset language selected to download does not change (remained as EN).

@ovidiuboca1 are you able to reproduce?

MichaelKohler:
> And to confirm, the website is triggered to localize to the language tied to the auto-complete email address (EN -> UK) ; the dataset language selected to download does not change (remained as EN).

Yes, the dataset language does not change, just the UI language.

ovidiuboca1:
@mbransn 
I tested this and I can reproduce it on Windows 10, Mac OS 10.14 with Chrome browser.

phirework:
Is the language code in the URL changing when you hit autocomplete? Or does the URL still say `/en/datasets` and you're just looking at Polish/German/whatever somehow?

ovidiuboca1:
The language code in the URL is changing. If I select Romanian the URL will be https://voice.mozilla.org/ro/datasets

phirework:
Cool, we'll try to get a fix for it in this sprint. Thanks for reporting @kdalkafoukis!

phirework:
[...]

PR #2620, opened 2020-03-11
clearer dataset state names

phirework:
This hopefully addresses #2614. It's a weird one - I can definitely reproduce the bug on prod, but not on any of our non-prod servers or in my local. Chrome Autocomplete doesn't even show the country associated with an email address anywhere except for that one form field on prod specifically. 

This doesn't happen on the other email forms on the site (the Contact form and the Request A New Language form). 

My best _guess_ is that the locale stored in `this.state` (which should be the locale that the user is trying to download) is somehow colliding with the locale in `this.props` and updating the latter when it shouldn't be. However, since I can't replicate the bug locally, I don't actually know if this fixes it! I've renamed the property in `this.state` which seems like a harmless and more clear change anyway, and hopefully this will fix it? 😅

### DIFF
--- a/web/src/components/pages/datasets/dataset-info.tsx
+++ b/web/src/components/pages/datasets/dataset-info.tsx
@@ -40,11 +40,11 @@ const DEFAULT_CATEGORY_COUNT = 2;
 const Splits = ({
   category,
   values,
-  locale,
+  bundleLocale,
 }: {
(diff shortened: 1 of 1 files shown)


rileyjshaw:
Hey, I'm all for clearer names! 😂

PR #2621, opened 2020-03-12
Deploy on merge to stage branch

The-smooth-operator:
Based on the "Transition branching and deployment" document, we want to deploy on merge to "stage" branch and not "staging". This change will accomplish that.


The-smooth-operator:
@phirework could you please review and merge? I don't have permissions for merging here.

PR #2623, opened 2020-03-12
Run optional chaining codemod and Prettier

rileyjshaw:
This is a pretty big change, so if it looks good we should land it sooner than later to avoid conflicts building up.

If we need to re-run it, here are the steps to reproduce:

```
npm i -g optional-chaining-codemod # https://github.com/villesau/optional-chaining-codemod
# From each source directory:
for d in $(find . -type d); do
  optional-chaining-codemod $d/*.ts --parser=ts;
  optional-chaining-codemod $d/*.tsx --parser=tsx;
done
yarn prettier
```

I did a bit of manual cleanup afterwards, but nothing to difficult.

### DIFF
--- a/web/src/components/layout/layout.tsx
+++ b/web/src/components/layout/layout.tsx
@@ -202,10 +202,7 @@ class Layout extends React.PureComponent<LayoutProps, LayoutState> {
     });
 
     const alreadyEnrolled =
-      this.state.showWelcomeModal &&
-      user.account &&
-      user.account.enrollment &&
-      user.account.enrollment.challenge;
+      this.state.showWelcomeModal && user.account?.enrollment?.challenge;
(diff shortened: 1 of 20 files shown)


web/src/components/layout/layout.tsx
```diff
@@ -202,10 +202,7 @@ class Layout extends React.PureComponent<LayoutProps, LayoutState> {
     });
 
     const alreadyEnrolled =
-      this.state.showWelcomeModal &&
-      user.account &&
-      user.account.enrollment &&
-      user.account.enrollment.challenge;
+      this.state.showWelcomeModal && user.account?.enrollment?.challenge;
```
phirework: 😍

phirework:
Looks great, thanks for landing this!

PR #2627, opened 2020-03-13
Fix #2398: Compressing over-sized avatars to allow uploading an avatar of any size

### LINKED ISSUE #2398: Improve avatar image size handling
Create larger image handling capabilities for those people not able to resize/compress images to 200x200px manually -- ideally this is a friendly experience that doesn't require manual image resizing on the contributor end.

Original reference #2388

osamaadam:
So I made what I believe to be a horrible fix to issue #2398. Please don't merge this just yet.
I'd like to brain-storm possible fixes to this problem, so, here's a brief about where the problem lies..

The resizing function was already in place, but the problem was that if the avatar quality is a little too high, the condition in the back-end would not be met so an error banner was being thrown in the front-end and of course, the avatar would not be updated.

What I've done is that I've made a "recursive call" (shouldn't be called more that thrice at worst) that would reduce the quality of the image to half each time the back-end rejects it. 
The problem of course is that I'm sending a full image to the back-end each time to check if the condition is met and TBH, this is making my eyes bleed every time I look at it.

So possible solutions would be:

- Removing the condition in the back-end and risk saving possibly the world's largest 200x200 avatar.
- Just keep this implementation I've done and take the hit if sending image files between the front-end and back-end won't be that bad of an idea considering the images are only 200x200 anyway.
- Sticking to a low image quality like 0.2 which shouldn't get out of hand, but then, some avatars would look really pixelated.

I wasn't smart enough to find a way of compressing images on the front-end beforehand. So, if there's a way please guide me and I'd be happy to give it a try.

Also, I've added a notification banner that shows up when the avatar is successfully uploaded, the string that show in the notification banner won't localize at least yet. I don't know if it's okay to add it now and localize later or if there's already a localized string that indicates something along the lines of "upload complete".

One more thing, I've changed the back-end endpoint to throw errors instead of sending error messages as json strings. We could definitely revert this but I believe it's a good practice? certainly more expressive when dealing with on the front-end.

### DIFF
--- a/server/src/lib/api.ts
+++ b/server/src/lib/api.ts
@@ -232,8 +232,8 @@ export default class API {
     { body, headers, params, user }: Request,
     response: Response
   ) => {
-    let avatarURL;
-    let error;
+    let avatarURL: string;
+    let error: string | null = null;
     switch (params.type) {
       case 'default':
         avatarURL = null;
@@ -273,9 +273,10 @@ export default class API {
 
     if (!error) {
       await UserClient.updateAvatarURL(user.emails[0].value, avatarURL);
+      response.sendStatus(200);
+    } else {
+      response.status(406).send(error);
     }
-
-    response.json(error ? { error } : {});
   };
 
   saveAvatarClip = async (request: Request, response: Response) => {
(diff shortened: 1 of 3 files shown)


OsamaAdam98:
Okay, so I found a more civilized way of compressing on the front-end before sending to the back-end.
I think this is a reasonable implementation. However, the localization issue I mentioned before is still unsolved.
Hope you guys could check this on Monday!

phirework:
Closing as the avatar system has been fixed and this level of compression is no longer necessary. Thanks for all your work!

PR #2629, opened 2020-03-17
Add resolutions pin for minimist

phirework:
There's a security vulnerability in `minimist`, which in our repo is mainly used by `db-migrate` and its dependencies: https://snyk.io/vuln/SNYK-JS-MINIMIST-559764

I've pinned a resolution for v1.2.3, which gave me some errors as some of db-migrate's dependencies are relying on minimist versions 0.0.1 (!!) and 0.0.8. I dropped my local db and re-ran all migrations and didn't have an error, so I think this is fine, but let me know if you found something in testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
   "resolutions": {
     "cryptiles": "4.1.2",
     "event-stream": "3.3.4",
-    "@types/react": "16.9.19"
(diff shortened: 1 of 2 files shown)


rileyjshaw:
Eek!! Thanks, lgtm :)

PR #2631, opened 2020-03-20
Upgrade @fluent/react, @fluent/bundle

stasm:
In https://github.com/projectfluent/fluent.js/pull/458 I've been working on a new version of `@fluent/react` (né `fluent-react`), rewritten in TypeScript. The rewrite required some API changes, and I used `voice-web` as a test project, given it's a large React app also written in TypeScript.

I'm opening this draft PR to share my changes so far. Once the new version of `@fluent/react` is released, I'll be happy to finish working on it.


web/src/components/layout/footer.tsx
```diff
@@ -15,8 +15,11 @@ import './footer.css';
 const LocalizedLocaleLink = ({ id, to }: { id: string; to: string }) => {
   const [locale] = useLocale();
   return (
-    <Localized id={id} onClick={() => trackNav(id, locale)}>
```
stasm: `Localized` doesn't render any own DOM, nor does it pass props down to children directly. The `onClick` here probably never worked.

web/src/components/pages/about/get-involved.tsx
```diff
@@ -46,9 +46,11 @@ const GetInvolved: React.ComponentType = () => {
...
         <Localized
           id="about-get-involved-text-2"
-          discourseLink={<DiscourseLink />}
-          githubLink={<GitHubLink />}
-          matrixLink={<MatrixLink />}>
+          elems={{
+            discourseLink: <DiscourseLink />,
+            githubLink: <GitHubLink />,
+            matrixLink: <MatrixLink />
+          }}>
```
stasm: The new API is more explicit about the elements passed into translations.

web/src/components/pages/contribution/success.tsx
```diff
@@ -149,8 +149,7 @@ function Success({
           <div className="info-card">
             <Localized
               id="help-reach-hours"
-              $hours={10000}
-              $language={getString(locale)}>
+              vars={{hours: 10000, language: getString(locale)}}>
```
stasm: The new API is more explicit about the variables passed into translations as arguments.

web/src/services/localization.ts
```diff
@@ -28,12 +29,12 @@ function* asBundleGenerator(
 ) {
   for (const [locale, messages] of localeMessages) {
     const bundle = new FluentBundle(locale, { useIsolating: false });
-    bundle.addMessages(
+    bundle.addResource(new FluentResource(
```
stasm: This is the `FluentBundle` API since version 0.14.

web/src/services/localization.ts
```diff
@@ -67,5 +69,6 @@ export async function createBundleGenerator(
     ])
   );
 
-  return asBundleGenerator(localeMessages, messageOverwrites);
+  return CachedSyncIterable.from(
+    asBundleGenerator(localeMessages, messageOverwrites));
```
stasm: I had to pull in the `cached-iterable` dependency to ensure that the generator is not depleted on re-renders of the `LocalizationProvider`. I'm working on streamlining the API to make it easier to use and not require such obscure measures.

phirework:
This is awesome, thanks Stas!

web/src/services/localization.ts
stasm: I published `@fluent/react` 0.12 which makes the API of `<LocalizationProvider>` simpler. It doesn't take the bundles generator as a prop anymore. Instead, the bundles generator can be used to instantiate a `ReactLocalization` which can then be passed to `<LocalizationProvider>` as well as stored in the state safely.

stasm:
I think this is now ready for a review. I tested this locally after merging master into the branch (I didn't include the merge in the PR), and things looked good. I don't know the whole app, however, so I'll be happy to do more manual testing if you'd like me to.

stasm:
@phirework Can you help me find a reviewer for this, please?

phirework:
@stasm Apologies for the delay, I'll take a more detailed look today or tomorrow. We've been busy with our migration to Kubernetes and I didn't want to merge in a big change before also changing server environments.

stasm:
Of course, no worries! I just wanted to make sure this is on your radar.

I've also just checked locally, and it looks like there haven't been any new `<Localized>` added since I created this PR, so it should be good to merge. More than bitrot, I was worried that there would be new code added on master which would use the old `Localized` API, and which would go unnoticed during the merge.

phirework:
Sorry for the late update! So I've finally had a chance to look through everything in more detail. Thanks again for doing this work. 

So the big piece blocking this is that not all of the places we make use of the `<Localized>` object directly references the element, instead it's being passed around in props. You can see missing variable warnings on `/speak`, `/languages`, `/datasets`, etc. So for example, in `contribute.tsx`, these changes need to be made for it to be passed down to `speak.tsx` and `listen.tsx`: 

```
@@ -91,7 +91,7 @@ interface Props extends WithLocalizationProps, PropsFromState {
   errorContent?: any;
   reportModalProps: Omit<ReportModalProps, 'onSubmitted'>;
   instruction: (props: {
-    $actionType: string;
+    vars: { actionType: string; }
     children: any;
   }) => React.ReactNode;
   isFirstSubmit?: boolean;

@@ -423,7 +423,7 @@ class ContributionPage extends React.Component<Props, State> {

               <div className="cards-and-instruction">
                 {instruction({
-                  $actionType: getString('action-click'),
+                  vars: { actionType: getString('action-click') },
                   children: <div className="instruction hidden-sm-down" />,
                 }) || <div className="instruction hidden-sm-down" />}

@@ -500,7 +500,7 @@ class ContributionPage extends React.Component<Props, State> {
             </div>

             {instruction({
-              $actionType: getString('action-tap'),
+              vars: { actionType: getString('action-tap') },
               children: <div className="instruction hidden-md-up" />,
             }) || <div className="instruction hidden-md-up" />}
```

Off the top of my head, the most fail-safe way to look for all of these would be to start with `web/locales/en/messages.ftl` and check all the translation strings that make use of elements or variables. That sounds very tedious and boring and I don't know if you would consider that in-scope for the work you're doing. Is this pull request a blocker to you all releasing a new version on your side?

stasm:
It should be fairly easy for me to collect the names of variables and elements used in all messages, and then grep for them in the code. I didn't realize they were passed like this in objects, thanks for pointing it out.

I'm not blocked on this PR, thanks for asking. I already released `@fluent/react` 0.12 some time ago. I'd like to help to get this PR landed so that it doesn't bitrot completely, making it hard to upgrade to new versions of `@fluent/react` in the future.

stasm:
I updated the PR to fix the cases where the props to `Localized` are passed around as objects. I think I got them all. I first queried the localization files in `web/locales/en` to look for [all variables and markup](https://gist.github.com/stasm/9e789e6841679d8a9c2c2a1fa87962db) used in the translations. I then grepped the source in `web/src` and reviewed each occurence.

One tricky piecy of code was this one:

https://github.com/mozilla/voice-web/blob/fd9d7c7c5bf45b438551c71c6ecb6ddeb8cf4a69/web/src/components/pages/datasets/dataset-info.tsx#L332-L338

It kind of demonstrates why the old API wasn't ergnomic. In the new API, the above becomes simply:

```jsx
<Localized
    id="dataset-description-hours"
    vars={globalStats}
    elems={{b: <b />}}
>
```

I also found two messages which may need some work:

- `dataset-description-hours` [references the `languagesLink` element](https://github.com/mozilla/voice-web/blob/fd9d7c7c5bf45b438551c71c6ecb6ddeb8cf4a69/web/locales/en/messages.ftl#L566) which doesn't seem to be passed into the [corresponding `Localized`](https://github.com/mozilla/voice-web/blob/fd9d7c7c5bf45b438551c71c6ecb6ddeb8cf4a69/web/src/components/pages/datasets/dataset-info.tsx#L329).

- `data-get-started` doesn't seem to be used anywhere. It references the `speechBlogLink` element which also seems to be missing from the codebase.

@phirework Would you mind taking another look at this?

phirework:
This is awesome, thank you! I will take care of cleaning those two instances up separately, good catch. Super appreciate all your work here.

stasm:
Thanks, @phirework!

Issue #2253, opened 2019-10-18
User Profile: Insufficient Space for Left Sidebar


mackncheesiest:
Env: FF 71.0b1

Couldn't find an existing issue that seemed to cover this, but there isn't enough space for the left sidebar on the user profile screen on my monitor (2560x1440)
![image](https://user-images.githubusercontent.com/8922497/67065577-af608b00-f123-11e9-9b10-cde982c043c6.png)

It can be solved by increasing the `max-width` of [this div](https://github.com/mozilla/voice-web/blob/c7233b70eab152a73141e05f17461441a4fedf2d/web/src/components/pages/profile/layout.css#L78), but issues like #1632 bring up wanting to ensure that the edges of each section remain within the logo through the language dropdown.

In that case, does it mean that [--wide-desktop-width](https://github.com/mozilla/voice-web/blob/c4c55866504b63ad87bbffe8f5f19bd4a177120b/web/src/components/vars.css#L34) is too small and should be adjusted?
Or potentially that there should be a different case of styling for `@media(--lg-up)` or `@media(--xl-up)` devices?
Or maybe the profile form should be shifted right a bit to fill in on that side?

It's been a bit since I've done webdev, but whatever the recommendation is, I'm willing to throw up a PR.

phirework:
Closing - can no longer replicate this issue.

mackncheesiest:
I thought I would check, and I can still replicate that screenshot on Win 10, FF 83.0b4

PR #2632, opened 2020-03-20
Fix Issue #2253: Insufficient space for left sidebar

manishkr1101:
### Fix Issue #2253
As mentioned in issue, it is solved by setting `max-width: var(--wide-desktop-width);` to match width with header and `margin: auto;` to align container in center.
Screenshot from 2020-03-20 23-00-44

In second commit both attribute `position` and `max-width` are useless and making design worse in large screen (2560x1440)


ubaidisaev:
is anyone working on this ? please check my solution https://github.com/mozilla/common-voice/pull/2915

PR #2636, opened 2020-03-21
fix(incorrect target anchor)

### LINKED ISSUE #2635: Inaccurate target used in anchor
## Description
No meaning of target="__blank" in the anchor tag for opening a url in a new tab target="_blank" is used:- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

## Additional Information
I would like to work on this

shakti97:
Fixes #2635


phirework:
Thanks for noticing and fixing these! Please refrain from tagging random people in your issues or PRs unless you have previously discussed it with that person, it generates unnecessary noise for them and they are often not the right person to be reviewing your PR. The engineering team monitors Github and we will get to all issues and PRs eventually.

PR #2638, opened 2020-03-21
fix(target vulnerability issue):- added noopener, noreferrer

### LINKED ISSUE #2637: Vulnerability issue with target
## Description
There are multiple links having target but without noreferrer and noopener, which may cause a serious security issue name tabnabbing (https://medium.com/sedeo/how-to-fix-target-blank-a-security-and-performance-issue-in-web-pages-2118eba1ce2f)

 Linking to another page with target="_blank" will run the new page in the same process as your page. If the new page executes JavaScript, your page's performance may suffer. (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a)

## Expected Solution
When using target, add rel="noreferrer noopener" to avoid exploitation of the window.opener API and performance issue as well

shakti97:
Fixes #2637

### DIFF
--- a/web/src/components/layout/footer.tsx
+++ b/web/src/components/layout/footer.tsx
@@ -70,6 +70,7 @@ export default React.memo(() => {
               <a
                 target="_blank"
                 href="https://www.mozilla.org/en-US/privacy/websites/#cookies"
+                rel="noopener noreferrer"
               />
             </Localized>
           </div>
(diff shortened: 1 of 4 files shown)


xrmx:
@shakti97 is there any reason you are citing me? I can't see one :|

shakti97:
> @shakti97 is there any reason you are citing me? I can't see one :|
you didn't have the permission for review and merge ? If not my bad,sorry

phirework:
This looks good, thanks! Again, please refrain from tagging in community members unless you know they are involved in the issue :)

PR #2645, opened 2020-03-25
Add prettier ignore file for sentence data

phirework:
If you try to merge `master` into a local branch, prettier will try to process all the sentence text files it adds and format them. Mostly this is just annoying, sometimes this actually breaks prettier and results in a GC out of memory error, and either way you'll probably end up with annoying merge conflicts.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+server/data/*


rileyjshaw:
Thanks for fixing this!

I'm not too familiar with how / when Prettier runs in our app, but this raises some questions for me. `server/data/*` probably isn't the only directory we want to ignore. For instance, we should also ignore our compiled `js` and `dist` directories, right?

What is triggering Prettier to run on `server/data/*` right now, and is it also running on other unwanted directories? I see that our `yarn lint` command runs `prettier \"**/*.{ts,tsx}\"` – do we need to move this blob into a config file for all our runners to know about it?

(Also – that command probably also run on css, html, and json files, no?)

phirework:
So this is addressing the command that runs as a pre-commit hook on all staged files as specified in `package.json`, which is why it was never triggered when merges are sent in from Pontoon/sentence collector PRs. Our compiled folders are all already in `.gitignore` and are never staged so are bypassed entirely. It's totally possible there are other folders I'm missing, definitely feel free to add if there are any other flags.

rileyjshaw:
Ohhhkay, got it. Thanks for clarifying!

I'll update our `yarn lint` command in another diff re: comments above.

PR #2647, opened 2020-03-26
Update analytics packages

phirework:
In preparation for instrumenting custom events, have moved Fullstory from just being an inline script to installing their full SDK. Also added Amplitude setup code - we don't have an API key right now, but it'll be in the same scope as FS tracking so I thought I'd handle it together.

### DIFF
--- a/web/src/components/app.tsx
+++ b/web/src/components/app.tsx
@@ -12,6 +12,8 @@ import {
 import { Router } from 'react-router-dom';
 import { createBrowserHistory } from 'history';
 import * as Sentry from '@sentry/browser';
+import * as FullStory from '@fullstory/browser';
+import * as Amplitude from 'amplitude-js';
 import { UserClient } from 'common';
 import store from '../stores/root';
 import URLS from '../urls';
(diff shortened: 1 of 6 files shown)


web/src/components/app.tsx
```diff
@@ -332,10 +339,28 @@ class App extends React.Component {
...
-        'https://4a940c31e4e14d8fa6984e919a56b9fa@sentry.prod.mozaws.net/491',
+        SENTRY_FE_DSN,
       environment: isProduction() ? 'prod' : 'stage',
       release: process.env.GIT_COMMIT_SHA || null,
     });
+
+    if (!isProduction()) {
+      console.log('disabling analytics on non production');
+    } else if (!doNotTrack()) {
+      console.log('do not track header set, disabling analytics');
```
rileyjshaw: Should this be `else if (doNotTrack()) {` ?

phirework: LOL duh yes, thank god for code reviews. I actually decided to remove the separate console messages entirely, I realized that since this check is already being done in `index_template` for Google Analytics, repeating the check here would simply result in redundant messaging.

rileyjshaw:
Thanks for the update :)

PR #2649, opened 2020-03-27
Create 2500 validated hours target on lanugages page

rileyjshaw:
Implements [OI-1458](https://jira.mozilla.com/browse/OI-1458)

re: https://discourse.mozilla.org/t/english-has-reached-its-goal-of-1200-hours-of-validated-hours-what-will-be-the-next-goal/53826,
this commit adds a _list_ of hour goals, so that popular languages (eg. English) can reach for higher targets.

I made the decision to keep all goals on the site the same except for
the language cards on /languages. I _think_ this makes sense, but I'm
happy to discuss! Let me know if you think we should do something
different.


phirework:
Looks good to me!

PR #2650, opened 2020-03-27
Add portuguese in portuguese

lokal-profil:


phirework:
This file is automatically generated from Pontoon, so it will just be overwritten next time we sync an export. It looks like this translation is already available and just needs someone to approve it: https://pontoon.mozilla.org/pt/common-voice/web/locales/en/messages.ftl/?status=unreviewed&string=201209 

@MarceloGhelman or @mchagas, could you help with approving this in Pontoon?

MarceloGhelman:
Hi @phirework,
Since Common Voice **pt-BR** locale was removed from Pontoon, united to **pt-PT** on **pt**, I have no rights to approve suggestions,
Can someone allow me to do it? @mchagas?
If so, I'll keep it full translated, as used to do on pt-BR.

phirework:
@peiying2 Is this something you can help with?

PR #2654, opened 2020-03-28
Fix(Search Empty Value): added trim to query

### LINKED ISSUE #2651: Search Should Only Work for not empty value
## Description
Currently Search on the language page is performing search for blank space, which result in the showing "no result"

## Current Behaviour
Search is performing search operation for blank spaces and showing no result found

## Expected Behaviour
Should performing Searching only for non empty value

## Additional Information 
I would like to work on it

shakti97:
Fixes #2651


phirework:
Looks good, thanks!

shakti97:
Thanks for review and merge

PR #2665, opened 2020-04-03
Adding margin left for logo container

aryamanpuri:
There is no left margin to the Common Voice logo when the screen size is above 1200px.
Before Change :
Screen Shot 2020-04-03 at 21 21 15

After Change: 
Screen Shot 2020-04-03 at 21 23 51

### DIFF
--- a/web/src/components/layout/footer.css
+++ b/web/src/components/layout/footer.css
@@ -94,6 +94,7 @@ footer {
             flex-direction: column;
             min-width: 9rem;
             margin-right: 8rem;
+            margin-left: 2rem;
             margin-top: 0;
         }
 


aryamanpuri:
Actually margin is equal in both the sides that is `2rem`

shakti97:
ohh cool, may be the image issue

phirework:
Thanks for submitting this. More broadly speaking the problem here is inconsistent box model usage across the footer, so the best fix would be to apply padding at the `.footer` level for each viewport and the distribute the elements internally, rather than using ad-hoc margins for each element.

aryamanpuri:
I got ur point, while there is a problem both in tablet(<1200) and desktop view(>1200).
Can you provide me with the dimensions to distribute the elements internally for both views?
Maybe @mbransn can help in that.

PR #2666, opened 2020-04-05
fix(button icon disappear on hover)

### LINKED ISSUE #2655: Button Icon Disappear on hover
## Description
Button Icon Disappear on hovering

## Current Behaviour
Icon is white and the background is also white, that's why the button disappear

## Expected Behaviour 
Icon color should also change with the text, and will be visible when hovering the button

## Related Image
Normal Button
Screenshot from 2020-03-28 15-59-32

Hover button
Screenshot from 2020-03-28 15-59-36

## Additional Information
I would like to work on it

shakti97:
Fixes #2655 

After fix
Screenshot from 2020-04-05 18-33-39

### DIFF
--- a/web/src/components/pages/datasets/dataset-info.css
+++ b/web/src/components/pages/datasets/dataset-info.css
@@ -244,6 +244,12 @@
             & span {
                 margin-inline-end: 20px;
             }
+            &:hover {
+                & path {
+                    fill: var(--black);
+                }
+            }
+            
         }
 
         & .why-email {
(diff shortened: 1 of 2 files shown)


phirework:
This hardcodes a colour value to the cloud icon, which affects everywhere it's used, not just the Dataset download page, and also makes it less flexible to work with in the future. Please use CSS `:hover` instead.

shakti97:
Thanks for the review , will do the recommended changes and update you :)

shakti97:
Added the recommended changes , can you please review again and suggest changes :)

phirework:
Thanks for making the changes!

PR #2669, opened 2020-04-07
Do not deploy on localization commit.

The-smooth-operator:
We do not want to deploy when there is a new commit coming from Pontoon. This will keep our deployment history cleaner


phirework:
Wonderful.

PR #2672, opened 2020-04-09
Add redirect to maintenance mode based on env var

phirework:
After discussion with Alberto, we decided it would be best if the app was still serving the maintenance page to avoid delays in DNS propagation or modifying load balancers to point to different S3 buckets. This commit: 

* Adds an env var for maintenance mode
* Sends all traffic to `/maintenance` if env var is set
  * This LOOKS like a big change, but all it does is move `staticOptions` up to be declared right away, add an `if` block, and move all other `app.use` cases into the `else` block in the same order.
* Adjusts the relative URLs in `/maintenance` to match the new serving location
* Removes `index.html` and adds to `.gitignore` as discussed in previous commit because now this will be built every time on deploy


phirework:
@The-smooth-operator FYI

.gitignore
```diff
@@ -1,6 +1,7 @@
 node_modules
 config.json
 newrelic_agent.log
+maintenance/index.html
```
rileyjshaw, line 4: Nice, thanks!

The-smooth-operator:
This is great :heart:  Thank you for implementing this! It will make smoother our migration, and help us in the future for announcing users that we are having issues.

PR #2679, opened 2020-04-14
Add a note about Docker's memory settings

rileyjshaw:
Auto-rebuilding from `/web` has been broken on my machine for the past
week. I had time to look into it this afternoon. Turns out Docker was
running out of memory and crashing with a cryptic error.

I was already allocating 2GB to Docker, so hopefully we can continue to
reduce the memory footprint of `voice-web`. Until then, I thought this
would be a useful note for folks running the app for the first time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,6 +52,8 @@ This is going to:
 
 You can visit the website at [http://localhost:9000](http://localhost:9000).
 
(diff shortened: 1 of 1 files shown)


phirework:
Is it possible that this has to do with the addition of the gulp task for `/maintenance`?

rileyjshaw:
@phirework it's possible – I haven't looked into the specific memory hog yet. But iirc this has been happening since the code sharing reorganization work.

PR #2680, opened 2020-04-14
Enable audio recording in iOS Safari

rileyjshaw:
This PR enables audio recording (and immediate playback, re-recording, etc) in iOS Safari.

It's split into a few commits, which may be useful to look at individually. There are three major parts to the PR:

1. Conditionally add audio-recorder-polyfill, which adds a functioning `MediaRecorder` class to iOS Safari
2. Retool some of our web audio code to work around iOS gotchas
3. Misc cleanup and refactoring, either shaken loose by iOS additions or noticed along the way

This PR also adds descriptive error states for iOS third-party browsers. Users should be notified that they need to use iOS Safari. Screenshots are in Jira: https://jira.mozilla.com/browse/OI-1499.


rileyjshaw:
Note: error states are presently only visible on iOS Chrome. An update for iOS Firefox, Firefox Focus, etc. is coming soon.

IMG_5A9EBAAB74C5-1

web/src/components/pages/contribution/speak/audio-web.ts
```diff
@@ -1,6 +1,6 @@
-import { isNativeIOS } from '../../../../utility';
+import { isNativeIOS, isIOS } from '../../../../utility';
 
-const AUDIO_TYPE = 'audio/ogg; codecs=opus';
+const AUDIO_TYPE = `audio/${isIOS() ? 'wav' : 'ogg'}; codecs=opus`;
```
rileyjshaw, line 3: Note: as [detailed in `audio-recorder-polyfill`'s documentation](https://github.com/ai/audio-recorder-polyfill#custom-encoder), it's possible to use a custom encoder to enable `ogg` encoding on iOS. I think using `wav` here is more straightforward, and simplifies immediate playback on the device. And we transcode the audio to `mp3` on the server anyway.

If you have any concerns or would prefer we keep everything as `ogg` client-side, let me know and I'll look into it!

web/locales/en/messages.ftl
```diff
@@ -643,8 +643,8 @@ contribute-more =
     }
 record-cta = Start recording
 record-platform-not-supported = We’re sorry, your platform is not currently supported.
-record-platform-not-supported-desktop = On desktop devices you can contribute by downloading...
-record-platform-not-supported-ios = <bold>iOS</bold> users can download our free app:
+record-platform-not-supported-desktop = On desktop devices you can contribute by downloading…
```
phirework, line 646: lmao @ ellipses

web/src/components/pages/contribution/speak/audio-web.ts
phirework: I'm curious about if there's a filesize difference in mp3s generated from wav vs. ogg, because my main concern would be if wav is significantly bigger because it compresses less, the mp3 will be too... There's not really an easy way of testing this, except maybe by recording clips on iOS from a specific account on staging and then examining the subsequent clips in AWS? Thoughts?

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -62,20 +59,41 @@ enum RecordingError {
...
-        <a rel="noopener noreferrer" target="_blank" href="https://www.firefox.com/" title="Firefox">
-          <FirefoxColor />
-        </a>{' '}
-        <a rel="noopener noreferrer" target="_blank" href="https://www.google.com/chrome" title="Chrome">
-          <ChromeColor />
-        </a>
-      </div>
+      {isIOS() && !isMobileSafari() ? (
+        <>
+          <BalanceText>
```
phirework, line 64: This is cool, do you know if this supports RTL languages? If yes, we should use this for the overall sentence display cards too.

web/src/utility.ts
```diff
@@ -41,46 +41,24 @@ export function isNativeIOS(): boolean {
...
  *
  * The logic is collected from answers to this SO question: https://stackoverflow.com/q/3007480
  */
-export function isMobileWebkit(): boolean {
+export function isMobileSafari(): boolean {
   return (
     isIOS() &&
-    isWebkit() &&
-    !/(Chrome|CriOS|OPiOS)/.test(window.navigator.userAgent)
+    !window.navigator.standalone &&
```
phirework, line 59: Why does it need to be out of standalone mode?

web/src/components/pages/contribution/speak/audio-web.ts
```diff
@@ -183,15 +188,30 @@ export default class AudioWeb {
 
     return new Promise<void>((res: Function, rej: Function) => {
       this.chunks = [];
-      this.recorder.ondataavailable = (e: BlobEvent) => {
-        this.chunks.push(e.data);
-      };
-
-      this.recorder.onstart = (e: Event) => {
+      // Remove the old listeners.
+      this.recorder.removeEventListener('start', this.recorderListeners.start);
```
phirework, line 192: Just out of curiosity, can you explain why you needed all the events to be on a new object instead of `this.recorder`? I skimmed the polyfill documentation but nothing jumped out at me and I'm sure there's a good reason for it.

web/src/utility.ts
rileyjshaw: Mic support in standalone mode is still sketchy on iOS 😩 even though third-party browsers and standalone mode are running the exact same browser under the hood, the Safari app is the only environment that works consistently across all the APIs we use.

web/src/components/pages/contribution/speak/audio-web.ts
rileyjshaw: **TL;DR,** because:

1. `this.recorder` isn’t an object signature that we created or control.
2. There’s no standard DOM method to access a node’s event listeners.
3. The polyfill expects us to use `add/removeEventListener()` rather than setting `onStart` directly.

I think it’s cleaner to create a state object rather than add nonstandard properties (eg. `_listeners`) to a `MediaRecorder` instance.

---

Good question! This isn't exactly polyfill-specific, so some example code might help clarify:

```.js
// Goal: Add a click listener, then remove it a bit later.
const el = document.createElement('div');

////////////////////////////////
// Easy mode: shared closure. //
////////////////////////////////
addAndRemoveHandler('clicked');

function addAndRemoveHandler (msg) {
	const handler = () => alert(msg);
	el.addEventListener('click', handler);

	// Remove the event listener after 5 seconds. We can do this because we
	// still have access to the initial handler within this scope.
	window.setTimeout(() => {
		el.removeEventListener('click', handler);
	}, 5000);
}

///////////////////////////////////
// Hard mode: separate closures. //
///////////////////////////////////
addHandler('clicked');
window.setTimeout(removeHandler, 5000);

function addHandler (msg) {
	const handler = () => alert(msg);
	el.addEventListener('click', handler);
}

function removeHandler () {
	el.removeEventListener('click', /* uh oh… we can’t access
	                                   the handler from here. */);
}
```

There are many ways around this problem. Some common solutions are:

1. Easiest, but a bit of a hack: avoid `add/removeEventListener()`. Instead, set the `onClick` (or `onStart`, etc) property directly. This overwrites any listeners that existed beforehand.
2. Still pretty easy, not _e l e g a n t:_ create a state object in the parent scope to keep track of active handlers. That way, you can still reference the handler when it’s time to remove it.
3. Fancy: pass everything around as arguments and returns, don't track state at the parent level.

We’ve gone with option 1 until now by setting `onStart` directly. Buuuut, the polyfill is just using a `documentFragment` internally; it knows how to handle `add/removeEventListener()`, but it doesn’t know what to do with `onStart`.

Which leaves us with option 2 or 3. Option 3 is nice in theory, but would have been a ridiculous tangle with our current implementation.

So, option 2 it is. And I still haven’t even answered your question yet.

We could add a custom property to `this.recorder`, but it’s similar to adding a custom `_listeners` property to `el` in the example above. If browser vendors or the polyfill author later decide to add a `_listeners` property to the `MediaRecorder` instance, it will cause a bug. So, keeping track of state in our own separate object is safer (and hopefully somewhat clearer).

web/src/components/pages/contribution/speak/speak.tsx
rileyjshaw: I wrote a quick custom test with `direction: rtl` on [the library's storybook](https://khan.github.io/react-balance-text/?selectedKind=BalanceText&selectedStory=Basic%20Text%20Balancing&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel) and it seemed to work 👍

web/src/components/pages/contribution/speak/audio-web.ts
rileyjshaw: Hmm, good question. We can also use `mp3`s directly here, the polyfill comes bundled with an encoder. Do you think that would be better?

rileyjshaw:
Note: just pushed [an update](https://github.com/mozilla/voice-web/pull/2680/commits/f831def84b420e8ace1296b3c5fefadc7ffb16b4) to limit this to iOS only to reduce our testing surface area.

web/src/components/pages/contribution/speak/audio-web.ts
phirework: Awesome, thanks for the clear explanation!

phirework: Awesome, thanks for the clear explanation!

web/src/components/pages/contribution/speak/audio-web.ts
rileyjshaw: We discussed this offline and are going to stick with .wav for now 👍

rileyjshaw:
Update: we were seeing CSP errors due to some recent changes. I've updated our policy in the most recent commit, so the PR should be testable again on voice.allizom.org.

> Note: just pushed [an update](https://github.com/mozilla/voice-web/pull/2680/commits/f831def84b420e8ace1296b3c5fefadc7ffb16b4) to limit this to iOS only to reduce our testing surface area.

I removed this restriction after a bit of testing. I'm seeing promising results on desktop Safari after the CSP fix, but the real clincher was that iPad Safari defaults to the same user agent as desktop Safari!

rileyjshaw:
Okay – just pushed the final rebase! Eeeee :)

Waiting for Travis, then I'll land this. Very exciting!

mikehenrty:
Hey @Gregoor , look at what they did! :1st_place_medal:

Gregoor:
Yaya awesome, that'll improve my sleep

PR #2686, opened 2020-04-19
fix(redundant faq search button)

### LINKED ISSUE #2640: No Need for Search Button
## Description
We are searching and showing result as soon as the user change the search input on FAQ, so there will be no use of search button on the right of the search box, we might just use that for the ui, but we can remove the handler from the search button

## Current Behaviour
We are showing search result as soon as the user start typing so there will be no need of search button, but user can click on the search button and search button doesn't do anything because the result is already showing

## Expected Behaviour
We should only use that search icon in ui and remove handler from the button

## Additional Information
I would like to work on it

shakti97:
Fixes #2640

### DIFF
--- a/web/src/components/pages/faq/faq.tsx
+++ b/web/src/components/pages/faq/faq.tsx
@@ -165,14 +165,13 @@ export default withLocalization(({ getString }: LocalizationProps) => {
                     setSearchString(event.target.value);
                   }}
                   placeholder={label}
-                  required
                 />
               )}
             </LocalizedGetAttribute>
 
-            <button type="submit" className="search-button">
+            <div className="search-button">
               <SearchIconCode className="icon" />
-            </button>
+            </div>
           </form>
 
           <div className="faq-section-container">
--- a/web/src/components/pages/faq/faq.css
+++ b/web/src/components/pages/faq/faq.css
@@ -489,7 +489,7 @@
             }
 
             & input,
-            & button {
+            & div {
                 height: 80px;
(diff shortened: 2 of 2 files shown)


web/src/components/pages/faq/faq.tsx
```diff
@@ -165,14 +165,13 @@ export default withLocalization(({ getString }: LocalizationProps) => {
                     setSearchString(event.target.value);
                   }}
                   placeholder={label}
-                  required
                 />
               )}
             </LocalizedGetAttribute>
 
-            <button type="submit" className="search-button">
+            <div className="search-button">
```
rileyjshaw, line 172: Rather than replace the button with a div, I'd prefer if the button was connected to the same action that takes place `onChange`. I agree that it's redundant, but the button is a good UI hint and this gives us a fallback in case `onChange` doesn't fire for some reason.

shakti97: Thanks, @rileyjshaw for the review. But if onChange doesn't work for some reason then the value of the input would never change as input value is controlled by the state variable which will be updated only on change, and in that case, if you try to call the same function with the value of the input you won't be able to do that because the value is not changed.

So I am thinking of converting this input  to be uncontrolled, then use ref to get the value of the input and use it at the time of `submit`

shakti97: Also, could you highlight some use cases where the `onChange` handler might fail to function?

PR #2689, opened 2020-04-22
Move Sentence and Clip types into common folder

phirework:
The `Sentence` interface was being separately defined in both `web` and `server`, this just pulls that as well as `Clip` into the 'common' module for easier sharing.


rileyjshaw:
Aaaah, amazing!!! I hope you know how happy this makes me 💖

PR #2695, opened 2020-04-27
Create mapping between legal locales and on-site locales

### LINKED ISSUE #2694: Legal documentation: add new languages, update locale codes to language codes
@mbransn we have added the following languages to the legal documents (Privacy and Terms). I also took the opportunity to update the locale code to language code that are good to change based on the [recent discussions](https://github.com/mozilla/voice-web/issues/2550#issuecomment-608112427) with the CV team.  Here is the summary, please make necessary changes on the code side accordingly:
- ar
- as
- cs
- el
- **fy** (directs to fy-NL)
- **ga** (directs to ga-IE)
- **hsb** (direct to de  - the German legal docs)
- id 
- ja
- lv
- mt
- **pa** (directs to pa-IN)
- pl
- **pt** (formerly known as pt-BR, we now have both legal docs)
- **rm** (Romansh for both Romansh Sursilvan and Romansh Vallader)
- ro
- sl
- uk

phirework:
Fixes #2694

### DIFF
--- a/server/src/fetch-legal-document.ts
+++ b/server/src/fetch-legal-document.ts
@@ -8,6 +8,16 @@ const cache: {
   };
 } = {};
 
+const localeMapping: { [key: string]: string } = {
+  'fy-NL': 'fy',
+  'ga-IE': 'ga',
+  hsb: 'de',
+  'pa-IN': 'pa',
+  'rm-sursilv': 'rm',
+  'rm-vallader': 'rm',
+  'sv-SE': 'se',
+};
+
 export default async function fetchLegalDocument(
   name: string,
   locale: string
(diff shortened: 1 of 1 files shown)


server/src/fetch-legal-document.ts
```diff
@@ -19,9 +29,10 @@ export default async function fetchLegalDocument(
   if (textHTML && fetchedAt > Date.now() - CACHE_AGE) {
     return textHTML;
   }
+  const legalLocale = localeMapping[locale] ? localeMapping[locale] : locale;
```
rileyjshaw, line 32: Absolutely not a blocker, but FYI Typescript now supports [Nullish Coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing)!

```typescript
// Before:
  const legalLocale = localeMapping[locale] ? localeMapping[locale] : locale;

// Functionally equivalent:
  const legalLocale = localeMapping[locale] || locale;

// Only falls back to default if localeMapping[locale] is null or undefined:
  const legalLocale = localeMapping[locale] ?? locale;
```

phirework: Very cool! Pushed the change to master directly, thanks for the flag.

PR #2696, opened 2020-04-28
Update major dev dependencies

rileyjshaw:
**Typescript: 3.7 -> 3.8**

No breaking changes here, the most interesting feature looks like [type only imports and exports]( https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#-type-only-imports-and-export).

I removed the duplicate TS deps from our `/web` and `/server` packages since we're trying to keep those versions locked.

**Prettier: 1 -> 2**

Major version, ooooouuuu. To support the new features added in TS 3.8, we need to upgrade to Prettier 2. The [breaking changes are listed here](https://prettier.io/blog/2020/03/21/2.0.0.html#breaking-changes), which I've dealt with in the following ways:

- Change default value for trailingComma to es5

We were already setting `trailingComma` to `es5` in `.prettierrc`, so I removed the redundant declaration.

- Change default value for arrowParens to always

The documentation makes some convincing arguments for this, and I'd be happy to flip the flag in the future. However, this change alone nearly doubled the surface area of files affected by this commit, so I decided to add the old default to `.prettierrc` for the time being.

- Drop support for Node versions older than 10

Good thing we [updated Node](https://github.com/mozilla/voice-web/commit/62084c715e29b524c4053b5d9fdede587da277fb)! :)

- Change default value for endOfLine to lf
- Fix config overrides pattern matching to include dot files
- Plugin API: changes in prettier.util
- Cache plugin search results
- Remove deprecated options and option values
- Remove the version parameter of prettier.getSupportInfo

No changes required.

---

There are a few other libraries that are behind a major version, but this change seems big enough. I'll try to get a follow-on diff out later in this sprint once this code settles :)

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,5 @@
 {
   "jsxBracketSameLine": true,
   "singleQuote": true,
-  "trailingComma": "es5"
+  "arrowParens": "avoid"
 }
--- a/package.json
+++ b/package.json
@@ -13,9 +13,9 @@
   "devDependencies": {
     "concurrently": "^5.1.0",
     "husky": "^4.2.1",
-    "prettier": "^1.19.1",
+    "prettier": "^2.0.5",
     "pretty-quick": "^2.0.1",
-    "typescript": "3.7.5"
+    "typescript": "3.8.3"
   },
   "workspaces": [
     "common",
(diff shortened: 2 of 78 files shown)


phirework:
This looks great, thanks for double-checking all the breaking changes!

PR #2698, opened 2020-04-29
missing text added

Angelblue1:
missing text added


phirework:
Hello! Thanks for taking the time to submit this PR. All localization changes need to be made through through Mozilla's localization platform Pontoon, otherwise they will be overwritten next time we sync from there. The page for Turkish can be found here: 

https://pontoon.mozilla.org/tr/common-voice/web/locales/en/messages.ftl/

PR #2703, opened 2020-04-30
Delete references to Nubis

The-smooth-operator:
We are not using Nubis anymore. Let's clean up references to it.


phirework:
You're the best! I was just thinking I needed to strip this stuff out last night.

phirework:
Actually, sorry, I should've realized this before now - what is controlling which tables and fields are synced to ES? Previously it was here: https://github.com/mozilla/voice-web/blob/f82074b0f181ee182ed642083e72c059ea6a1a6f/nubis/puppet/files/confd/templates/river.toml.tmpl but I assume you've moved that to K8s, so now I don't know where to go to look for / modify that.

The-smooth-operator:
The Kubernetes objects defining the es-sync process are here: https://github.com/mozilla-it/helm-charts/blob/master/charts/mozilla-common-voice/templates/es-sync.yaml
The configuration file is a secret because it contains credentials for accessing the database, and they can't be decoupled from the file. So that means that for now, the configuration specifying which tables to use was a secret manually created by me. You can check it in the cluster, prod or stage namespace with: `kubectl get -o yaml secret es-sync`.
Let's discuss in our next meeting how to better handle this file.

PR #2708, opened 2020-05-03
Fix empty value check

shakti97:
Its accepting space to bypass the check, if someone put space in it the button gets enabled and the user can submit the report


phirework:
Thanks!

PR #2723, opened 2020-05-14
fix some grammar

Rexogamer:
I added a missing ' and changed some wording to be a bit clearer


phirework:
Hi there - thanks for this! The language we use in `email-already-used` was actually suggested by our copy-editor, and we'd prefer it to stay the way it is. Also, for future record, if you make any updates to localization strings, you need to create a whole new id for it (e.g. `email-already-used-v2`) because otherwise they will not show up in Pontoon to be re-translated into other languages, and this is such a minor change that we normally wouldn't bother setting off the whole retranslation process anyway. 

If you'd like to undo that change and leave in the missing `'` I can merge this in. Thanks!

phirework:
Thank you!

PR #2724, opened 2020-05-14
Enable collection for target segments

phirework:
75% of this PR is changes you've already seen in the [last one](https://github.com/mozilla/voice-web/pull/2701/files/21c8eeef326c6bfaa90187a7d8fbd8dc25d58723). The big new pieces are: 

* an API endpoint for the feature flag so that the frontend can know about ENV var changes
* the frontend banner view
* makes some changes to db queries and database schema to improve performance
* a few additional words to accommodate a FxVoice request from Jofish

I've left inline comments as much as possible, sorry for the large PR 😬


server/src/lib/api.ts
```diff
@@ -116,13 +117,39 @@ export default class API {
...
+  ) => {
+    let featureResult = null;
+
+    try {
+      const featureToken = feature as FeatureToken;
+      if (featureTokens.includes(featureToken)) {
+        const featureTemp = features[featureToken];
+        if (
+          (!featureTemp.locales || featureTemp.locales.includes(locale)) &&
+          getConfig()[featureTemp.configFlag as keyof CommonVoiceConfig]
```
phirework, line 141: This is ugly, but without the `as keyof` typescript throws a bunch of `implicit any` errors, and I wanted to avoid modifying `CommonVoiceConfig` itself as much as possible since it's used in so many places. If you know of a different way of accessing a variable key, please do let me know!

server/src/lib/model/db.ts
```diff
@@ -170,22 +218,103 @@ export default class DB {
             FROM clips
             WHERE clips.original_sentence_id = sentences.id AND
                   clips.client_id = ?
+          ) AND NOT EXISTS (
+            SELECT *
+            FROM skipped_sentences skipped
+            WHERE skipped.sentence_id = sentences.id AND
+              skipped.client_id = ?
```
phirework, line 225: As discussed, added skipped_sentence handling to the regular sentence serving logic too.

server/src/lib/model/db.ts
```diff
@@ -194,26 +323,86 @@ export default class DB {
...
-                FROM votes
-                WHERE votes.clip_id = clips.id AND client_id = ?
-              )
+        WHERE is_valid IS NULL AND clips.locale_id = ? AND client_id <> ?
+        AND NOT EXISTS(
+            SELECT *
+            FROM votes
+            WHERE votes.clip_id = clips.id AND client_id = ?
+          )
+        AND (sentences.clips_count <= 10 OR sentences.clips_count IS NULL)
```
phirework, line 332: I added this filter here because I noticed during testing that finding clips was pretty slow if you're a user who has already voted on a lot of things (1-2s when you hit the db directly), and this very quickly narrows the field of clips, especially since all of our other logic already prioritizes clips with low clip_counts and the likelihood of a sentences increasing past that is minimal. 

There are some sentences that have a clip count above 10 where none of the clips are valid, but they tend to be super old clips from before that other logic was implemented. It'll be moot soon anyway once we implement single-sentence record limits.

server/src/lib/model/db.ts
```diff
@@ -194,26 +323,86 @@ export default class DB {
...
+              LEFT JOIN taxonomy_entries entries
+                ON clips.original_sentence_id = entries.sentence_id
+              WHERE clips.locale_id = ?
+              AND (votes.client_id = ?)
+              AND entries.term_id = ?
+              GROUP BY original_sentence_id
+              HAVING vote_count >= 2
+            ) vote_counts
+          )
+        )
```
phirework, line 374: This is the biggest change in actual logic compared to the first PR, and it's also because during testing this query ground to a halt if you had already voted hundreds of thousands of times, or had tens of thousands of clips. The logic is kind of backwards - it looks for a list of sentences that match the taxonomy that the user has already voted twice on, and removes it from a full list of sentences matching taxonomy. Because it's primarily selecting for taxonomy_entries though, it's super fast.

server/src/lib/model/db/migrations/20200422152745-cascade-skipped-sentence-deletion.ts
```diff
@@ -0,0 +1,12 @@
+export const up = async function (db: any): Promise<any> {
+  return db.runSql(
+    `
+    ALTER TABLE skipped_sentences DROP FOREIGN KEY skipped_sentences_ibfk_1;
+    ALTER TABLE skipped_sentences ADD CONSTRAINT skipped_sentences_ibfk_1 FOREIGN KEY (sentence_id) REFERENCES sentences (id) ON DELETE CASCADE;
```
phirework, line 5: I'm not sure if this is strictly necessary, but in `import-sentences` we do a bulk deletion of unused sentences that don't match a current version and I thought it was possible that that could deadlock at some point if we try to delete a sentence that's referenced in `skipped_sentences` (but not in anything else, where it wouldn't affect `is_used`).

web/src/components/layout/layout.tsx
```diff
@@ -31,7 +33,13 @@ import {
   challengeTeamTokens,
   ChallengeToken,
   challengeTokens,
+  FeatureType,
+  FeatureToken,
+  features,
 } from 'common';
+import API from '../../services/api';
+import NotificationBanner from './../notification-banner/notification-banner';
+import { Notifications } from '../../stores/notifications';
```
phirework, line 42: I'm not using the Notification.add actions, because those notifications are not really meant to be persistent, and they are identified by ID, so every time you clicked into a page that used layout.tsx, went to speak.tsx, and came back there would be another banner, and no way to check for uniqueness (as far as I could tell) because the banner IDs just auto-increment. This solution is kind of wonky, but I think is the lowest overhead solution. I think there might be value in separating banners entirely from notifications, but I wasn't sure if at some point in the future we may want temporary banners?

web/src/stores/sentences.ts
```diff
@@ -3,7 +3,8 @@ const contributableLocales = require('../../../locales/contributable.json') as s
 import StateTree from './tree';
 import { Sentence } from 'common';
 
-const CACHE_SET_COUNT = 10;
+const CACHE_SET_COUNT = 25;
```
phirework, line 6: I ended up increasing the cache_count to 25, because after you submitted the first 5 sentences, the subsequent query would try to get another 5 sentences, find 7 unread target-segment sentences, and then append them, so that you end up with repeats anyway.

web/src/components/notification-banner/notification-banner.css
```diff
@@ -1,70 +1,119 @@
...
+        position: fixed;
+        top: auto;
+        bottom: 0;
     }
 
-    & .spacer {
-        margin-right: auto;
+    .notification-text {
+        text-align: center;
+        color: #f89096;
```
rileyjshaw, line 29: This is used a bunch in our codebase, we should move it into a variable.

web/src/components/notification-banner/notification-banner.tsx
```diff
@@ -12,20 +12,30 @@ export const Banner = React.forwardRef(
...
     } & React.HTMLProps<HTMLDivElement>,
     ref: any
   ) => (
     <div ref={ref} className={'banner ' + className} {...props}>
-      <div className="spacer" />
-      <h1>{children}</h1>
-      <LinkButton {...ctaButtonProps} className="cta" onClick={onClose} />
+      <h2 className="notification-text">{children}</h2>
+      {bannerProps.links.map((cta: any, key: number) => {
+        const persistafterclick = cta.persistafterclick;
```
rileyjshaw, line 28: Nit: can we change this (and all related) to `persistAfterClick`?

web/src/components/notification-banner/notification-banner.tsx
```diff
@@ -42,28 +52,27 @@ export default function NotificationBanner({
   const [show, setShow] = useState(false);
 
   const el = useRef(null);
+  const { storageKey = '' } =
+    notification.kind == 'banner' && notification.bannerProps;
```
rileyjshaw, line 56: Destructuring here is a bit weird, because you might end up checking for `false.storageKey`. I think:

```js
  const storageKey = notification.kind === 'banner' ? notification.bannerProps : '';
```

could work instead?

web/src/components/notification-banner/notification-banner.tsx
```diff
@@ -42,28 +52,27 @@ export default function NotificationBanner({
...
 
   useEffect(() => {
     setShow(true);
   }, []);
```
rileyjshaw, line 60: I'm confused about what's happening here. It looks like we set the initial state to `false`, then immediately cause a rerender by updating it to `true`? Do you know why we're not just setting `const [show, setShow] = useState(true);` at the outset?

web/src/components/pages/contribution/contribution.css
```diff
@@ -593,7 +641,7 @@
             }
         }
 
-        & kbd {
+        kbd {
```
rileyjshaw, line 644: Annnnnd today I learned there's a `<kbd>` element, cool!

web/src/components/notification-banner/notification-banner.tsx
```diff
@@ -42,28 +52,27 @@ export default function NotificationBanner({
...
-              className: 'cta',
-            }
-          : {}
-      }
-      onClose={() => setShow(false)}
+      bannerProps={notification.kind == 'banner' && notification.bannerProps}
+      onClose={() => hideBanner(storageKey)}
       className="notification-banner"
-      style={{ transform: `translateY(${show ? 0 : -100}%)` }}
+      style={{ display: `${show ? 'flex' : 'none'}` }}
```
rileyjshaw, line 73: Could we just not render it at all if `show` is false, rather than returning the component with `display: none`?

```jsx
return show && <Banner …
```

web/src/components/notification-banner/notification-banner.tsx
```diff
@@ -42,28 +52,27 @@ export default function NotificationBanner({
...
+  const { storageKey = '' } =
+    notification.kind == 'banner' && notification.bannerProps;
 
   useEffect(() => {
     setShow(true);
   }, []);
 
+  function hideBanner(storageKey?: string) {
+    setShow(false);
+    storageKey && localStorage.setItem(storageKey, JSON.stringify(true));
```
rileyjshaw, line 64: Not blocking, but `localStorage` access should be behind a `useEffect`. Check out https://joshwcomeau.com/react/persisting-react-state-in-localstorage/ for more info

web/src/components/layout/layout.tsx
```diff
@@ -194,6 +272,9 @@ class Layout extends React.PureComponent<LayoutProps, LayoutState> {
             teamToken={challengeTeamToken}
           />
         )}
+        {targetSegment && !localStorage.getItem(targetSegment.storageKey) && (
```
rileyjshaw, line 275: Commented on this below, but we might wanna use this `useStickyState` hook: https://joshwcomeau.com/react/persisting-react-state-in-localstorage/

web/src/components/layout/layout.tsx
rileyjshaw: I think separating banners from notifications makes sense. This is a rough patch in our code already, so would you mind leaving a TODO (or Jira ticket or something) to capture that there's work to be done here?

server/src/lib/api.ts
phirework: It does if I create another type called `FeatureFlag` or something and directly specify `BENCHMARK_LIVE` as the string literal, but that also felt ugly because there would be a set of string literals for `FeatureToken`, another corresponding set for `FeatureFlag`, and then the `FeatureType` that combines the two sets in a way where if you want to make changes you'd have to update several places. 

Instead of specifying `BENCHMARK_LIVE` in `CommonVoiceConfig`, I did consider something like:

```
FEATURE_FLAGS: {
   [key: string]: boolean;
}
```

But then the problem was to cleanly set default values for an unknown set of potential env vars. None of these options are particularly great.

server/src/lib/api.ts
```diff
@@ -116,13 +117,39 @@ export default class API {
...
   }
 
+  getFeatureFlag = (
+    { params: { locale, feature } }: Request,
+    response: Response
+  ) => {
+    let featureResult = null;
+
+    try {
+      const featureToken = feature as FeatureToken;
```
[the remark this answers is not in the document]
phirework: I don't think you can type the argument signature here because `feature` is part of the `Request` object and its parameters only returns `string: string` key-value pairs.

common/feature.ts
```diff
@@ -0,0 +1,37 @@
...
+      'ta',
+      'tr',
+      'tt',
+    ],
+  },
+};
+
+export const featureTokens = Object.values(features).map(
+  feature => feature.token
+);
```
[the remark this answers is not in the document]
phirework: I... feel like there was a good reason I did this, but now I don't remember what it is ?_?

web/src/components/notification-banner/notification-banner.tsx
phirework: TIL! Thank you!

web/src/components/notification-banner/notification-banner.tsx
phirework: Hmm, I think it's probably a hold-over from using banners as notifications.

web/src/components/layout/layout.tsx
phirework: I added the custom hook to our stores and also to `notification-banner`, but I'm not going to bother with Layout - it's a class component and it seemed kind of overkill to wrap it in an HOC just for this (esp since it's already wrapped in `withRouter`).

PR #2725, opened 2020-05-15
Change minimum volume handling code for /speak

rileyjshaw:
We've had some reports of loud clips being flagged as silent (eg.
https://discourse.mozilla.org/t/cant-submit-recordings-because-im-told-they-are-too-quiet-even-though-theyre-not/59373).

This commit ensures that a low volume error is thrown only if our
volume-checking utility is working and has been called. For various
reasons, some platforms may never receive an `onaudioprocess` event. In
such cases, we shouldn't assume that the recording is silent. I’ve added
notes to the server to check for silent clips.

This commit also removes any mention of "visualizer" from `audio-web.ts`,
since our visualizer is not audio-reactive.

### DIFF
--- a/web/src/components/pages/contribution/speak/audio-web.ts
+++ b/web/src/components/pages/contribution/speak/audio-web.ts
@@ -37,7 +37,7 @@ export default class AudioWeb {
       dataavailable: null,
       stop: null,
     };
-    this.visualize = this.visualize.bind(this);
+    this.analyze = this.analyze.bind(this);
   }
 
   private isReady(): boolean {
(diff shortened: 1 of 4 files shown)


web/src/components/pages/contribution/speak/audio-web.ts
```diff
@@ -93,29 +93,14 @@ export default class AudioWeb {
...
-    let sum = 0;
-    for (var i = 0; i < this.frequencyBins.length; i++) {
-      sum += this.frequencyBins[i];
-    }
-
-    let average = sum / this.frequencyBins.length;
-
     if (this.volumeCallback) {
-      this.volumeCallback(average);
+      this.volumeCallback(Math.max(...this.frequencyBins));
```
rileyjshaw, line 100: Getting the average volume of a frequency spectrum is pretty complicated, especially when the data represents logarithmic data on a linear scale. And the old code _definitely_ didn't do what it said it did 😅 I think finding the max dB value is actually a closer measure to the signal's power, so I changed it here.

phirework:
Hmm I would argue that the visualizer is still valuable even though it's not audio-reactive, because it's a visual signal that the microphone is active. Like, if your browser is asking for permission for the mic, or sluggish in enabling it, the visualizer not working is a good signal that the mic isn't ready. Thoughts?

rileyjshaw:
@phirework sorry for the confusion – this commit doesn't remove our visualizer, or any visualization! We had some methods in `audio-web.ts` called `startVisualize` etc. which had nothing to do with visualization, so I renamed 'em. That's all :)

PR #2728, opened 2020-05-16
Add files via upload

Isomorph70:
6448 sentences extracted from europarl-v7.da-en.da


Isomorph70:
Update to the first pull request, because of a fix in extraction script.

PR #2731, opened 2020-05-18
layout/demo-routes

NanaKay007:


NanaKay007:
Added a new layout (demo_layout.js) for the demo workflow. Contains routes which render temporary dummy components

web/src/components/app.tsx
```diff
@@ -280,7 +281,11 @@ let LocalizedPage: any = class extends React.Component<
                   }
                 />
               ))}
-              <Layout />
+              {location.pathname.includes('/demo') ? (
```
phirework, line 284: I see that you've added the `/demo` path to the `URLS` constant, can you replace these strings with a reference to `URLS`?

web/src/components/layout/demo_layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+          <Route
+            exact
+            path={toLocaleRoute('/demo')}
+            render={() => {
+              return <h1>in get started</h1>;
+            }}
+          />
+          <Route
+            exact
+            path={toLocaleRoute('/demo/language-select')}
```
phirework, line 22: Same as above re: referencing `URLS` instead of manually writing out the route.

web/src/components/layout/demo_layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Switch, Route, withRouter } from 'react-router';
+import { useLocale } from '../locale-helpers';
+import { Spinner } from '../ui/ui';
+
+function DemoLayout() {
+  const [_, toLocaleRoute] = useLocale();
+
+  return (
+    <div>
+      <React.Suspense fallback={<Spinner />}>
```
phirework, line 11: Nitpick - I know it's used this way elsewhere in the codebase, but I tend to think it's cleaner to separately import the Suspense Component and then just use it as `<Suspense>`

web/src/components/layout/demo_layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+          />
+          <Route
+            exact
+            path={toLocaleRoute('/demo/language-select')}
+            render={() => {
+              return <h1>in language select</h1>;
+            }}
+          />
+          {/* more routes to be added */}
+          <Route path="*" render={() => <h1>not found</h1>} />
```
phirework, line 28: The last `<Route>` in a `<Switch>` will always be used as the fallback if nothing else matches, so `path="*"` is not needed here. Though I would suggest redirecting this to just the home `/demo` if nothing else is found.

NanaKay007:
@phirework 
Made the changes you requested :)

phirework:
You're missing the `/demo` route in `app.tsx`, otherwise you're good to go!

NanaKay007:
all done! @phirework

PR #2736, opened 2020-05-21
layout/intro-page

NanaKay007:


web/src/components/demo_pages/Intro/intro.css
```diff
@@ -0,0 +1,91 @@
...
+    background: linear-gradient(180deg, #f3f2f1 0%, #ffffff 101.23%);
+}
+
+.text-box {
+    display: flex;
+    flex-direction: column;
+    padding: 0px;
+    height: 173px;
+    left: calc(50% - 700px / 2 + 1px);
+    bottom: 68px;
```
phirework, line 16: So generally speaking I would avoid specifying absolute pixel values like this. When it comes to heights, you don't know how long strings will be when localized into other languages, so you'll want to account for fluid text lengths in your layout. And since this demo will be available on desktops as well rather than just set tablet sizes, you'll want this to be fluid. While in this case the overflow from the container gets displayed anyway because there's nothing else around/underneath, in more complex layouts the fact that the button pushes beyond the bounds of the text-box at different widths can become a problem. 

I'm also not clear on what the `bottom` value is accomplishing here, or why you're using calculated positioning instead of just centering the text? A thing you'll also want to think about is how this will look in languages that are RTL, so if you're specifying positions based on their relationship to a starting point on the left that can have unintended consequences.

web/src/components/demo_pages/Intro/intro.css
```diff
@@ -0,0 +1,91 @@
...
+}
+
+.text > h1 {
+    font-style: normal;
+    font-weight: normal;
+    font-size: 36px;
```
phirework, line 22: Not a big deal, but this is an example of what I mean about not worrying about being pixel perfect. Figma might specify this as 36px, but our typical largest font is 34px according to `vars`, and I would prefer to err on the side of consistency within the site rather than creating edge cases for every new design.

web/src/components/demo_pages/Intro/intro.css
```diff
@@ -0,0 +1,91 @@
...
+    line-height: 43px;
+    text-align: center;
+}
+
+.text > p {
+    font-style: normal;
+    font-weight: 300;
+    font-size: 24px;
+    line-height: 33px;
```
phirework, line 31: Similarly to my above comments, the font-size here is specified in `vars`, and for the sake of accommodating people who may change their browser's font-size settings it's always a good idea to specify things like line-height in relative/unitless terms (e.g. `1.375` in this case, although if you wanted to call it `1.4` I also would not have any objections.) 

More info on unitless here: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

web/src/components/demo_pages/Intro/intro.tsx
```diff
@@ -0,0 +1,45 @@
...
+            <Localized id="welcome-subheader" />
+          </p>
+          <Link
+            to={toLocaleRoute('/demo/language-select')}
+            className="btn-get-started">
+            <span>
+              <Localized id="get-started" />
+              <img src={arrow} alt="arrow right" />
+            </span>
+          </Link>
```
phirework, line 40: You can use `<LinkButton />` from our UI components instead of manually specifying everything here, and just change the fill colours. This is also a good example of why you'd want `Localized` to wrap around the element, instead of treating it as a string.

web/src/components/demo_pages/Intro/intro.css
```diff
@@ -0,0 +1,91 @@
...
+    display: flex;
+    background: var(--blue);
+    border-radius: 1000px;
+    margin: 30px auto;
+    align-self: center;
+    align-items: center;
+    justify-content: center;
+    color: white;
+    width: 240px;
+    height: 57px;
```
phirework, line 47: Same comments apply re: absolute values and keeping in mind localization changes.

web/src/components/demo_pages/Intro/intro.css
```diff
@@ -0,0 +1,91 @@
...
+    display: flex;
+    justify-content: center;
+    width: 100vw;
+    height: 530px;
+}
+
+.waves {
+    width: 100vw;
+    height: 324px;
+    top: 169px;
```
phirework, line 69: Same comments apply re: absolute values and keeping in mind localization changes

web/src/components/demo_pages/Intro/intro.tsx
NanaKay007: The LinkButton component doesn't seem to render children, hence unable to render the arrow image in the prototype. Do I ignore adding the arrow, or is there something I might be missing?

phirework: You can specify icon objects in `en/messages.ftl` as well, and both `notification-banner.tsx` and `dataset-info.tsx` should have examples of link buttons with an embedded object.

phirework:
This looks great, thanks for making all those changes!

PR #2737, opened 2020-05-23
Remove JSON log outputs

phirework:
Back in the days when Kibana was our main logging viewer we needed JSON logs. Now that we're on Papertrail, JSON logs are very messy and hard to read. Dev/Sandbox have been using the default console.log with no issues, so let's try also doing this on stage. cc @The-smooth-operator 

I'm leaving `lib/logger.ts` for now in case we end up needing it for something else.


rileyjshaw:
Woohoo, this is awesome!

PR #2750, opened 2020-05-28
Add Catalan discourse

nukeador:


nukeador:
@phirework my commit was auto formatted to one line, and I'm not sure if that's ok. Can you check?

Thanks!

phirework:
Yup, that's fine. The most recent upgrade to Prettier (which we use to autoformat code) had some config changes like this of when to split into multiple lines and when to leave on the same.

PR #2764, opened 2020-06-04
Proof of concept: associate demographic metadata with clips

rileyjshaw:
We create clip metadata on the fly for each dataset release, based on the speaker’s demographic info available at that time. This creates a problem for long-term users who need to change their profile’s age, sex, or accents; all previous clips will be updated with new, inaccurate metadata.

This commit is a first step toward addressing this problem. It creates a new `demographics` table to hold the demographic history of our users. Clips can be locked to a specific entry in this table, so we don’t risk changing metadata in the future.

As discussed offline, there are a few inelegant hacks going on here. To avoid locking up our DB with a huge backfill on `clips` and `user_clients`, I’ve created two one-to-one tables:

```
CREATE TABLE user_client_demographics…
CREATE TABLE clip_demographics…
```

These respectively fake an additional `demographics` column on `user_clients` and on `clips`. I don’t like it, and if you have any ideas about how we can backfill our tables without taking the site down I’d be happy to do that instead (or try something else entirely).

Accents are a bit of a pickle, so I’m keeping them out of this commit. Profiles can have multiple languages, so there may be a one-to-many relationship between users and accents. It will probably be easiest to keep track of historical accent data in a separate table, and I thought there was already enough going on here for a proof of concept. If it gets the go-ahead, I’ll start moving on accents.


server/src/lib/model/db/migrate-demographics.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const db = getMySQLInstance();
+
+export async function migrateDemographics() {
+  await db.query(
+    `
+      INSERT INTO demographics (client_id, age, gender)  (
+        SELECT client_id, age, gender
+        FROM user_clients
+      )
```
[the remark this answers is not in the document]
rileyjshaw: Hmm, good point; I’ve added a `FIXME` for now. I want to ensure that the join logic is sound everywhere in case a record isn’t found. I can also see problems with our current `SELECT id FROM demographics WHERE client_id = ? ORDER BY updated_at DESC LIMIT 1` solution, both for folks who don’t have a `demographics` record _and_ for folks who later add one but don’t want it to apply to their prior activity.

Should be easy accommodations to make, and I agree that it would be nice to avoid useless records where possible.

server/src/lib/model/db/migrations/20200603213417-add-demographics-table.ts
```diff
@@ -0,0 +1,37 @@
...
+        FOREIGN KEY (client_id) REFERENCES user_clients (client_id),
+        UNIQUE (client_id, age, gender)
+      );
+
+      CREATE TABLE user_client_demographics (
+        client_id BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
+        demographic_id BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT,
+        FOREIGN KEY (client_id) REFERENCES user_clients (client_id),
+        FOREIGN KEY (demographic_id) REFERENCES demographics (id)
+      )
```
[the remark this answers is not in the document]
phirework: Totally agreed about not running a big backfill!

Re: keeping everything in one table - my understanding of SQL sort is that it only acts on the preceding result set, so unless we have a user who's updating their demographics millions of time the sort should be on < 100 entries. (Also, MySQL 8.0 supports `DESC` indices! We're currently still on 5.7 but specifying a `DESC` index is just ignored in 5.6/5.7 and we plan on upgrading to 8.0 in Q3, so you could pre-emptively add it to `demographics.updated_at` if you'd like.) I know joins are more efficient than subqueries, but since joins are matrix multiplication it feels like it'll be slower than sorting because the volume of data it has to examine is multiplicative? I also have not tested this though and could be totally offbase 😅 

I'm not against having two tables, but if we do stick to them could we name them something more self-explanatory? like `demographics_current` and `demographics_historical`? I just know I'm going to forget which is which in six months.

server/src/server.ts
```diff
@@ -285,6 +286,7 @@ export default class Server {
 
     try {
       await this.model.performMaintenance();
+      await migrateDemographics();
```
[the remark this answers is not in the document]
phirework: Yeah, I think this can be just run the same way other db upgrades are done through `server/src/lib/model/db/migrations`.

server/src/lib/model/db/migrate-demographics.ts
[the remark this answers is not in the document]
phirework: Hmm, I'm not sure I understand what you mean about the select problems.

server/src/lib/model/db/migrate-demographics.ts
```diff
@@ -5,9 +5,35 @@ const db = getMySQLInstance();
...
+
+      /* Sex and gender exist across multiple spectra which are not adequately
+         represented in our current database schema.
+
+         Including demographic data with our dataset is necessary to ensure its
+         utility to researchers, and helps us track the fullness and inclusion
+         of our collection process. We will ALWAYS keep this field optional. We
+         only store records for users who opt in, and will always (at very
+         least) maintain an "Other" option.
+      */
```
phirework, line 28: Woohoo!

server/src/lib/model/db/migrate-demographics.ts
```diff
@@ -5,9 +5,35 @@ const db = getMySQLInstance();
...
+      INSERT INTO sexes (sex) VALUES ('female', 'male', 'other')
+
+      /* FIXME: Only insert records with non-null demographics. */
+      INSERT INTO demographics (client_id, age, sex)  (
+        SELECT user_clients.client_id, ages.id, sexes.id
         FROM user_clients
+        INNER JOIN ages ON ages.age = user_clients.deprecated_age
+        INNER JOIN sexes ON sexes.sex = user_clients.deprecated_gender
```
phirework, line 36: I'm not super familiar with multiple inner joins, will this only insert records that have specified both an age AND a gender?

web/src/components/pages/datasets/stats.ts
```diff
@@ -38,7 +38,7 @@ export default {
           eighties: 0,
           nineties: 0,
         },
-        gender: { '': 0.39, male: 0.46, female: 0.13, other: 0.02 },
+        sex: { '': 0.39, male: 0.46, female: 0.13, other: 0.02 },
```
phirework, line 41: This is something we'll have to discuss, because these labels are also used in our dataset, and changing a column name is a breaking change as far as dataset continuity is concerned and right now if we changed it outright it would break DeepSpeech. We can either ask them to conform to our new label (and I don't know if there are industry standards regarding this) or handle the label translation in our bundler code (no input from ML team needed). Any thoughts on this @mbransn?

server/src/lib/model/db/migrate-demographics.ts
mbransn: Yes! :D

web/src/components/pages/datasets/stats.ts
mbransn: If I'm reading this correctly, the proposal is to move to `sex` from `gender` as this column label? If yes, I'm surprised this wasn't already the case in the dataset since `sex` is what is utilized across the platform. This was an intentional choice based on industry standard for articulating sex in datasets (and I believe Kelly was part of that discussion with Henretty at the time?). At minimum let's handle on the bundler side and ideally work for uniformity with DS.

server/src/lib/model/db/migrations/20200603213417-add-demographics-table.ts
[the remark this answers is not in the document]
rileyjshaw: These are some awesome points and I am fully in favour of dropping any confusing tables that we don't need. Will update the PR :)

server/src/server.ts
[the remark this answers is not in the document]
rileyjshaw: Moved it all into the same migration 👍

server/src/lib/model/db/migrate-demographics.ts
rileyjshaw: Your instinct is correct! Fixed in latest commit, sorry for not catching this earlier.

server/src/lib/model/db/migrate-demographics.ts
rileyjshaw: 🎉

rileyjshaw:
Okay! After focusing on other things, I'm back at it with this diff  💪

As you know from our pairing session, the way that we search and return user accounts is strange. I rabbit-holed on that for a while after our meeting, and I think there's a lot of cleanup that can happen by handling accent lists in the SQL query itself, and out of bespoke reducers. BUT, the changes introduced in this commit are already huge, and I didn't want to make these changes any more risky than they need to be. Now that we've removed the `user_client_demographics` table, I've kept all the demographics bits into subqueries with `LIMIT 1`.

server/src/lib/model/db/migrate-demographics.ts
```diff
@@ -0,0 +1,26 @@
...
+        SELECT user_clients.client_id, demographics.id AS demographic_id
+        FROM user_clients
+        INNER JOIN demographics ON user_clients.client_id = demographics.client_id
+      )
+
+      INSERT INTO clip_demographics (clip_id, demographic_id)  (
+        SELECT clips.id AS clip_id, demographics.id AS demographic_id
+        FROM clips
+        INNER JOIN demographics ON clips.client_id = demographics.client_id
+      )
```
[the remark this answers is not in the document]
rileyjshaw: Done! I think the `INNER JOIN` will help a lot with performance now that the `demographics` table doesn't include null entries.

server/src/lib/model/db/migrate-demographics.ts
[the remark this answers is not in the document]
rileyjshaw: I just did some pretty extensive testing for the issue I mention above, so no worries! I updated the table, it's sparse now :)

server/src/lib/model/db/migrations/20200603213417-add-demographics-table.ts
```diff
@@ -0,0 +1,79 @@
...
+        FOREIGN KEY (sex_id) REFERENCES sexes (id),
+        UNIQUE (client_id, age_id, sex_id)
+      );
+
+      CREATE TABLE clip_demographics (
+        clip_id BIGINT(20) UNSIGNED NOT NULL PRIMARY KEY,
+        demographic_id BIGINT(20) UNSIGNED NOT NULL,
+        FOREIGN KEY (clip_id) REFERENCES clips (id),
+        FOREIGN KEY (demographic_id) REFERENCES demographics (id)
+      );
```
rileyjshaw, line 31: Will work on populating this tomorrow now that the proof of concept has passed initial review!

rileyjshaw: Done!

server/src/lib/model/user-client.ts
```diff
@@ -75,10 +108,24 @@ const UserClient = {
...
         LEFT JOIN locales on accents.locale_id = locales.id
+        -- TODO: This subquery is VERY awkward, but safer until we simplify
+                 accent grouping.
+        CROSS JOIN
+          (SELECT ages.id AS age_id, sexes.id AS sex_id
+            FROM user_clients
+            LEFT JOIN demographics ON user_clients.client_id = demographics.client_id
+            WHERE user_clients.${
+              client_id ? `client_id = ${client_id}` : `email = ${email}`
+            }
```
phirework, line 123: This honestly feels like magic haha 🎉

rileyjshaw: Hahah I know 😞 dark, dangerous magic. Leaving a note for my future self below, feel free to disregard.

---

Hi Riley! I hope 2021 is off to a good start. Yes, this is what was causing that horrible bug – sorry.

server/src/lib/model/user-client.ts
```diff
@@ -65,6 +65,39 @@ async function updateLocales(
...
+  );
+
+  const [[{ id: sexId }]] = await db.query(
+    `
+    SELECT id
+      FROM sexes
+      WHERE sex = ?
+  `,
+    [sex]
+  );
```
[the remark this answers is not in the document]
rileyjshaw: This is awesome feedback, and may have caught a bug early. I'm going to look at this tomorrow with fresh eyes; since `age` and `sex` aren't necessarily sent up with every profile update, I want to make sure we're not eg. wiping someone's demographic data every time they toggle their `visible` setting.

rileyjshaw:
Merged and rebased, but I just realized this is still trying to merge into `master`! I'll open up a new PR for `main`.

PR #2785, opened 2020-06-24
Ensure client_ids are RFC4122 compliant GUIDs on the server

rileyjshaw:
Jira reference: https://jira.mozilla.com/browse/OI-1294

We generate client_ids and auth tokens on the client, which has the advantage of
easy analytics continuity for phantom / pre-signup users. However, our server
doesn’t verify that clients send standard patterns for these fields, which has
led to some… interesting entries in our DB.

Initially, I flipped the logic to generate everything on the server. At lunch I
realized I’d completely overkilled the problem; I think a simple server-side
check does what we’re trying to do well, and avoids needing to manage the
pipeline and state synchronization between client / server (which was getting
very messy with analytics, localStorage, etc).

_**IMPORTANT:**_ This commit alters our user verification code, and will restrict
access for the 421 users we currently have with non-standard entries:

```
mysql> SELECT COUNT(*) AS nonstandard
    ->   FROM user_clients
    ->   WHERE client_id NOT REGEXP '^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$'
    ->   OR (
    ->     auth_token IS NOT NULL AND
    ->     auth_token NOT REGEXP '^[0-9a-f]{40}$'
    ->   );
+-------------+
| nonstandard |
+-------------+
|         421 |
+-------------+
```

Rather than continue to special-case these users, I move that we amend the
nonstandard client_ids, and wipe any invalid tokens. Happy to write a quick
script for that, but I’ll wait for your initial feedback first.

---

This commit also removes a utility file that wasn’t being used.


phirework:
Before we merge this in and migrate the non-standard accounts, can you check: 

* How many of those nonstandard accounts have emails attached (I don't think we allow login without email anymore)
* How many of them have any associated entries in user_client_activities (which will tell us when they were last on site)
* How many of them have any associated clips or votes (since deleting user accounts will cascade everywhere else)

I would suspect that most of the ones that don't conform to our GUID are from people who were stress-testing our system rather than actually submitting, and we'll know how safe it is to wipe them.

server/src/lib/model/db.ts
```diff
@@ -413,6 +413,13 @@ export default class DB {
     id: string,
     auth_token?: string
   ): Promise<boolean> {
+    const guidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
+    const authRegex = /^[0-9a-f]{40}$/;
+
+    if (!guidRegex.test(id) || (auth_token && !authRegex.test(auth_token))) {
```
phirework, line 419: Is it every likely for the id/token to not be all lowercase? It might be safer to just do a quick transform to lowercase, just in case.

rileyjshaw: >     * How many of those nonstandard accounts have emails attached (I don't think we allow login without email anymore)

…exactly ZERO accounts.

>     * How many of them have any associated entries in user_client_activities (which will tell us when they were last on site)

139 accounts, responsible for 5781 activities. Some as recent as a few days ago.

>     * How many of them have any associated clips or votes (since deleting user accounts will cascade everywhere else)

13 have associated clips (12 of which are from telegram), 1 has associated votes, none of this activity occurred within the past year.

rileyjshaw:
@phirework thanks for the review!

>     * How many of those nonstandard accounts have emails attached (I don't think we allow login without email anymore)

…exactly ZERO accounts.

>     * How many of them have any associated entries in user_client_activities (which will tell us when they were last on site)

139 accounts, responsible for 5781 activities. Some as recent as a few days ago.

>     * How many of them have any associated clips or votes (since deleting user accounts will cascade everywhere else)

13 have associated clips (12 of which are from telegram), 1 has associated votes, none of this activity occurred within the past year.

server/src/lib/model/db.ts
rileyjshaw: Discussed offline and updated `authRegex` 👍

phirework:
I'd say let's get this merged in and wipe the non-standard accounts that don't have activity within the last year OR any clips and see if anyone complains about no longer being able to login :P

rileyjshaw:
> I'd say let's get this merged in and wipe the non-standard accounts that don't have activity within the last year OR any clips and see if anyone complains about no longer being able to login :P

Awesome! Saving this for later reference:

```
SELECT client_id, auth_token
	FROM user_clients
	WHERE (
		(
			auth_token IS NOT NULL
			AND auth_token NOT REGEXP '^[[:alnum:]]{40}$'
		)
		OR client_id NOT REGEXP '^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$'
	)
	AND client_id NOT IN (SELECT DISTINCT client_id FROM clips)
	;
```

PR #2788, opened 2020-06-26
Layout/create account

NanaKay007:
@phirework the only thing I did different was importing a chevron-left icon. I know that a chevron-right could be rotated to a left icon, but when working with RTL languages extra styling will be required to make sure both icon variants flip in the right directions (I know its difficult to visualize this).


phirework:
Nope that totally makes sense to me re: having a separate left and right arrow, thanks for being thoughtful about this! Everything looks good here.

PR #2794, opened 2020-06-30
Don't rely on client's locale when saving clips

rileyjshaw:
We've been seeing reports of clips of one language showing up on the
/listen page of another language. Our client does its best to be
responsive and quick when a user changes their language… as it should.
But when someone changes their language mid-way through recording a
batch of clips, it shouldn't affect the language those clips are saved
under.

This is a small and simple commit to move locale-handling to the server.
It adds a db call with `db.findSentence`, but removes three calls by
avoiding `db.getLocaleId`… so I think this might be a performance win in
the end.

### DIFF
--- a/server/src/lib/clip.ts
+++ b/server/src/lib/clip.ts
@@ -121,9 +121,9 @@ export default class Clip {
    * TODO: Check for empty or silent clips before uploading.
    */
   saveClip = async (request: Request, response: Response) => {
-    const { client_id, headers, params } = request;
-    const sentence = decodeURIComponent(headers.sentence as string);
-    const sentenceId = headers.sentence_id;
+    const { client_id, headers } = request;
+    const sentenceId = headers.sentence_id as string;
+    const sentence = await this.model.db.findSentence(sentenceId);
 
     if (!client_id || !sentence || !sentenceId) {
       console.log(`sent headers: ${JSON.stringify(headers)}`);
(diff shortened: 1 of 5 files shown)


server/src/lib/model/db.ts
```diff
@@ -721,6 +719,14 @@ export default class DB {
     )[0][0];
   }
 
+  async findSentence(id: string) {
+    return (
+      await this.mysql.query('SELECT * FROM sentences WHERE id = ? LIMIT 1', [
```
phirework, line 724: Nitpick - since we only need the `locale_id` and the `text`, we should specify those two columns instead of grabbing everything. I know most of the other columns are pretty small, but since the sentences table is in our top 3 biggest tables, the less we have to scan through the better.

rileyjshaw: Fixed in latest commit ✅

server/src/lib/clip.ts
```diff
@@ -121,9 +121,9 @@ export default class Clip {
...
    */
   saveClip = async (request: Request, response: Response) => {
-    const { client_id, headers, params } = request;
-    const sentence = decodeURIComponent(headers.sentence as string);
-    const sentenceId = headers.sentence_id;
+    const { client_id, headers } = request;
+    const sentenceId = headers.sentence_id as string;
+    const sentence = await this.model.db.findSentence(sentenceId);
 
     if (!client_id || !sentence || !sentenceId) {
```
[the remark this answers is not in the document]
rileyjshaw: Good point. Updated this (and the `saveClipVote` logic that it was based on) in my most recent commit, they now send a `400` for missing parameters and a `422` if the provided ID doesn't match a resource on our server.

PR #2801, opened 2020-07-03
Update/readme

NanaKay007:
updated readme to contain instructions on possible manual installation errors one may encounter.


CONTRIBUTING.md
```diff
@@ -125,6 +139,35 @@ web        | [BE] NEW_RELIC_APP_NAME. Not starting!
...
+[BE]     sql: '\n' +
+[BE]       '      ALTER TABLE user_clients\n' +
+[BE]       '        ADD COLUMN sso_id VARCHAR(255) UNIQUE,\n' +
+[BE]       "        ADD COLUMN username TEXT NOT NULL DEFAULT '',\n" +
+[BE]       '        ADD COLUMN basket_token TEXT;\n' +
+[BE]       '    '
+```
+
+First make sure you have the correct version of MySQL (v8.0 or greater) installed. If the problem still persists, the following will prove useful:
+**You only need to follow these steps once. After that be sure to discard all changes made to the relevant files after migrations are successful.**
```
phirework, line 164: Slight nitpick to avoid making people do unnecessary install steps - we're actually using MySQL 5.7 on the server, haha, and the `ER_BLOB_CANT_HAVE_DEFAULT` issue is actually something that even 5.x versions enforce: https://dev.mysql.com/doc/refman/5.7/en/data-type-defaults.html

The other way to allow migration to pass through is to turn off in your local, which I didn't bother trying to figure out because we were screensharing: 

>  If strict mode is not enabled, MySQL sets the column to the implicit default value for the column data type. 

Your stated workaround here is still fine, just remove the bit about server versioning.

NanaKay007: ha okay got it.

PR #2802, opened 2020-07-05
Use compat build of Fluent

kellda:
Use compat build of fluent to support Edge older than 79.

See projectfluent/fluent.js#431 and mozilla-iot/gateway#2238

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -3,8 +3,8 @@
   "version": "1.0.0",
   "private": true,
   "dependencies": {
-    "@fluent/bundle": "^0.15.0",
-    "@fluent/react": "^0.12.0",
+    "@fluent/bundle/compat": "^0.15.0",
+    "@fluent/react/compat": "^0.12.0",
     "@fullstory/browser": "^1.3.1",
     "@types/amplitude-js": "^5.8.0",
     "@types/classnames": "^2.2.9",
--- a/web/src/components/app.tsx
+++ b/web/src/components/app.tsx
@@ -44,7 +44,7 @@ import {
   LocalePropsFromState,
 } from './locale-helpers';
 import { Flags } from '../stores/flags';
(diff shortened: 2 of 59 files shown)


kellda:
Done. Doesn't work however.

The PR I linked doesn't seems to change `package.json`, so I would gess changing it doesn't really help. Maybe there just isn't a compat build for this version or for `@fluent/react`.

Note that I just redid what was done that PR, but without really understanding much of what I was doing.

phirework:
So looking at this more closely, we actually dropped the `compat` in a recent PR, because Common Voice doesn't make use of any of its components and we use Webpack/Babel to handle ES6 compatibility: https://github.com/mozilla/voice-web/pull/2631. Have you confirmed that Edge 78 and older show a blank screen on CV? We use localized strings everywhere and I haven't heard any bug reports about this issue.

kellda:
Oh didn't known that.

I can confirm that common voice shows a blank screen on Edge 18. However it likely works on Edge 78.

phirework:
Ahh gotcha, thanks for the clarification. Our official browser support is set up for the last 2 major versions of each, with legacy IE excluded, so that would be why. I'm going to close this PR for now, but if the lack of support for legacy IE is preventing you from being able to contribute Common Voice can you open up a new issue? Thanks again!

kellda:
Ok, no problem.

In fact this has prevented me to use Firefox Send, since the recipient wasn't able to open my link. Then I searched why it didn't worked and found that was because of Fluent. So I checked that other sites known to use Fluent worked with Edge 18.

So this isn't preventing me to contribute to common voice, as I use Firefox, but it may prevent others to do so (caniuse says Edge 12-18 has 1.65% global usage).

phirework:
That's good context to know, thanks.

Issue #2786, opened 2020-06-24
Automatically change language of tops


deevroman:
When switching the language in the header, the language of the tops does not change, although it would be more convenient and logical if it did.  

![CommonVoiceFeature](https://user-images.githubusercontent.com/25708359/85627472-384aa180-b677-11ea-8397-62eb8e02153f.gif)

phirework:
Thanks for reporting! This was included in the [sprint 31 release](https://github.com/mozilla/voice-web/releases/tag/release-v1.31.0).

PR #2807, opened 2020-07-06
Feature/auto switch local for stats card

NanaKay007:
@phirework. this is a PR for issue #2786; I basically passed in dashboardLocale as a prop for the stats-card component, and set-up a side-effect to change the card's locale whenever dashboardLocale changes. To avoid making unnecessary changes in other locations, I made the new stats-card prop (currentLocale) optional.


phirework:
Looks good to me!

Issue #2791, opened 2020-06-27
Cannot choose language in Dashboard for Italian on mobile


Sav22999:
**Describe the bug**
"Your language" and the select-box is not visible in mobile website for Italian language

**To Reproduce**
1. Go to voice.mozilla.org/it/dashboard
2. See error (You cannot chabge language)

**Expected behavior**
It should show the select-box

**Screenshots**
![Screenshot_20200627-140022.png](https://user-images.githubusercontent.com/32335378/85921778-eaa08580-b87e-11ea-92a5-316da82f1a3c.png)

**Desktop or Mobile (please complete the following information):**
 - OS: Android
 - Browser: Firefox and Firefox Preview
 - Version: 68.9.0

phirework:
Hello, thanks for reporting this! Where are you expecting to be able to select language, in that screenshot you posted? If you click on your profile pic, does a menu like this show up?

<img width="390" alt="Screen Shot 2020-06-30 at 18 29 38" src="https://user-images.githubusercontent.com/582683/86183286-b091e680-baff-11ea-8217-2099345edd47.png">

Sav22999:
@phirework there you change the language of the website, not the language of statistics. It should appear like this screenshot (it's from another device)

![Screenshot_20200701-014406.png](https://user-images.githubusercontent.com/32335378/86187582-6844e980-bb3c-11ea-8ce3-82e918a02e72.png)

phirework:
Ahh got it, thanks for the clarification!

phirework:
Thanks for reporting! This should be fixed with [release 31.0](https://github.com/mozilla/voice-web/releases/tag/release-v1.31.0).

PR #2816, opened 2020-07-13
Bug fix: styling for dashboard language switch

NanaKay007:
This is a fix for the styling issue reported in #2791 .

### DIFF
--- a/web/src/components/pages/dashboard/dashboard.css
+++ b/web/src/components/pages/dashboard/dashboard.css
@@ -118,6 +118,7 @@
         display: flex;
 
         @media (--xs-only) {
+            margin-top: 10px;
             flex-direction: column;
             justify-content: space-between;
         }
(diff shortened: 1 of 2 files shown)


web/src/components/pages/dashboard/dashboard.css
```diff
@@ -71,6 +76,7 @@
 
         a {
             margin-inline-end: 25px;
+            overflow-wrap: anywhere;
```
phirework, line 79: Hmm, this starts wrapping even when all three words are still visible on the page. I personally don't think it's necessary - it's not ideal for the last word to be cut off (if it is, depending on the language) but with individual letters wrapped it seems even harder to read.

NanaKay007: fixed :)

phirework:
Excellent, thanks for bearing with me!

PR #2819, opened 2020-07-13
Scrub leaderboard data if there are fewer than 5 users on list

phirework:
@mbransn According to the strictest definition of the rule of 5, we should be hiding leaderboard entries for all languages that have fewer than 5 unique speakers, which as of right now is just Votic. But since leaderboards are influenced by user settings, I've set it up so that leaderboard entries are hidden for all languages that have fewer than 5 visible speakers, which feels more in line with the spirit of the rule. Let me know if you have any concerns about this!


rileyjshaw:
Great idea, and very clear implementation. Sweet!

mbransn:
@phirework brilliant, makes total sense to me. Thank you!

PR #2821, opened 2020-07-13
Replace localized string with corrected closing html element

### LINKED ISSUE #2811: `en` source string contains HTML markup error creates errors for all locales
**Describe the bug**
Notice that <privacy notice> markup misses the *closing* element, causing problem for all other locales with the same problem: https://github.com/mozilla/voice-web/blob/master/web/locales/en/messages.ftl#L13 

**Expected behavior**
- Remove this string,
- Add a new string ID with the bug fix. 

IMPORTANT: do not reuse the same string ID with new content. Updating content with the same string ID will not generate new unlocalized string. 

@mbransn

phirework:
Fixes #2811 - @peiying2, just want to make sure that this is what you meant?


phirework:
Will do, thank you!

PR #2823, opened 2020-07-14
Feature/intro locale switch

NanaKay007:
This is the implementation for the locale switch on the intro page.

### DIFF
--- a/web/src/components/demo-pages/intro/intro.css
+++ b/web/src/components/demo-pages/intro/intro.css
@@ -10,6 +10,13 @@ html {
     width: 100vw;
     background: var(--white) url('./assets/waves.svg') no-repeat center;
     background-size: contain;
+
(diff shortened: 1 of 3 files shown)


web/src/components/demo-pages/intro/intro.css
```diff
@@ -54,7 +61,7 @@ html {
 }
 
 #intro-container--gradient-layer {
-    min-height: 100vh;
+    min-height: 105vh;
```
phirework, line 64: Was this just a visual change?

phirework:
Looks great! Good use of code from elsewhere :D

web/src/components/demo-pages/intro/intro.css
NanaKay007: yes it was

Issue #2718, opened 2020-05-14
Code snippet/icon missing for ‘add to homescreen’ on iOS


Githubuxer:
Code snippet/icon missing for ios ‘add to homescreen’ function.

It’s great that voice is now a website to reduce app development issues and user install issues, but the code snippet and icon is missing to make it look like an app on iOS, which is very easy to implement:

https://www.create.net/support/adding-an-ios-home-screen-icon-for-your-website

This is what it currently looks like:

![image](https://user-images.githubusercontent.com/62911518/81934179-87071500-95bc-11ea-81b7-59632ef0a997.jpeg)

phirework:
Hello! Thanks for reporting this. Are you actually able to record when you open the site from the homescreen app? In our testing we found that only opening it in a regular instance of Safari directly worked, it's possible we overlooked something.

Githubuxer:
Are there any free open source screen recording apps that are safe to use? It just opens right to the website and you can press the microphone button.

phirework:
Your iPhone should have screen recording natively, but you might have to [add it to the control center](https://support.apple.com/en-us/HT207935) if it's not already there, to easily get to it.

Once you've recorded those clips, are you then also able to play them back? If you tap on any of the numbers once you've recorded on it, a little control tray should pop up.

Githubuxer:
I was able to open it and record/play it back to ensure it heard my voice. 


Download “IMG_6909.MP4” with Firefox Send: simple, safe file sharing https://send.firefox.com/download/d64ad6282ffd58eb/#18skwopMV9WC2C8vfqr54A

phirework:
Awesome, thanks for the confirmation, we'll definitely look into it.

PR #2831, opened 2020-07-17
Feature/apple touch icon

NanaKay007:
@phirework  This is a fix for the issue mentioned in #2718 
I tested this on my iphone 8+ device (via remote debugging), and it works.


web/index_template.html
```diff
@@ -51,6 +51,7 @@
       href="https://fonts.googleapis.com/css?family=Zilla+Slab:400,600"
       rel="stylesheet"
     />
+    <link rel="apple-touch-icon" href="./apple-touch-icon-150x150.png" />
```
phirework, line 54: Can you use the same convention as the other static icon images in this file, i.e. using the static img file? Also, I don't think the filename itself has to be called apple-touch-icon or whatever, as long as it's specified in the `rel`. 

I've gotten a better icon for this from the design folks as well. 
CV iOS App Icon

NanaKay007: Hmmm...I dont know what you mean by 'static img file'. Is this what you mean?

`<link rel="apple-touch-icon" href="https://voice.mozilla.org/img/voice-touch-icon.png" />`

phirework: Yep exactly. We're using the `/img` folder for static meta image assets, so let's stick to the convention and avoid unnecessary files on the root level.

NanaKay007:
forgot to do that. thanks

Issue #2719, opened 2020-05-14
Blurry image on specific page of new website


OddHouseGames:
https://voice.mozilla.org/en/profile/settings

This image is blurry on the new website:

[image](https://user-images.githubusercontent.com/62911518/81936030-8cb22a00-95bf-11ea-9eeb-f32322a6a170.jpeg)

bacharakis:
hi @Githubuxer , can you please share with us the device you are experiencing this problem?
Thanks

Githubuxer:
iPhone 6S with latest iOS.

Rexogamer:
Also an issue on my iPad Air 3 running iOS 13.4.1:
![061535DD-8ACE-432F-BF68-5746D436FEF9](https://user-images.githubusercontent.com/42586271/82556510-c22bba00-9b61-11ea-9218-c64c24ae4bc1.jpeg)

PR #2836, opened 2020-07-20
added new images and styling

NanaKay007:
This a PR for the issue described in #2719 
I took screenshots of feedback pages on the CV site and updated the existing images. Also added styling for visual consistency.


phirework:
Hmm, do you know why the `-lg` file jumped up in size by 500%? Can I get you to run that image through some optimization protocols to try and get it to below 100kb at least? 

Alternatively - these are screenshots with no transparency so no reason they need to be png, if a high-quality JPEG is smaller let's switch to using that.

NanaKay007:
> Hmm, do you know why the -lg file jumped up in size by 500%? Can I get you to run that image through some optimization protocols to try and get it to below 100kb at least?
> 
> Alternatively - these are screenshots with no transparency so no reason they need to be png, if a high-quality JPEG is smaller let's switch to using that.

PR #2840, opened 2020-07-21
Add a redesigned 404 / 503 error page 

rileyjshaw:
This commit updates and unifies some of our error pages to make them prettier and more informative.

This commit also removes accidental suppression of 404 errors. In the past, visiting a non-existent URL _without_ a locale (eg. https://voice.mozilla.org/this-is-a-bad-url) would silently fail and redirect to a localized homepage. The old 404 page could only be seen when a locale was already included in the URL (eg. https://voice.mozilla.org/en/this-is-a-bad-url).

Now, _any_ bad URL will redirect to /<locale>/404, locale or not.

---

Note 1: I'm torn about this decision, but for simplicity I've updated the URL to `/:locale/:errorCode` whenever a 404 or 503 are encountered. I personally prefer for the old URL to hang around, because it makes wayfinding and refreshing easier. The following reasons made me opt for an explicit error code in the URL:

• We currently assign page-level classNames based on the URL. To get custom page CSS (eg. layout, hiding `#help-links`) we need a page-level classname. To do this without the help of the URL would require custom Redux work for this page alone, and it seemed like too big of a change for this initial commit.
• Our routing happens in several layers due to localization, and makes heavy use of the `<Redirect />` component. Following this pattern also requires we keep the error state in the URL.

I kinda hate it, but I don't hate it hate it. Let me know if you hate it hate it and I can make a ticket in Jira, but for now it seems pretty low priority.

---

Note 2: since the 503 page is shown whenever we hit the top-level `componentDidCatch`, it might appear for client-side errors. Since `componentDidCatch` is only triggered, I'd wager client-side errors will _usually_ be the cause.

---

Note 3: The new design removes the report + reload functionality from our previous barebones error page. I'm fine releasing as-is for now, but very happy to discuss if you've got ideas about whether to re-introduce reports, or flows where the new error page might cause problems.

---

Note 4: While working on this commit, Prettier and I made some small cleanup fixes (removed unused imports, used CSS variable names, etc). So as not to complicate the review, I'm going to add those in a follow-on commit called "Misc cleanup" – feel free to skip that one during review.

---

Test plan:

• Navigate to https://localhost:9000/en/this-is-a-bad-url

404 page should appear.

• Navigate to https://localhost:9000/this-is-a-bad-url

404 page should appear.

• Trigger the top-level `componentDidCatch` handler

503 page should appear.

In all above tests, layout should be responsive and animations performant.


rileyjshaw:
cc. @mbransn

rileyjshaw:
@phirework as requested during our chat earlier this week, I've added a custom GA event to track error page views. I cheated by squishing the entire initial path into the third argument, which we typically use for `locale` :)

Screen Shot 2020-07-24 at 12 46 57 PM

web/src/components/pages/error-page/error-page.css
```diff
@@ -0,0 +1,166 @@
...
+}
+
+@keyframes error-page-grow {
+    from {
+        transform: scale(1);
+    }
+    to {
+        transform: scale(1.3);
+    }
+}
```
phirework, line 166: These are delightful :D

web/src/components/pages/error-page/error-page.tsx
```diff
@@ -0,0 +1,78 @@
...
+            ].map(
+              ([LinkComponent, IconComponent]: [any, any], index: number) => (
+                <LinkComponent
+                  key={`get-involved-icon-${index}`}
+                  className="round-button">
+                  <IconComponent />
+                </LinkComponent>
+              )
+            )}
+          </div>
```
phirework, line 57: Totally out of scope for this particular issue, but we should abstract these icons into a "get involved" component, since they're re-used across the site in a few places.

phirework:
LGTM! Feel free to merge if you're ready.

Issue #2834, opened 2020-07-17
English shown last


davidak:
**Describe the bug**
English has the most recordings, but is shown last on the german site.

**To Reproduce**
1. Go to https://voice.mozilla.org/de/languages
2. Click on 'Mehr anzeigen'
3. Scroll down to 'Englisch'

**Expected behavior**
first in list

**Screenshots**
![Screenshot from 2020-07-17 21-17-08](https://user-images.githubusercontent.com/91113/87823758-48b7fb80-c874-11ea-85a1-f5bb76c87091.png)

on the english site it's correct
![Screenshot from 2020-07-17 21-23-23](https://user-images.githubusercontent.com/91113/87823780-579eae00-c874-11ea-85e2-3dbea2dab834.png)

**Desktop or Mobile (please complete the following information):**
 - OS: NixOS 20.03.1950.48723f48ab9 (Markhor)
 - Browser: Chromium 81.0.4044.138

phirework:
Thanks for reporting this! The intended behaviour is that the language that's currently selected should be first, and then everything else is sorted by # of hours validated, so that's been fixed now. https://commonvoice.mozilla.org/de/languages

PR #2844, opened 2020-07-24
Sort languages by active hours other than currently selected language

phirework:
Once upon a time in the halcyon days of 2018, we decided that languages on the `/languages` page should be sorted by the currently selected locales and then by active hours, except for English, which would always be last (unless you were currently localized as English). No one currently knows why this is, and it is a confusing user experience (see #2834) so we're changing this to prioritize the currently selected language and otherwise sort by active hours as normal.


rileyjshaw:
lmao… er, i mean, lgtm 😂

PR #2846, opened 2020-07-27
Prepare for voice -> commonvoice domain change

phirework:
There's a few places where it's hardcoded and a few documentation pages. I moved the URLs into the `urls.ts` file to make it easier in the future (though hopefully this will never need to happen again).

### DIFF
--- a/server/src/auth-router.ts
+++ b/server/src/auth-router.ts
@@ -76,8 +76,8 @@ if (DOMAIN) {
       clientSecret: CLIENT_SECRET,
       callbackURL:
         (({
-          stage: 'https://voice.allizom.org',
-          prod: 'https://voice.mozilla.org',
+          stage: 'https://commonvoice.allizom.org',
+          prod: 'https://commonvoice.mozilla.org',
           dev: 'https://dev.voice.mozit.cloud',
           sandbox: 'https://sandbox.voice.mozit.cloud',
         } as any)[ENVIRONMENT] || '') + CALLBACK_URL,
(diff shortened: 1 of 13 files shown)


The-smooth-operator:
LGTM

server/src/auth-router.ts
```diff
@@ -76,8 +76,8 @@ if (DOMAIN) {
       clientSecret: CLIENT_SECRET,
       callbackURL:
         (({
-          stage: 'https://voice.allizom.org',
-          prod: 'https://voice.mozilla.org',
+          stage: 'https://commonvoice.allizom.org',
+          prod: 'https://commonvoice.mozilla.org',
```
The-smooth-operator, line 80: we haven't modified the callback in Auth0 side, is the app enforcing somehow this callback?

web/src/urls.ts
```diff
@@ -34,4 +34,8 @@ export default Object.freeze({
     'https://discourse.mozilla.org/t/help-create-common-voices-first-target-segment/59587',
   TARGET_SEGMENT_INFO_ES:
     'https://discourse.mozilla.org/t/ayuda-a-crear-el-primer-objetivo-segmentado-de-common-voice/60472/',
+
+  HTTP_ROOT: 'https://commonvoice.mozilla.org',
```
The-smooth-operator, line 38: this should be an array to allow us using voice.m.o and commonvoice.m.o without a banner.

The problem using only one is that the deployment of the new image and the redirection have to be done exactly at the same time, otherwise users will see the banner.
Flux will deploy both changes at once, however there's a delay while the new container is pulled, so for some seconds the redirection will be applied, and the new code won't be running.

phirework:
> LGTM.
> 
> However I'd ideally display the banner based on `$CV_ENVIRONMENT=prod` defined [here](https://github.com/mozilla/voice-web/blob/master/server/src/config-helper.ts#L28) to be more flexible, the URL should not determine the environment.

Great callout, I'll make a note to update that in a future ticket. For future reference, we discussed this live and decided to go against the array approach because the `URLS` object is used in a lot of places and I didn't want to change the internal types on the fly. I'm going to close this for now since the URL changes were patched to production directly for this morning's release.

Issue #2832, opened 2020-07-17
Make verbs variables to address more forms of singular/plural forms in non-English languages


peiying2:
**Describe the bug**
Singular and plural forms in many languages come in more than 2 forms like English. The current structure limits localizers' ability to make grammatically correct in target language.

activity-needed-calculation-plural = { NUMBER($totalHours) -> 
  [one] {$totalHours} hour
  *[other] {$totalHours} hours
 } `is` achievable in just over { NUMBER($periodMonths) -> 
  [one] {$periodMonths} month
  *[other] {$periodMonths} months
 } if { NUMBER($people) -> 
  [one] {$people} person
  *[other] {$people} people
  } `record` { NUMBER($clipsPerDay) -> 
  [one] {$clipsPerDay} clip 
  *[other] {$clipsPerDay} clips 
  } a day.

**To Reproduce**
https://github.com/mozilla/voice-web/blob/695884b07107372a9718040bd68fab91ecd3b70d/web/locales/en/messages.ftl#L780

**Expected behavior**
Make the verb in the string a variable.

PR #2847, opened 2020-07-27
Fixes localization issue #2832

phirework:
Since the plural form of the verbs depends on the number in the preceding value, I rolled them into the existing variables. Let me know if this is ok.

### DIFF
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -641,7 +641,7 @@ contribute = Contribute
 listen = Listen
 skip = Skip
 shortcuts = Shortcuts
-clips-with-count-pluralized = { $count -> 
+clips-with-count-pluralized = { $count ->
                     [one] <bold>{ $count }</bold> Clip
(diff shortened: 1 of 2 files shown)


web/locales/en/messages.ftl
```diff
@@ -774,24 +774,24 @@ help-reach-hours-pluralized = Help reach { NUMBER($hours) ->
...
   [one] {$hours} hour
   *[other] {$hours} hours
  } in a language with a personal goal
 set-a-goal = Set a goal
 cant-decide = Can't decide?
-activity-needed-calculation-plural = { NUMBER($totalHours) -> 
-  [one] {$totalHours} hour
-  *[other] {$totalHours} hours
- } is achievable in just over { NUMBER($periodMonths) -> 
+activity-needed-calculation-plural-v2 = { NUMBER($totalHours) ->
```
peiying2, line 783: Since the old one string contains grammatical error, you can keep the string ID and replace with the correction in grammar.

PR #2848, opened 2020-07-27
message strings from an earlier commit

NanaKay007:
@phirework The previous merge from master erased all message strings for the demo mode. This PR fixes that.


phirework:
Shoot, sorry about that and thanks for correcting quickly!

PR #2854, opened 2020-07-29
Change repo name and default branch name

phirework:
This is mostly documentation updates. Travis is now looking at a new branch, but it's still pushing to `itsre/voice-web` because that's a setting in Docker Hub that I don't think we're planning on changing right now, right? Unless you want to move everything into the new Docker Hub org now as well...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ deploy:
   provider: script
   on:
     all_branches: true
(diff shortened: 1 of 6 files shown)


The-smooth-operator:
@phirework I've plans to move to a new Dockerhub repository, which is already created under Mozilla's Dockerhub org.
However I've been very busy and couldn't find time yet to do it. Will let you know

The-smooth-operator:
btw this is great :tada:

The-smooth-operator:
@phirework I've just noticed you merged this one to master instead of main, so changes are not present in main

PR #2859, opened 2020-07-31
Add logging for voice avatar and base64 upload handlers

rileyjshaw:
We have a lot of dead code in our codebase built around the
eventual desire to ship the voice avatar.

I am investigating bugs around clip recording and uploading, so I want
to ensure these paths are never hit in our codebase. By adding some
simple logs, I can check back in Papertrail after a week or so to ensure
these conditions weren't triggered. If they weren't, good. If they were,
hello bug!


phirework:
LGTM!

phirework:
We also do have Papertrail logging for the other environments, so if you want to do some clicking around and testing in Dev you can also see if those paths are hit.

PR #2863, opened 2020-08-06
Clear some cobwebs from our page layout

rileyjshaw:
Prompted by a surprising layout breakage on our new error pages, I found some oddities in our page layout. This is my attempt to simplify a few years of uh, iterative development.

I did a bunch of detective work and testing while going through this, which I've tried to document below. But please let me know if you have any questions or concerns!

I've got a big change to follow this one to get rid of the `#scroller` and `#scrollee` containers, but that deserves its own commit and explanation. This is some comparatively simple prep work.

Here are the changes, explained:

---

- Remove an unused `.focus` style.

Added in 2017 (pre 60ae00771519), this class was used to add custom Firefox Focus scroll behaviour. The class was removed in 16d38229594d, but I accidentally left a custom rule in place (unused since 16d38229594d).

- Remove some unused / duplicate `hasScrolled` state from `app.tsx`.

Honestly I didn't dive deep into this one because it's obviously not being used. The same `hasScrolled` behaviour is duplicated in `layout.tsx` so I assume it was just moved-but-not-deleted somewhere along the way.

- Remove an unused `.hidden` class (and associated `hasScrolledDown` logic).

It looks like there were plans to make the header hide itself on scroll, but that never happened. This was added in 225b84b75136 and never used. Better solutions exist in 2020, so rather than keep this around we should get rid of it.

- Remove an unused `.hidden` style.

Confusing! There _was_ a `.hidden` class (see above), but not within `.faq-hero`. This was added in 06b8c3ceb471 and apparently never used.

---

And that's it. Just in time for some real QA before release! Which is great, because this is gonna need it :)

### DIFF
--- a/web/src/components/app.tsx
+++ b/web/src/components/app.tsx
@@ -78,7 +78,6 @@ interface LocalizedPagesProps
 }
 
 interface LocalizedPagesState {
-  hasScrolled: boolean;
   l10n: ReactLocalization | null;
   uploadPercentage?: number;
 }
@@ -89,7 +88,6 @@ let LocalizedPage: any = class extends React.Component<
 > {
   seenAwardIds: number[] = [];
   state: LocalizedPagesState = {
-    hasScrolled: false,
     l10n: null,
     uploadPercentage: null,
   };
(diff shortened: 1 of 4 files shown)


phirework:
This all looks great to me! Thanks for the detailed explanation and documentation of initial changes, that made this way easier to look at :D

PR #2864, opened 2020-08-06
Remove the "scroller" and "scrollee" scroll containers

rileyjshaw:
# Depends on https://github.com/mozilla/common-voice/pull/2863

Remove the "scroller" and "scrollee" scroll containers

A more ambitious follow-up to my [my previous layout PR](https://github.com/mozilla/common-voice/pull/2863).

> Prompted by a surprising layout breakage on our new error pages, I found some oddities in our page layout. This is my attempt to simplify a few years of uh, iterative development.

Prior to this change, each of our pages lived within a nested `div#scroller` and `div#scrollee` container. On desktop these were inconspicuous wrapper `div`s. But on mobile, `#scroller` took on a `position: fixed` over the viewport, and `#scrollee` scrolled within it. With banners, dynamically-sized elements, and new flexbox layouts, keeping everything in a fixed panel has started to cause surprising bugs:

An error page, ironically displaying a horrible layout bug

The solution is one part new, one part old. I've combined the simple scroll check introduced in 1c58f6df2a5c with `position: sticky` for a _waaaay_ less verbose and mobile-platform-breaking experience!

It also fixes a bunch of weird overflow bugs that were too annoying to fix individually but really piled up. For instance: banners no-longer block their underlying content!

Before:

Banner is overlapping the underlying content

After: 

Banner is NOT overlapping the underlying content

---

*This change should go through some heavy QA!* Things to look out for:

- Pages that now have content overflowing horizontally, including the header bar, backgrounds, and "floating" images.
- Pages with wonky scroll behaviour, especially at the top or bottom of pages.
- Banners or other fixed-position elements that now seem to be in the wrong position.
- Pages that cannot scroll at all…? Hope not!

### DIFF
--- a/web/src/components/layout/footer.tsx
+++ b/web/src/components/layout/footer.tsx
@@ -16,10 +16,7 @@ const LocalizedLocaleLink = ({ id, to }: { id: string; to: string }) => {
   const [locale] = useLocale();
   return (
     <Localized id={id}>
-      <LocaleLink
-        to={to}
-        onClick={() => trackNav(id, locale)}
-      />
+      <LocaleLink to={to} onClick={() => trackNav(id, locale)} />
     </Localized>
   );
 };
@@ -29,10 +26,7 @@ export default React.memo(() => {
   return (
     <footer>
       <div id="help-links">
-        <LocaleLink
-          to={URLS.FAQ}
-          onClick={() => trackNav('faq', locale)}
-        >
+        <LocaleLink to={URLS.FAQ} onClick={() => trackNav('faq', locale)}>
           <SupportIcon />
           <Localized id="faq">
             <div />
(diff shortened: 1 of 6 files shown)


web/src/components/layout/footer.tsx
```diff
@@ -58,11 +52,13 @@ export default React.memo(() => {
...
-                  rel="noopener noreferrer"
-                  href="https://www.mozilla.org/en-US/foundation/licensing/website-content/"
-                />
+                licenseLink: (
+                  <a
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    href="https://www.mozilla.org/en-US/foundation/licensing/website-content/"
+                  />
+                ),
```
phirework, line 61: Is this just a linter change or was there a functional reason?

phirework:
Oh, there's a couple of conflicts (I think from merging #2863) so feel free to land once those are resolved.

web/src/components/layout/footer.tsx
rileyjshaw: Bah, yep! Just Prettier doing its thing, thanks for checking.

rileyjshaw:
Fixed the merge conflicts. Thanks for the review!

Issue #1210, opened 2018-07-03
Download personal recorded dataset


irvin:
One of the local researchers is trying to collect the voice samples of ALS patients, in order to synthesize the patient's voice when necessary. I feel we can really help with a "download personal dataset" feature.

mikehenrty:
This usecase is also good for building TTS (speech synthesis) based on your own voice.

irvin:
And I feel it’s also something we should implement in order to respect the principle of GDPR data portablility right.

kdavis-mozilla:
This use case can also be of interest in creating a STT engine for, say, people with a particular speech impediment.

mikehenrty:
As Irvin mentions, we will have to do this due to GDPR. Plus it's a nice feature to have.

Gregoor:
#434 is a related issue (maybe we should merge them)

Gregoor:
![Settings Build Profile_06 1](https://user-images.githubusercontent.com/4051932/61617979-2b098380-ac6c-11e9-9020-6b0d37dbf4b3.png)

We have designs for it, but so far haven't found time to look at implementation. Happy to review PRs for it!

kjreills:
I'd like to start working on this feature. This will be my first contribution, so let me know if you have any advice or additional information I should know.

Thanks!

phirework:
@kjreills Please feel free! If you run into any issues getting the repo set up locally for development please submit a PR to CONTRIBUTING.md so that we can improve our documentation :)

irvin:
the feature had landed to beta, 
ref: https://github.com/common-voice/common-voice/releases/tag/release-v1.55.0
#3242 #2865

PR #2865, opened 2020-08-06
Download personal recorded dataset (#1210).

tchief:
This PR addresses the issue described in #1210.

### Notes
-  `Recordings.zip` contains all mp3 files in `client_id/` S3 'folder' (see `Bucket.getRecordedClips`).
-  `Profile.txt` and `recordings.zip` are downloaded as 2 separate files.

### Questions
1. Is `### k/mb` a placeholder for actual size of items?
2. What should we do with huge downloads?
3. Is filtering by `mp3` extension logic is good enough?
4. Is downloading 2 separate files is good enough?
5. Is endpoint `/user_client/recordings` properly named?
6. Have I properly figured out where to put methods in `bucket.ts`, `api.ts`, `download.tsx`?
7. Should download functions be moved out of `download.tsx`?

### Need help with
- Styling.
- Translations.

### Tested
 - Locally using docker with Auth0 and S3.
 - Needs testing for bigger accounts.

### UX notes
Should Download button be green? There are different button colors in Profile now:
 - `Download` - green in mock design
 - `Save` - grey
 - `Delete` - blue
 - `Record voicewave` - white

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -50,6 +50,7 @@
     "redlock": "^4.0.0",
     "request": "^2.85.0",
     "request-promise-native": "^1.0.7",
+    "s3-zip": "^3.1.3",
     "stream-transcoder": "0.0.5",
     "ts-node": "^8.6.2",
     "universal-analytics": "^0.4.20"
--- a/server/src/lib/api.ts
+++ b/server/src/lib/api.ts
@@ -83,6 +83,7 @@ export default class API {
     );
     router.post('/user_client/avatar_clip', this.saveAvatarClip);
     router.get('/user_client/avatar_clip', this.getAvatarClip);
+    router.get('/user_client/recordings', this.getRecordings);
     router.get('/user_client/delete_avatar_clip', this.deleteAvatarClip);
     router.post('/user_client/:locale/goals', this.createCustomGoal);
     router.get('/user_client/goals', this.getGoals);
@@ -353,6 +354,12 @@ export default class API {
       response.statusMessage = 'save avatar clip error';
       response.json(error);
     }
+  };
+
+  getRecordings = async (request: Request, response: Response) => {
+    response.attachment('recordings.zip');
+    let recordings = await this.bucket.getRecordedClips(request.client_id);
+    recordings.pipe(response).on('end', () => { response.end(); });
   };
 
   getAvatarClip = async (request: Request, response: Response) => {
--- a/server/src/lib/bucket.ts
+++ b/server/src/lib/bucket.ts
@@ -1,3 +1,4 @@
+const s3Zip = require('s3-zip')
 import { S3 } from 'aws-sdk';
 import { getConfig } from '../config-helper';
 import Model from './model';
@@ -61,6 +62,20 @@ export default class Bucket {
       console.log('aws error', e, e.stack);
       return [];
     }
+  }
+
+  async getRecordedClips(client_id: string) {
+    const bucket = getConfig().BUCKET_NAME;
+    const folder = `${client_id}/`;
+    const params = { Bucket: bucket, Prefix: folder };
+
+    const allFiles = await this.s3.listObjectsV2(params).promise();
+    const recordings = allFiles.Contents
+      .filter(item => item.Key.endsWith('mp3'))
+      .map(item => item.Key.substr(folder.length));
+
+    return s3Zip
+      .archive({ s3: this.s3, bucket: bucket }, folder, recordings);
   }
 
   getAvatarClipsUrl(path: string) {
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -834,6 +834,17 @@ why-delete-recordings = Common Voice recordings are used by academics, small bus
     Can you let us know why you would like your recordings deleted?
 profile-form-delete = Delete Profile
 
+## Profile Download
+download-q = Need to download your data?
+download-info = Tell us what you'd like to download:
+download-profile-title = Profile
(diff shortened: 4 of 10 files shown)


tchief:
1210-download-personal-records

tchief:
This would be my first contribution.
@Gregoor, please, take a look when you have some time.

I hope this would make it easier to synthesize voice of people you care.

> One of the local researchers is trying to collect the voice samples of ALS patients, in order to synthesize the patient's voice when necessary. I feel we can really help with a "download personal dataset" feature.


> This usecase is also good for building TTS (speech synthesis) based on your own voice.

phirework:
Hi @tchief! Thank you for submitting this, this looks great. Gregor is no longer working on this project, so I will be reviewing this PR. I'm tagging @mbransn for a design review as well.

tchief:
Hi, @phirework! is there any progress merging that? Thanks.

tchief:
Happy 2021!

@phirework, is there any progress merging that? This is my longest outstanding PR ever, 6 months :)

Thanks.

phirework:
Hi @tchief, thanks for your patience and I apologize for the long delay. As you might be aware, this PR came in a week before Mozilla laid off 25% of the company, which also significantly impacted the Common Voice team. We've been in maintenance mode since, so my ability to support third party development has been pretty minimal. 

One of the reasons I haven't been able to just merge this in is precisely because of the large files issue. We have users who have tens of thousands of recordings, and the zip for that would both take a significant amount of time, and be subject to potential interruptions, such that you would want them to be able to restart or resume the download again without needing to rezip the entire file. Moreover, the platform currently doesn't have the ability to send out transactional emails, due to a quirk around Mozilla's internal CRM integrations. 

How I would approach solving the above issue would be:

* A database table to track recently generated user data zips (one per user)
* A server-side process that creates the zip, and puts it on the AWS server (at `bucket_name/client_id_timestamp.zip`, for example)
* The front-end would need to:
  * Check whether an archive is in the process of being generated, in which case the user will not be able to generate a new archive
  * Check if an archive exists, in which case create a download link to that archive 
  * If no archive exists, generate a new archive, with instructions for the user that they would need to check back in to see when the download is ready
* Modifying the server startup logic to purge archives older than 30 days 

The small bit of good news is that I would say you're safe to just zip the entire `client_id` folder entirely, especially if this process is running in the background. For new users, this will include Profile.txt and all mp3s. Some of our older users will also get records of votes and the sentence text for the clips, but as these are plaintext files, the incremental size is pretty negligible. 

Otherwise, this feature looks great, and your naming and organization seem fine. Translations for localization will happen via Pontoon (our crowdsourced platform) if/when this is merged, so you don't have to worry about that.

I'm happy to discuss this further, let me know how you'd like to proceed.

AigizK:
can the priority of this pull request be increased? @phirework

Zopieux:
>A server-side process that creates the zip, and puts it on the AWS server (at bucket_name/client_id_timestamp.zip, for example)

Alternatively, creating the zips can be done inside an AWS Lambda. The benefit would be that large files and processing would stay within the internal AWS network, possibly cutting egress/ingress costs. Whether it's actually cheaper than downloading from S3 + zipping on the server (is it EC2 or some Mozilla owned thing?) + sending back to S3 remains to be determined, and I cannot assess this myself, having no knowledge of the prod setup.

By the way, here is a generic user journey for scaling to any sized datasets, which I think isn't too inconvenient and quite simple to implement:

1. user requests their data on the FE
2. this creates a _pending_ takeout request on the BE
3. an offline process (AWS Lambda, background process on the server, whatever) starts handling the pending takeouts, changes status to _inprogress_
4. get total client folder size from S3, seems like it's [possible](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html#AmazonS3-Type-Object-Size)
5. if it's larger than an arbitrary size ceiling, say 100M, then split files in ~100M chunks, creating `N` ≥ 1 chunks
6. for each chunk, zip chunk files into `data-<request-id>-<x>-of-<N>.zip`, store that in S3 with a reasonable [expiration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html), eg. 30 days
7. once all chunks are done:
   * set the takeout to _available_
   * set the chunk count `N`
   * set the expiration date (now + a bit less than S3 expiration)
   * on the takeout page, display a list of the zip links for the user to download, in a clickable + line-delimited format suitable for passing it to a download manager for tech-savvy users
   * send an email to the user that the takeout is ready and expires on <date>
8. an offline process deletes (or sets to _expired_) the expired takeout requests

Therefore the table would look like this:

```sql
CREATE TABLE takeout_request (
  id BIGINT(20) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY,
  client_id CHAR(36),
  state CHAR(36) NOT NULL,  -- 'pending', 'inprogress', 'available'
  requested_date DATE NOT NULL,
  expires_date DATE NULL,   -- set when the request becomes 'available'
  chunk_count INT UNSIGNED NULL,
)
```

Open questions:

1. can the user request a new takeout while one is already running? I'd say no, it's asking for a cheap DoS.
2. can the user request a new takeout right after the last one becomes ready/available? I'd say yes, for ease of implementation, because an arbitrary delay would make little sense, and detecting "changes since last takeout" sounds error-prone and hard to do correctly.

phirework:
Hey folks, we're in the process of releasing @Zopieux's work in #3133 (continued in #3242, internal design review happening right now) so I'm closing this PR. Thank you so much for all your initial work @tchief, it was a great jumping off point.

PR #2867, opened 2020-08-07
Improve mobile error page styling

rileyjshaw:
Screenshot

### DIFF
--- a/web/src/components/pages/error-page/error-page.css
+++ b/web/src/components/pages/error-page/error-page.css
@@ -27,6 +27,7 @@
     }
 
     .error-page-container-outer {
+        border-top: 1px solid var(--light-grey);
         background: url('../images/wave-footer@3x.png') no-repeat bottom/100%;
 
         @media (--xs-down) {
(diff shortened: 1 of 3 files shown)


web/src/components/pages/error-page/error-page.css
```diff
@@ -42,13 +43,18 @@
...
-        @media (--sm-down) {
-            padding-bottom: 20px;
+        @media (--xs-down) {
+            padding-top: 32px;
+            padding-bottom: 40px;
         }
     }
 
-    .error-page-container-text {
+    .error-page-container-inner > * {
```
phirework, line 52: Hmm, looking at the template code for that, can this just be a `> div`? While the performance hit of `*` selectors is probably negligible on most modern systems, it's always good to reduce their use where possible.

rileyjshaw: No problem, I just replaced this with the specific child classes. For what it's worth, `> *` was an intentional decision – I think it reads clearer, and I haven't seen compelling evidence that selectors are a bottleneck. But either is fine by me!

phirework: Totally fair! Next 1:1 let's start hammering out some CSS guidelines maybe, since we want to start focusing on frontend / accessibility in the next little while.

PR #2889, opened 2020-08-27
Removed lines that just says a number.

Oymate:
Example:
Five
Six


MichaelKohler:
Why do you think those should be removed?

Oymate:
@MichaelKohler I felt like these do not add much value in terms of language detection as numbers were discouraged initially. But if it is acceptable than this PR may be closed.  :)

MichaelKohler:
Numbers are discouraged, but only if they are actual numbers like 1, 2, etc, absolutely fine if they are written out like "one", "two". Thanks though!

Oymate:
@MichaelKohler Thanks for reviewing  :)

PR #2891, opened 2020-08-31
Remove New Relic package, imports

phirework:
cc @rileyjshaw


The-smooth-operator (approved):
It looks good to me. There are few dependencies which have changed, but I'm confident are dependencies from NR package.

PR #2901, opened 2020-09-11
Use CDN for files smaller than 20gb

phirework:
Thanks to Alberto's work in https://github.com/mozilla-it/voice-infra/pull/111 we now have a CDN for Common Voice! Unfortunately Amazon Cloudfront's max is 20gb, so the English dataset will still need to be served from our regular bucket.


phirework:
cc @The-smooth-operator

PR #2903, opened 2020-09-21
Added luganda validated sentences

jmukiibi:
Have added a Luganda text-file with 5115 sentences for common voice

### DIFF
--- a/server/data/ug/sentence-collector.txt
+++ b/server/data/ug/sentence-collector.txt
@@ -1,0 +1,5115 @@
+Tugende e Makindye tulabe abayimbi n'abazannyi ba katemba abalungi.
+Emmere etali nsiike ebeera nnungi eri emibiri gyaffe kuba tebaamu butwa.
+Abawala abalungi baagalwa buli muntu naye abasinga tebabeera na mpisa.
+Wali obaddeko awo n'owulira ng'omutwe gulinga ogukwetooloola?
+Abayizi abatuula mu maaso batera okuba nga bategeera mangu ebibeera bisomesebwa.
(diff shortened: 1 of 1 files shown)


rileyjshaw:
This is a great contribution, thanks so much!

MichaelKohler:
@jmukiibi thanks for your work here, sorry I didn't comment on this before (I'll also follow up on the email from Remy)

@rileyjshaw This needs to be reverted and some changes need to be done. I wanted to write here earlier, but I had to finish something.

The sentence-collector files are reserved for the Sentence Collector (https://github.com/Common-Voice/sentence-collector). These file would be overwritten by a future export of the sentences from that tool. As these are already-validated sentences, it's fine to add them through a Pull Request, you can name the file whatever is most descriptive with a .txt ending - as long as it's not sentence-collector.json or sentence-collector.txt.

Additionally the correct folder to put those in would be "lg". Feel free to create the folder as well, as it currently doesn't exist. We need to use the same code as on Pontoon, as otherwise we're gonna end up with a wrong import: https://pontoon.mozilla.org/lg/common-voice/

Subsequent validated submissions can then be added through a second, third, etc file in the same "lg" folder.

I'll follow up for the rest in the mail in a second.

rileyjshaw:
Whoops, sorry everyone for the overeager merge; I've just reverted the commit. @jmukiibi, please create a new PR with the above changes are addressed and I'll merge it back in.

Thanks all!

jmukiibi:
@rileyjshaw and @MichaelKohler I have made another PR addressing the highlighted issues. Thanks a lot.

PR #2905, opened 2020-09-23
Add Luganda as a contributable language

rileyjshaw:


MichaelKohler (approved):
Was about to do that with my next export, but the sooner the better :)

PR #2910, opened 2020-10-01
Update CONTRIBUTING.md

uvinduperera:
centos capability


MichaelKohler (commented):
Yarn is not preinstalled on Cent OS. And "Better works on cent OS" is definitely not something we want in the README, and definitely not at that place.

PR #2915, opened 2020-10-12
Another way to fix  Issue #2253

ubaidisaev:
before
after

### DIFF
--- a/web/src/components/pages/profile/layout.css
+++ b/web/src/components/pages/profile/layout.css
@@ -42,6 +42,7 @@
         background: #f6f6f5;
 
         @media (--md-up) {
+            position: relative;
             margin-inline-end: 50px;
             padding-top: 50px;
             padding-inline-end: 30px;
@@ -74,10 +75,9 @@
             @media (--md-up) {
                 flex-direction: column;
                 margin: 0 auto;
-                width: 100%;
+                width: auto;
                 max-width: var(--wide-desktop-width);
                 position: absolute;
-                left: 0;
                 right: 0;
             }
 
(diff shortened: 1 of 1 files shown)


web/src/components/pages/profile/layout.css
```diff
@@ -74,10 +75,9 @@
             @media (--md-up) {
                 flex-direction: column;
                 margin: 0 auto;
-                width: 100%;
+                width: auto;
                 max-width: var(--wide-desktop-width);
                 position: absolute;
-                left: 0;
```
ubaidisaev, line 80: here we need to apply `left` or `right` based on language `dir`.

web/src/components/pages/profile/layout.css
```diff
@@ -74,10 +75,9 @@
...
                 right: 0;
             }
```
ubaidisaev, line 82: here we need to apply `left` or `right` based on language `dir`.

phirework:
Hi there, I'm not able to replicate that old UI bug anymore on the current version of CV, can you tell me what your steps to reproduce were and what your browser/OS/screen size specs are? https://commonvoice.mozilla.org/

ubaidisaev:
> Hi there, I'm not able to replicate that old UI bug anymore on the current version of CV, can you tell me what your steps to reproduce were and what your browser/OS/screen size specs are? https://commonvoice.mozilla.org/

Hi. In resolution `~ 2000px` and higher. 
`Winddow  10`
`Chrome-86`
`firefox-82`

phirework:
Gotcha. Instead of adding custom LTR/RTL rules I would actually prefer to refactor the column setup of the profile view to not rely on `--var(--wide-desktop-width)` for positioning. Is that something you'd feel comfortable taking on? You'd need to look at `layout.tsx` as well.

ubaidisaev:
Fixed without relying on LTR/RTL rules.

phirework:
Perfect, thank you!

PR #2920, opened 2020-10-24
remove dup or broken, typo sentences from zh-TW

irvin:
remove zh-TW duplicated or broken, typo sentences come from collector from Jan to Oct 2020


MichaelKohler (changes requested):
Thanks for reporting these. If we delete those only from the .txt file they won't be deleted from the Sentence Collector database, and get re-exported with the next export. I'll delete those from the db and then close this PR.

MichaelKohler:
https://github.com/Common-Voice/sentence-collector/commit/097724f0cfb66ecbe437d77dd31a2a18dfe087fc

irvin:
@MichaelKohler What is the proper way to deal with such cases? Should I just open similar MR to yours to sentences collector next time?

MichaelKohler:
Sure, if you want to.

PR #2946, opened 2020-12-02
NEW PARTNER RW BATCH 9 SENTENCES

rutsam:
**New partner RW sentences**

**partner-9.rw.txt**

- Number of sentences: 56,280  

- Error ratio from the QA: 97%

- Reviews: [1](https://github.com/rutsam/commons_dataset/blob/master/partner-9_verification/ALI.xlsx) [2](https://github.com/rutsam/commons_dataset/blob/master/partner-9_verification/PATRICK.xlsx) [3](https://github.com/rutsam/commons_dataset/blob/master/partner-9_verification/SAM.xlsx) [4](https://github.com/rutsam/commons_dataset/blob/master/partner-9_verification/VITAL.xlsx)


phirework:
Hello, just a heads up that the Common Voice default branch is `main`, not `master`. I've fixed it so that the partner sentences are in `main` as well and I'm going to go ahead and delete the `master` branch from remote so that this doesn't happen again. Thanks!

rutsam:
Great thanks for the heads up

PR #2950, opened 2020-12-09
Disable the restriction of single sentence limit for all locales

irvin:
Temp. revert feature from #2757 that restricts all sentences to be recorded only for once. 
Ease the contributing for all locales without new sentences add fast enough. 
resolve https://github.com/mozilla/common-voice/issues/2948


phirework:
Hey - I'm going to close it, we have a different fix for this in the works and hardcoding conditions like that isn't good practice. It'll go out later this week.

PR #2988, opened 2021-01-21
add more bulgarian sentences

pavlovt:
Hi,
I have prepared almost 40 000 sentences in bulgarian language according to the Commonvoice rules.
The authors I used are the following: Елин Пелин, Захари Стоянов, Иван Вазов, Йордан Йовков.
According to the bulgarian laws you can use any authors' works only if 70 yeas have passed after his/her death and this is the rule that I followed when choosing the authors.
All the authors works I got from here: https://chitanka.info/
I do not provide an excel file with checked sentences because I checked all the sentences that I provide.


phirework:
Hi, thanks for submitting this! Given that it's a source we haven't used before, I've passed it onto our legal team to review.

pavlovt:
Thanks! I am very glad to hear this :)
Please do not forget to  merge more-sentences.txt into /server/data/bg/sentence-collector.txt

phirework:
Hello! sentence-collector.txt is autogenerated from https://commonvoice.mozilla.org/sentence-collector/ and we do not manually change that file. Each text file in the language directory will get imported into the database next time there's a release (which will probably be end of this week), they don't have to be all part of the same file.

Issue #2314, opened 2019-11-13
[mobile] Logged in profile UI updates


georai:
https://www.dropbox.com/s/yvvq7u1ugx5ch70/Screenshot%202019-11-13%2018.44.19.png?dl=0

1. The spacing between the tabs needs to be like in design. Let me know if you need the measures. 
2. Weird box in the footer, please remove.
3. Can we have the dividers in the footer too: https://www.dropbox.com/s/l44lhpj9g2ajfkd/Screenshot%202019-11-13%2018.45.27.png?dl=0

@mbransn @rileyjshaw

rileyjshaw:
Related: https://github.com/mozilla/voice-web/issues/2313

rileyjshaw:
These are not Voice Challenge specific. Removing that label.

Dnouv:
I want to work on this issue. Please assign me on this if it's available.

Dnouv:
@georai, @phirework, I understand you may be busy, but I will truly appreciate if you could look into my request.

Dnouv:
@mbransn, could you please look into my request.

Dnouv:
@phirework, thank you so much for assigning me on this issue.

PR #3011, opened 2021-02-12
Logged In Profile UI updates

Dnouv:
As mentioned in Issue(#2314 ), there was a weird box. It is updated now, with new background color only for share icons. And, two new dividers are included as well in the footer. Regarding the first point, "the spacing between tabs..." appeared to me like already in the design. If there are any suggestions, please provide them. Here are the screenshots of the updated page.

On the page, before hovering on icons, there is also a divider at the bottom.
footer_noclick

On the page, after hovering on the icon(Twitter), the color is reversed. 
footer_after_hover

Dividers in the footer.
divider_footer

Feedbacks are most welcomed.

Thank You!

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -6,7 +6,6 @@
     "@fluent/bundle": "^0.15.0",
     "@fluent/react": "^0.12.0",
     "@fullstory/browser": "^1.3.1",
-    "@material-ui/core": "^4.11.3",
     "@types/amplitude-js": "^5.8.0",
     "@types/classnames": "^2.2.9",
     "@types/lodash.memoize": "^4.1.6",
--- a/web/package.json
+++ b/web/package.json
@@ -6,6 +6,7 @@
     "@fluent/bundle": "^0.15.0",
     "@fluent/react": "^0.12.0",
     "@fullstory/browser": "^1.3.1",
+    "@material-ui/core": "^4.11.3",
     "@types/amplitude-js": "^5.8.0",
     "@types/classnames": "^2.2.9",
     "@types/lodash.memoize": "^4.1.6",
--- a/web/src/components/layout/footer.css
+++ b/web/src/components/layout/footer.css
@@ -121,6 +121,17 @@ footer {
         }
     }
 
+    & .bottom_divider {
+        border-top: 1px solid hsl(0, 0%, 20%);
+        width: 20rem;
+        margin-top: 1em;
+        display: block;
+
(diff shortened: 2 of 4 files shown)


phirework:
Hey, thanks so much for contributing this and welcome to the project! Just wanted to give you a heads up that the team is offline and undergoing an org-wide transition until March. I will review this once I'm back online. More info in this Discourse post: https://discourse.mozilla.org/t/update-on-common-voice-mozilla-foundation/74526/

Thanks for your patience!

phirework:
Hello! Thanks for your patience and I appreciate your work. This PR has a few issues that need to be corrected before I can accept.  

1. Please do not add a whole new package just to fix a very minor cosmetic issue, that contributes unnecessary bloat to the codebase. I'm not actually clear why Material UI was added at all, since it doesn't seem like you're actually importing or using it anywhere. 
2. We would prefer that the social media icons remain the way they currently are on site, i.e. no background colour. 
3. We do actually want the dark grey box behind the social media icons to remain. The weird box that George Railean was referring to in the issue is an almost imperceptible purple overlay underneath the checkbox. 
4. Please follow existing naming conventions when naming your elements - the name of the element should be first, the modifier should follow the name of the element, and generally speaking we use dashes rather than underscores. In thise case, the elements would be better referred to as `.divider-vertical` and `.divider-bottom`

phirework (changes requested):
Hello! Thanks for your patience and I appreciate your work. This PR has a few issues that need to be corrected before I can accept.

    Please do not add a whole new package just to fix a very minor cosmetic issue, that contributes unnecessary bloat to the codebase. I'm not actually clear why Material UI was added at all, since it doesn't seem like you're actually importing or using it anywhere.
    We would prefer that the social media icons remain the way they currently are on site, i.e. no background colour.
    We do actually want the dark grey box behind the social media icons to remain. The weird box that George Railean was referring to in the issue is an almost imperceptible purple overlay underneath the checkbox.
    Please follow existing naming conventions when naming your elements - the name of the element should be first, the modifier should follow the name of the element, and generally speaking we use dashes rather than underscores. In thise case, the elements would be better referred to as .divider-vertical and .divider-bottom

Dnouv:
@phirework, thank you for your review; sorry, I was not aware of the standard naming conventions. I have fixed those names and removed the material-ui package.

phirework:
Awesome, thanks for making the mods so quickly!

PR #3015, opened 2021-02-17
Adding CoVoST 2 Translations

jmp84:
This is similar to https://discourse.mozilla.org/t/adding-covost-sentences-to-common-voice/68677 and https://github.com/common-voice/common-voice/pull/2975 but this is adding all translations from CoVoST 2, not only Catalan. The idea is that once the translations are released with audio samples, it will be very easy to create a speech to speech aligned corpus.

The translation were downloaded from https://github.com/facebookresearch/covost, deduped and empty translations and translations containing `REMOVE` were discarded. For translations into English from different source languages, the translations were put together under the `covost2-xx_en.txt` file. I tried [langid](https://github.com/saffsd/langid.py) and [fastText](https://fasttext.cc/docs/en/language-identification.html) but after checking a few examples, I noticed that LID filtering would discard good translations in the correct language so I didn't apply LID (since the translations are done by humans, occurrences of wrong language are rare).


jmp84:
@phirework Would you be able to take a look? Also, I was wondering if you would have an estimate of when those translations would be surfaced in a next release of Common Voice, assuming they are OK to be merged. TIA

ftyers:
Hi @jmp84 there will be a [pause in activity](https://discourse.mozilla.org/t/update-on-common-voice-mozilla-foundation/74526) from the Common Voice team until the middle of March. Probably PRs won't be able to be reviewed until that time. But thanks in the meantime for your work! :)

jmp84:
@ftyers thanks for the reply. I'm following up since per https://discourse.mozilla.org/t/update-on-common-voice-mozilla-foundation/74526 the transition is until early March.

phirework:
Hi @jmp84, thanks for your patience! Can you just confirm that the `xx_en` iterations don't have any overlap or duplication with the existing `en` sentence corpora?

jmp84:
@phirework thanks very much for the reply. There were ~ 2000 sentences that were already in the txt file in the `server/data/en` directory, I just pushed a change to remove them.

phirework:
Awesome, thanks for all this!

moonhouse:
FYI: I have written two posts in the [CoVoST thread on Discourse](https://discourse.mozilla.org/t/adding-covost-sentences-to-common-voice/68677/9) referring to this commit after noticing questionable quality of the sentences in Swedish. The articles by Facebook are a bit vague about exactly how the "professional translations" were made but I have seen a noticeable amount of grammar errors.

There also are sentences that are in correct Swedish but might not be something one would want as training data such as sentences with a lot of English names of persons and places since that would probably suggest to the reader to pronounce them in English. There also are some abbreviations in the data (albeit a small relative number).

The #2975 PR linked here does do normalization of quotes but the data here doesn't seems to have quotes normalized.

There also seems to be a fair number of sentences in the Swedish dataset that are pure English which might suggest that some error in the pre-processing step like missing to filter out parts of the dataset that were in the test partition.

jmp84:
@moonhouse thanks for the investigation. I can apply fasttext LID classification to Swedish and fix it. I can also run normalization of quotes but I'm wondering how that affects the reading part.

ksmammar:
Hi @jmp84 @moonhouse FYI, For Tamil, There are many English words (written in English scripit) in the sentences, It created confusion among contributors, and also the most of the texts has an un natural sentence sturcture (like machine translations).

ftyers:
This is also the case for Turkish.

phirework:
Given the additional feedback, I think the safest thing to do is to reverse this merge until @jmp84 and team are able to provide a cleaner extract. Does that work for everyone?

ftyers:
I'd like to add a comment in general that the kind of corpora used for training machine translation can be of very low quality, especially if they are translated _from_ English into another language by non-professional translators or translators translating out of their usual domain. These should in general be avoided. Text that has been written by native speakers in their own language is better.

jmp84:
@phirework  sure, that works! Let me resubmit a cleaner PR.

jmp84:
@phirework qq to prepare for resubmitting. I'm planning to apply more preprocessing (LID, character set, quote normalization) but I can't really manually check the large volume of text (especially since I don't speak all these languages) so I'm wondering what else can be done to avoid this situation. For example, in the crowdsourcing pipeline, can volunteers simply discard sentences they feel are not fluent enough?

phirework:
@jmp84 We have a [bulk reviewing process](https://github.com/common-voice/common-voice/blob/main/docs/SENTENCES.md#bulk-submission) for language communities to manually review a sample of sentences from a given corpus, and submit only if the error rate is below 5%. You could submit the new sentences as separate PRs per language and ask Discourse/Matrix/etc to help with the reviewing.

PR #3024, opened 2021-02-19
Fix strings from collector

enolp:
There are some mistakes on sentences submitted by the collector


MichaelKohler (commented):
Thanks for filing this PR. The sentences in this file come from the Sentence Collector, and this file gets overwritten every time there is an automatic export (every week). So these changes will need to be done inside of the Sentence Collector. Here's what we're gonna do:

* I will delete the wrong sentences tomorrow and report back here
* Once that is done, you can add the corrected sentences as new sentences through https://commonvoice.mozilla.org/sentence-collector/

Does that sound okay?

enolp:
It is ok. I'm going to re-upload them via collector.

Thank you.

MichaelKohler:
The wrong sentences now have been deleted.

PR #3025, opened 2021-02-19
Update Frases

enolp:
Added period on every string except for every non-? or non-! string, replaced one string, sorted strings, and fixed mistakes.


enolp:
Do not merge.

Issue #3009, opened 2021-02-09
Keep getting "Upload of this clip keeps failing" error with local deployment


jarivk:
**Describe the bug**
Have deployed a local server. Able to access the server using localhost:9000 and record the clips. But after I hit submit, the browser keeps showing "Upload of this clip keeps failing" error.

Docker log of web container shows below error.

```
[BE] { NoSuchBucket: The specified bucket does not exist
[BE]     at Request.extractError (/code/node_modules/aws-sdk/lib/services/s3.js:816:35)
[BE]     at Request.callListeners (/code/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
[BE]     at Request.emit (/code/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
[BE]     at Request.emit (/code/node_modules/aws-sdk/lib/request.js:683:14)
[BE]     at Request.transition (/code/node_modules/aws-sdk/lib/request.js:22:10)
[BE]     at AcceptorStateMachine.runTo (/code/node_modules/aws-sdk/lib/state_machine.js:14:12)
[BE]     at /code/node_modules/aws-sdk/lib/state_machine.js:26:10
[BE]     at Request.<anonymous> (/code/node_modules/aws-sdk/lib/request.js:38:9)
[BE]     at Request.<anonymous> (/code/node_modules/aws-sdk/lib/request.js:685:12)
[BE]     at Request.callListeners (/code/node_modules/aws-sdk/lib/sequential_executor.js:116:18)
[BE]     at Request.emit (/code/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
[BE]     at Request.emit (/code/node_modules/aws-sdk/lib/request.js:683:14)
[BE]     at Request.transition (/code/node_modules/aws-sdk/lib/request.js:22:10)
[BE]     at AcceptorStateMachine.runTo (/code/node_modules/aws-sdk/lib/state_machine.js:14:12)
[BE]     at /code/node_modules/aws-sdk/lib/state_machine.js:26:10
[BE]     at Request.<anonymous> (/code/node_modules/aws-sdk/lib/request.js:38:9)
[BE]   message: 'The specified bucket does not exist',
[BE]   code: 'NoSuchBucket',
[BE]   region: null,
[BE]   time: 2021-02-09T11:24:46.255Z,
[BE]   requestId: '4442587FB7D0A2F9',
[BE]   extendedRequestId: undefined,
[BE]   cfId: undefined,
[...]

PR #3029, opened 2021-02-23
fix "NoSuchBucket" error in local deployment with docker

### LINKED ISSUE #1859: validation problem with the docker installation
with the help of @Gregoor  i managed to install locally the project using the docker setup; recording is working but can not validate, the listen page show the audios but when i click on play button the audio is not stopping and i can only click the NO button and not the OK(button is disabled) to validate the recording.

Any help please.

a-oz:
Server store recordings in `CV_CLIP_BUCKET_NAME`, and the bucket create in s3proxy should have the same name, in this case `common-voice-clips`

This resolves #1859, #3009

### DIFF
--- a/docker/prepare_s3.sh
+++ b/docker/prepare_s3.sh
@@ -2,14 +2,14 @@
 
 # Check if voice-web bucket exists if not create it
 echo "Checking if s3 bucket exists..."
-status=$(curl -XGET s3proxy/common-voice-corpus --write-out "%{http_code}\n" --output /dev/null --silent)
+status=$(curl -XGET s3proxy/common-voice-clips --write-out "%{http_code}\n" --output /dev/null --silent)
 
 echo "GET bucket code: $status"
 
 if [ $status != "200" ]
 then
     echo "Bucket does not exist. Creating one..."
-    status=$(curl -XPUT s3proxy/common-voice-corpus --write-out "%{http_code}\n" --output /dev/null --silent)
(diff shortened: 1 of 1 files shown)


docker/prepare_s3.sh
```diff
@@ -2,14 +2,14 @@
 
 # Check if voice-web bucket exists if not create it
 echo "Checking if s3 bucket exists..."
-status=$(curl -XGET s3proxy/common-voice-corpus --write-out "%{http_code}\n" --output /dev/null --silent)
+status=$(curl -XGET s3proxy/common-voice-clips --write-out "%{http_code}\n" --output /dev/null --silent)
```
jarivk, line 5: Instead of hardcoding bucket name here, should we take it from environment variable CV_CLIP_BUCKET_NAME  and set it in .env-local-docker ?

a-oz: `CV_CLIP_BUCKET_NAME` is an item in `.env-local-docker`, it's an environment variable.

a-oz: `CV_CLIP_BUCKET_NAME` is not an environment variable, it's an item in config file and not available to `docker/prepare_s3.sh`

a-oz: A possible solution is to put in in `.env`, and compose will pick it up as a environment variable.

phirework:
Hello! Thanks for submitting this PR. There's actually some other problems with the Docker config relating to race conditions and the s3proxy container not being ready yet when `prepare_s3.sh` tries to initialize that will be resolved in #3049, so if you're still running into S3 issues be sure to pull from main once that merges!

PR #3054, opened 2021-03-28
Remove sentences with problems

palonso:
The considered problems are:
- Sentences using closing single quotation mark U+2019 (’) as apostrophe. As typewriter apostrophe,  U+0027 ('),  is more commonly used
- Sentences with Latin quotes (« and »). As they normally don't provide pronunciation hints.
- Incorrectly accented sentences

The sentences will be corrected and re-uploaded via the sentence-collector once removed.


MichaelKohler:
The sentences have been removed from the Sentence Collector. You can now upload the corrected sentences. Thanks for reporting!

PR #3112, opened 2021-05-07
Enable sending to Cinchy via SQS

phirework:
* Added feature flag
* Wrapped activity functions using feature flag
* Added config objects for Cinchy
* Separately added SQS endpoint secrets to AWS Secrets Manager for
  sandbox and stage

### DIFF
--- a/server/src/lib/basket.ts
+++ b/server/src/lib/basket.ts
@@ -176,7 +176,6 @@ async function syncFullAccount(currentUserStats: UserEmailStats) {
     two_day_streak: Boolean(computed.two_day_streak),
   };
 
-  // @TODO: How does this handle Basket unsubscribes?
   syncToEmailProvider(data);
 }
 
@@ -184,7 +183,7 @@ async function syncFullAccount(currentUserStats: UserEmailStats) {
 
 /*
  * Sync current account info to email provider if
- * there is updated data and it's been more than 1 hour
+ * there is updated data and it's been more than 5 minutes
  */
 export async function sync(client_id: string) {
   await computeGoals(client_id);
--- a/server/src/lib/basket.ts
+++ b/server/src/lib/basket.ts
@@ -2,6 +2,7 @@ import * as sendRequest from 'request-promise-native';
 import { getConfig } from '../config-helper';
 import { getMySQLInstance } from './model/db/mysql';
 import { computeGoals } from './model/goals';
+import { AWS } from './aws';
 
 type UserEmailStats = {
   client_id: string;
@@ -20,14 +21,15 @@ type UserEmailStats = {
 
 const { BASKET_API_KEY, ENVIRONMENT } = getConfig();
 const db = getMySQLInstance();
+const sqs = AWS.getSqs();
 
-export const API_URL =
+export const BASKET_API_URL =
   ENVIRONMENT == 'prod' || ENVIRONMENT == 'stage'
     ? 'https://basket.mozilla.org'
     : 'https://basket-dev.allizom.org';
 
 const basketConfig = {
-  uri: API_URL + '/news/common-voice-goals/',
+  uri: BASKET_API_URL + '/news/common-voice-goals/',
   method: 'POST',
   headers: {
     'x-api-key': BASKET_API_KEY,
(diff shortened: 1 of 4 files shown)


server/src/lib/basket.ts
```diff
@@ -38,11 +40,39 @@ function toISO(date: string) {
...
+
+/*
+ * Fetch the most recent time the user was synced
+ */
 async function getCurrentStatus(client_id: string) {
   const [[row]] = await db.query(
     `SELECT u.*,
       NOW() AS new_last_active,
-      HOUR(TIMEDIFF(NOW(), u.last_active)) AS hours_elapsed
+      MINUTE(TIMEDIFF(NOW(), u.last_active)) AS mins_elapsed
```
benhohner, line 75: Is there a reason we're changing the sync time from 1 hr to 5 min? Do we know approximately the volume of events being currently sent?

server/src/lib/basket.ts
```diff
@@ -143,19 +176,24 @@ async function updateFullBasket(currentUserStats: UserEmailStats) {
...
-  }).catch(err => console.error(err.message));
+  // @TODO: How does this handle Basket unsubscribes?
+  syncToEmailProvider(data);
 }
 
+
+
+/*
+ * Sync current account info to email provider if
+ * there is updated data and it's been more than 1 hour
```
benhohner, line 187: This might need to be updated if we decide to stick with 5 min.

server/src/lib/basket.ts
```diff
@@ -143,19 +176,24 @@ async function updateFullBasket(currentUserStats: UserEmailStats) {
...
 export async function sync(client_id: string) {
   await computeGoals(client_id);
   const currUserStats = await getCurrentStatus(client_id);
 
   // If no newsletter prefs exist, they don't get email triggers
-  // If last update was within the past hour, do not send another sync
-  if (!currUserStats || parseInt(currUserStats.hours_elapsed) === 0) {
+  // If last update was within the past 5 minutes, do not send another sync to
+  // prevent spamming endpoint
+  if (!currUserStats || parseInt(currUserStats.mins_elapsed) <= 5) {
```
benhohner, line 196: See comment above about 5 min. sync time

benhohner (commented):
Thanks @phirework, I'll pass along to @KalobTaulien for additional review

KalobTaulien:
This all looks good to me. @benhohner your comments were perfect, I had the same questions. Once your comments are addressed, I'll be happy to approve this.

server/src/lib/basket.ts
```diff
@@ -143,19 +176,24 @@ async function updateFullBasket(currentUserStats: UserEmailStats) {
     two_day_streak: Boolean(computed.two_day_streak),
   };
 
-  sendRequest({
-    ...basketConfig,
-    form: data,
-  }).catch(err => console.error(err.message));
+  // @TODO: How does this handle Basket unsubscribes?
```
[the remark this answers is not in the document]
phirework: That note is more about the fact that that newsletter unsubscribes happen in the newsletter management centre `https://www.mozilla.org/en-US/newsletter/existing/<basket_token>` and then there's nothing that gets sent back to the CV platform to indicate that the user has changed their preferences, so it will keep syncing to Cinchy/Acoustic to keep sending goal emails even after they unsubscribe from a newsletter. The easiest fix would be to make a call to Basket before every update to Acoustic but that seems very not-performant and I wanted to think through whether there's a better/different solution.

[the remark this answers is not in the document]
jackielu-mozilla: We will know in Cinchy if someone has changed their preferences, as CTMS is where universally, newsletter preferences are being stored (and provided to Acoustic). Cinchy will be integrated w/ that entire ecosystem.

server/src/lib/basket.ts
benhohner: 5 minutes sounds good for starters! R+

server/src/lib/basket.ts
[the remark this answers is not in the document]
phirework: @jackielu-mozilla Oh I didn't realize that, that's awesome. Thanks for clarifying!

PR #3126, opened 2021-05-22
Fix minor auth0 UI discrepency in DEVELOPMENT.md docs

Zopieux:


docs/DEVELOPMENT.md
```diff
@@ -136,9 +136,9 @@ If you want to work with login-related features (Profile, Dashboard, Goals, ...)
 
 1. Create an [Auth0](https://auth0.com/) account.
 2. Click "Applications" from the dashboard. Create a new one, or use the default application.
-3. Go to "Applications" and click on the Settings icon next to your application.
+3. On "Applications" still, next to your application, click the "three dots" icon, then Settings.
 4. Add `http://localhost:9000/callback` to the "Allowed Callback URLs" list.
-5. If you're using Docker, copy the following keys from the Auth0 application into your configuration file. These are found in the same Settings tab as the "Allowed Callback URLs".
+5. If you're using Docker, copy the following keys from the Auth0 application into your `.env-local-docker` file. These are found in the same Settings tab as the "Allowed Callback URLs".
```
MichaelKohler, line 141: Given that this is not only for docker, I would not change this. The exact config files are mentioned in the two setup descriptions above. If this shouldn't be clear enough, an option might be to add the relevant config file next to "Docker:" and "Local development:" below.

Zopieux: The sentence starts with _If you're using Docker_ though, hence the possible confusion. WDYT of the latest commit?

MichaelKohler: Yep, true. I like it!

MichaelKohler (approved):

PR #3131, opened 2021-05-22
Document the Return key shortcut for submitting clips.

Zopieux:
image


phirework:
@mbransn can I get your eyes on this?

mbransn:
Yeah, I think this submit got lost in translation at some point? Thought we had submit at one point. 

Anywho, good catch @Zopieux! My only change request is to utilize this icon instead to match with the rest of the site UI.
https://feathericons.com/?query=corner-down-left

Zopieux:
Done, please take another look. Updated the screenshot. I had to add a new field to accommodate the icon.

phirework:
Excellent, this looks great. Thanks @Zopieux!

PR #3165, opened 2021-07-04
german wikipedia import: removed abbreviations

ubahnverleih:
Sentences should not include abbreviations, but the german wikipedia import still contained some abbreviations. So I removed all sentences with abbreviations collected in this discourse thread: https://discourse.mozilla.org/t/satzzeichen-in-den-texten/56985/11

abbreviations removed are:

- hl.
- no.
- Mr.
- Ms.
- Mrs.
- geb.
- bspw.
- Dr.
- Fürstl.
- mind.
- Gen.
- Hll.
- evtl.
- Ebf.
- kath.
- Frhr.
- St.
- Geh.
- Vol.
- Jr.
- Bf.
- subsp.
- bzgl.
- franz.
- ahd.
- Ch.
- Fa.
- ev.
- Lt.
- Ltd.
- Inc.


stefangrotz:
Hey, thanks for your work :) If a review of the Div is necessary, I am happy to help (I just can't see the div right now on mobile).  I can already say that all the abbreviations above are valid German abbreviations, and this PR looks good so far.

phirework:
Thanks!

Issue #2105, opened 2019-07-13
Add Cancel button when recording 


12people:
When recording, it'd be nice to have a cancel button for cases when I mess up (say the wrong words or stutter).

Right now, I have to finish recording, then click back to the step, wait for the step to load (it's not immediate), and then re-record. I'd really like a "Cancel" button to avoid having to go back a step.

Issue #1052, opened 2018-06-04
Revisit Keyboard Shortcuts


m-branson:
@Gregoor let's review our current shortcuts vs. those proposed by @georai and make a decision for what to take forward for localization and implementation.

Current:
![screen shot 2018-06-04 at 8 56 43 am](https://user-images.githubusercontent.com/18686986/40928383-19621aa6-67d6-11e8-9591-2bc993bf626b.png)

Proposed:
![screen shot 2018-06-04 at 9 02 53 am](https://user-images.githubusercontent.com/18686986/40928391-1e68f6e6-67d6-11e8-9d07-3d81de21eacb.png)

Gregoor:
Thanks! @jdittrich since you started our interest in adding shortcuts, I wonder what your take on this is. 🙂 

Here's mine: While I'd like to use the arrow keys, I find it slightly confusing that the action they perform are not the inverse of one another. I.e. press left to rerecord previous sentence, but when you press right now you won't get back to the top of the queue but instead stay with that clip and get a different sentence. I think that'd feel a bit strange.

And for completeness sake, the shortcuts for the listen screen are:
**P** - Play/Pause
**Y** - Vote Yes
**N** - Vote No

jdittrich:
Hi @Gregoor, it is hard for me to judge – what is the current UI-element triggering the proposed left-key action?

m-branson:
@jdittrich currently to move back, the UI requires a click on the previous clips at the right of the screen (on desktop web view). In the proposed shortcuts the left arrow would move you back through the clips.

jdittrich:
In this case, I find the mapping off: 

Pressing the left arrow key moves a clip back. The same action is done by buttons on the right. The direction of where these previous clips are is also the right. 

I would assume that future is right and past is left (in LTR languages); however, past clips are right, future clips are left in this UI. This coexists with the idea of "previous screen is left", as with the ← button in the upper left corner.

flekkowich:
Note to this discussion as well from Y-combinator
https://news.ycombinator.com/item?id=17436958

Input: The other thing is that it's very cool to see the "you helped us reach out x% goal" thing but it locks up all the previous / next shortcuts which means I have to switch back to the mouse after 5 entries. 

___

What can we do to create shortcut to make it easy to continue contributions.

@m-branson @Gregoor

m-branson:
>What can we do to create shortcut to make it easy to continue contributions.

[...]

PR #3219, opened 2021-08-05
speak: add "discard ongoing recording" shortcut

olejorgenb:
Allow the user to signal that he messed up and want to begin again.

Could also consider starting recording again immediately? (this would be my personal preference, but without extra work it isn't really visible to the user that something happened. If there was a duration counter visible it would be less of an issue.)

Related to: #1052, #2105

### DIFF
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -287,7 +287,7 @@ shortcut-record-toggle-label = Record/Stop
 shortcut-rerecord-toggle = [1-5]
 shortcut-rerecord-toggle-label = Re-record clip
 
-shortcut-discard-ongoing-recording = Backspace
+shortcut-discard-ongoing-recording = ESC
 shortcut-discard-ongoing-recording-label = Discard ongoing recording
 
 shortcut-submit = Return
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -287,6 +287,9 @@ shortcut-record-toggle-label = Record/Stop
 shortcut-rerecord-toggle = [1-5]
 shortcut-rerecord-toggle-label = Re-record clip
 
+shortcut-discard-ongoing-recording = Backspace
+shortcut-discard-ongoing-recording-label = Discard ongoing recording
+
 shortcut-submit = Return
 shortcut-submit-label = Submit clips
 
--- a/web/src/components/pages/contribution/speak/speak.tsx
+++ b/web/src/components/pages/contribution/speak/speak.tsx
@@ -207,7 +207,7 @@ class SpeakPage extends React.Component<Props, State> {
     this.audio.setVolumeCallback(this.updateVolume.bind(this));
 
     document.addEventListener('visibilitychange', this.releaseMicrophone);
-    document.addEventListener('keyup', this.handleKeyUprerecording);
+    document.addEventListener('keyup', this.handleKeyUp);
 
     if (
       !this.audio.isMicrophoneSupported() ||
@@ -218,7 +218,7 @@ class SpeakPage extends React.Component<Props, State> {
   }
 
   async componentWillUnmount() {
-    document.removeEventListener('keyup', this.handleKeyUprerecording);
+    document.removeEventListener('keyup', this.handleKeyUp);
 
     document.removeEventListener('visibilitychange', this.releaseMicrophone);
     if (!this.isRecording) return;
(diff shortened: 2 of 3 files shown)


web/locales/en/messages.ftl
```diff
@@ -959,4 +962,4 @@ contribution-just-unsure-title = Just Unsure?
 contribution-just-unsure-description = If you come across something that these guidelines don’t cover, please vote according to your best judgement. If you really can’t decide, use the skip button and go on to the next recording.
 
 see-more = <chevron></chevron>See more
-see-less = <chevron></chevron>See less
\ No newline at end of file
+see-less = <chevron></chevron>See less
```
olejorgenb, line 965: This snuck into the commit - I leave it since .editorconfig mandates a final newline

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -229,26 +229,31 @@ class SpeakPage extends React.Component<Props, State> {
...
     } else if (event.code === 'Digit3' || event.code === 'Numpad3') {
-      index = 2;
+      reRecordIndex = 2;
     } else if (event.code === 'Digit4' || event.code === 'Numpad4') {
-      index = 3;
+      reRecordIndex = 3;
     } else if (event.code === 'Digit5' || event.code === 'Numpad5') {
-      index = 4;
+      reRecordIndex = 4;
+    } else if (event.key === "Backspace" || event.key === "x") {
```
olejorgenb, line 245: I used `key` instead of `code` since I think we should respect the user's choice of keyboard layout? I might miss some rationale for using `code` though.

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -744,6 +749,13 @@ class SpeakPage extends React.Component<Props, State> {
                 label: 'shortcut-rerecord-toggle-label',
                 action: this.handleRecordClick,
               },
+              {
+                key: 'shortcut-discard-ongoing-recording',
+                label: 'shortcut-discard-ongoing-recording-label',
+                icon: <span>⌫</span>,
```
olejorgenb, line 755: A proper icon would be nice, but I'm no artist :) If we can't trust that all fonts have this unicode symbol, a textual representation might be better (but "BACKSPACE" overflows the shortcutbox, so it's will require some work)

web/src/services/tracker.ts
```diff
@@ -69,6 +69,7 @@ export function trackRecording(
     | 'record'
     | 'submit'
     | 'rerecord'
+    | 'discard-ongoing'
```
olejorgenb, line 72: I think this is sufficient distinct action from "rerecord" to warrant a separate tracking key, but should maybe go in a separate commit?

olejorgenb:
Sidenote: this project have a very nice first-contributor experience. `docker-compose up` and everything works!

phirework:
So glad to hear you had a good with the docker container, that has been a long journey of trial and error 😁 I'm going to tag in @mbransn for a quick design review. Also, I'm thinking this should be ESC instead of / in addition to backspace, thoughts?

olejorgenb:
Yeah, not sure why I didn't think of ESC. It's one of my favorite keys :D

mbransn:
Thanks for this suggestion @olejorgenb :) I agree that having a shortcut to restart recording is a nice addition to the re-record option at the end of the clip. ++ to using `ESC` for the action key and the UI tag.

web/src/components/pages/contribution/speak/speak.tsx
phirework: Agreed! Would you be interested in submitting a separate PR to fix the rest of these as well?

web/src/components/pages/contribution/speak/speak.tsx
phirework: If we're changing to ESC, I think just using `ESC` would be fine for the "icon"

phirework (changes requested):
As discussed - can you change the trigger key from Backspace to Escape? otherwise this PR is good to go1 Thanks so much :D

web/src/components/pages/contribution/speak/speak.tsx
olejorgenb: Its probably easier if you just do it :)

web/src/components/pages/contribution/speak/speak.tsx
olejorgenb: Agreed

phirework:
This looks great, thanks!

PR #3298, opened 2021-10-13
Add Esperanto (eo) Wikipedia Extract re-run on new articles

MichaelKohler:
Re-run on the Esperanto Wikipedia on newly created articles since the last extraction we did at the end of 2019. No validation needed as the rules file previously has been validated.

Extraction pipeline: https://github.com/common-voice/cv-sentence-extractor/actions/runs/1338878795


stefangrotz:
The evaluation can be found here: https://github.com/common-voice/cv-sentence-extractor/pull/158

We found around 5% error rate. Since this is a relatively small extraction, I will remove some obviously wrong sentences manually in another pull request after this one got merged.

PR #3301, opened 2021-10-17
Sort languages in the Data Sets > Languages dropdown, accounting for localization.

### LINKED ISSUE #3295: Languages should be sorted alphabetically based on what users see
**Describe the bug**
On the [main website's dataset's page](https://commonvoice.mozilla.org/en/datasets), the languages are sorted based on their ISO code, but the text is the actual language name in English. This makes it hard to find the language user is after. For instance I was looking for "Persian" among items starting with "P" and when I didn't find it I tried to find "Farsi" among "F"s and at the first look I failed to find it but I was sure that it exists because I myself contributed a lot, so I grep the page source code and found:

```html
<option value="eu">Basque</option>
<option value="fa">Persian</option>
<option value="fi">Finnish</option>
<option value="fr">French</option>
<option value="fy-NL">Frisian</option>
```

This would affect a lot of languages including but not limited to Welsh (`cy`), Spanish  (`es`), 

So based on a quick analysis these are the distances that the current situation is compared to what users expect them to see:

image

As you can see the distance can be extremely far for some of the entries!

**To Reproduce**

1. Open https://commonvoice.mozilla.org/en/datasets

**Expected behavior**
Languages are sorted alphabetically and based on the visible text to user and not the language codes that is not visible to user.

**Screenshots**

image

**Desktop or Mobile (please complete the following information):**
 - OS: Manjaro
 - Browser: Firefox
 - Version 93.0 (64-bit)

 **Additional Hardware (were you using headphones, an external speaker or an external microphone?):**
--irrelevant--

**Additional context**
--irrelevant--

ChristianMMacy:
This PR resolves #3295.

### DIFF
--- a/web/src/components/language-select/language-select.tsx
+++ b/web/src/components/language-select/language-select.tsx
@@ -1,4 +1,4 @@
-import {Localized, withLocalization, WithLocalizationProps} from '@fluent/react';
+import { Localized, withLocalization, WithLocalizationProps } from '@fluent/react';
 import Downshift from 'downshift';
 import * as React from 'react';
 


ChristianMMacy:
@phirework : I'd be happy to. I'll take a look tonight and see about the best way to handle that.

ChristianMMacy:
I went with a hook that returns two forms of a sorted list of locales: one with just the locales and the other with the locales and their localized names.

I didn't want to make too many assumptions, so I tried to keep the implementation of the sorted lists similar to what's already in place—I just swapped out the unsorted/locale-sorted lists for name-sorted versions.

I'm inclined to believe that there are other places with lists of locales that have their own sorting logic, but I'm wary of making any PR too big. However, if you'd like me to pursue this further I can do so either as part of this or as a separate scope of work.

phirework:
Excellent, thanks so much! This will go out tomorrow :)

PR #3309, opened 2021-10-20
Re-add fluent syntax dependency to fix import-locales script

MichaelKohler:
The recent dependency upgrade PR removed the `fluent-syntax` package, which broke the `yarn import-locales` script. This re-adds the dependency from the old name, as the package under `@fluent/syntax` requires Node >= 12.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "repository": "mozilla/voice-web",
   "private": true,
   "dependencies": {
+    "fluent-syntax": "^0.14.0",
     "http-status-codes": "^1.4.0",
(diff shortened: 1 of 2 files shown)


phirework:
Ah shoot, sorry about that. We're still on node 10 because I haven't had time to triple-check that upgrading won't have any adverse consequences, primarily. I've actually got upgrading Node, React, and TS on my board for this upcoming sprint. In the meantime we can go back to the old `fluent-syntax` package - do you want me to re-run `import-locales` manually or would you prefer to trigger it from SC?

PR #3320, opened 2021-10-23
Update Terms and Privacy content on Locale change

### LINKED ISSUE #3287: [Bug] Language changes are not reflected to main content on Terms and Privacy pages.
On the "terms" and "privacy" pages, when you change the language from the top right pull-down menu, the UI text like menu items changes, but the main content does not change.

When you refresh the page, correct content is displayed.

ChristianMMacy:
Resolves #3287 

There are two commits in this PR:
* The latest is a state-only change that switches to `ComponentDidUpdate` so that Redux changes trigger a redraw of the inner HTML.
* The prior commit accomplishes the same thing but refactors the component to a functional one.

I wasn't sure which would be preferred, so I included both--deferring to minimizing unnecessary change. If the functional version is preferred I can re-push with just that. If the class-based version is preferred and the extra commit needs to be pruned, I'm happy to do that as well.

For my part, I like the simplicity of functional components and their integration with the Hooks API. Unfortunately the spacing (between parens and braces) isn't right on that commit. I'll fix it if we choose that one.

### DIFF
--- a/web/src/components/pages/document-page.tsx
+++ b/web/src/components/pages/document-page.tsx
@@ -3,43 +3,47 @@ import { connect } from 'react-redux';
 import API from '../../services/api';
 import StateTree from '../../stores/tree';
 import { Spinner } from '../ui/ui';
-import { useEffect, useState } from "react";
 
 interface PropsFromState {
   api: API;
   locale: string;
 }
 
-type Props = {
-  name: 'privacy' | 'terms' | 'challenge-terms'
-} & PropsFromState
-
 /**
  * Renders an HTML document determined by the <c>name</c> prop.
  * @param {{name:string} & PropsFromState} props Props passed to the component.
  */
-function DocumentPage(props: Props) {
-  const {api, name, locale} = props;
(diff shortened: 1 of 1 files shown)


phirework:
Hello! So sorry this completely fell off my radar. I prefer the functional version as well - would you mind reverting?

ChristianMMacy:
No problem at all. Could you point me to the ESLint config? I didn't find the presets in the places I normally look. (`package.json` and `.eslintrc`)

phirework:
We don't actually use ESLint, we use Prettier - if you run `yarn run pretty-quick --staged` on your staged files it should clean those up. There's a pre-commit hook set up with husky - I'm not sure why it isn't running for you.

ChristianMMacy:
I'm not sure if IntelliJ was messing things up with the pre-commit hook, but I added the Prettier plugin and it's all gravy, now. Let me know if you'd like to see anything else from this PR :-)

phirework:
Excellent, thanks so much! 🎉

PR #3328, opened 2021-10-29
Fix animation of cards to go to the left side in LTR

### LINKED ISSUE #3297: Small quirk: slide animation on next should go to the left
**Describe the bug**
The slide animation currently goes to the right once you go to the next item, in both listen and speak mode.
In my opinion, it should go left, as the next item is "waiting" right off the screen. 

**To Reproduce**
1. Go to 'Listen' or Speak
2. Complete an item

**Expected behavior**
Item leaves the screen on the left side, new item comes from the right.

**Current behavior**
Items moves to the right, new one comes from the left. 

**Screenshots**

https://user-images.githubusercontent.com/230782/137162319-6ffc8f53-afbb-4886-845f-609c669822e1.mp4

**Mobile:**
 - OS: Android 11
 - Browser Fx Mobile Beta 94.0.0-beta.2 (Build #2015839051)

guerda:
Currently, the animation when reviewing or speaking a sentence is that the card leaves to the right side of the screen (LTR direction). This is counterintuitive from the animation point, as the slider on mobile looks like it should leave on the left side.

Therefore I changed the direction in the animation,
Fixes #3297

Please let me know if there's anything missing or wrong with this PR.

### DIFF
--- a/web/src/components/pages/contribution/contribution.tsx
+++ b/web/src/components/pages/contribution/contribution.tsx
@@ -455,7 +455,7 @@ class ContributionPage extends React.Component<Props, State> {
                           transform: [
                             `scale(${isActive ? 1 : 0.9})`,
                             `translateX(${
-                              (document.dir == 'rtl' ? -1 : 1) *
+                              (document.dir == 'rtl' ? 1 : -1) *
                               (i - activeSentenceIndex) *
                               -130
                             }%)`,


phirework:
This also changes the direction of the animation as well as the card queue on desktop, which visually interferes with the review panel. Can you find a way to make this happen only on mobile? Otherwise I don't think this is a change that we actually want design-wise - @mbransn to confirm 

Screen Shot 2021-11-03 at 12 37 12 PM

guerda:
Thanks for the feedback! I can try, I'm  not really sure if that's easy to do. Am 03.11.2021 17:38 schrieb JZ ***@***.***>:
This also changes the direction of the animation as well as the card queue on desktop, which visually interferes with the review panel. Can you find a way to make this happen only on mobile? Otherwise I don't think this is a change that we actually want design-wise - @mbransn to confirm

guerda:
While we're at it: Would it be in general a wanted change if the desktop view is still intact?

phirework:
I think just changing the mobile direction would be fine. What you would probably need to do is turn the in-line styles into multiple classes, add/remove classes depending on the values of `isActive`, `activeSentenceIndex`, and `document.dir`, and target transform direction based on viewport. I'm not sure if the refactoring is worth it for a minor cosmetic change like that.

mbransn:
Hi both :) agreed that the mobile direction change would make sense, it was a nice-to-have UI polish that was never prioritized on production as it was inverse to the desktop experience. ++ to keeping the desktop direction as currently implemented and improving mobile if technically feasible. thank you!

guerda:
Thank you for the feedback, Megan. I will try to have a look.

jessicarose:
Thanks so much for your work on this PR, it's really exciting to see someone passionate about improving the UI flow. And my apologies for the pretty extreme wait you've had in hearing from the team!

We've seen some pretty dramatic UI changes over the past 2 years that would render a lot of this work inapplicable in it's current state. I'll close this now but if you're ever interested in contributing in the future, I can promise a shorter wait for feedback!

PR #3339, opened 2021-11-08
Fix & improve OCR numbers recognition

yks72p:


yks72p:
accidental PR from other fork. sorry :)

PR #3343, opened 2021-11-13
updated Uzbek localization

abbosjon-kudratov:


phirework:
Hi there, all localization updates have to go through the Pontoon project, otherwise they will be overwritten on next sync. https://pontoon.mozilla.org/uz/common-voice/

PR #3406, opened 2022-01-17
fix(onboarding): various changes for new Apple Macbooks

zaccolley:
See commit list for individual changes. Can break these out into seperate PRs if needed.

These are the changes I needed to do to get the project running on a MacBook Pro (13-inch, M1, 2020) with a M1 Chip.


package.json
```diff
@@ -4,19 +4,6 @@
   "license": "MPL-2.0",
   "repository": "mozilla/voice-web",
   "private": true,
-  "dependencies": {
```
zaccolley, line 7: these dependancies are used in the dev scripts OR `http-status-codes` which belongs in the `server/package.json` depdendancies

mozgzh:
Curious about renaming `.env-local-docker` to `.env-local-docker.example` but otherwise looks good

zaccolley:
> Curious about renaming `.env-local-docker` to `.env-local-docker.example` but otherwise looks good

Because each user will and can change `.env-local-docker` (for example to add the Auth0 credentials), we don't want someone accidentally committing environment variables. I think maybe we can simplify/combine the `config.json` and this environment file though cause I don't know why we have two.

zaccolley:
Looks like my change to the http codes import is failing will push a fix for that.

mozgzh:
> > Curious about renaming `.env-local-docker` to `.env-local-docker.example` but otherwise looks good
> 
> Because each user can change `.env-local-docker` (for example to add the Auth0 credentials), we don't want someone accidentally committing environment variables. I think maybe we can simplify/combine the `config.json` and this environment file though cause I don't know why we have two.
> 
> And adding `.example` makes sure Docker doesn't process it as a file.
> Can change/remove that commit.

I like that - committing the file with individual options is annoying. I think if we change the file to  be `.example`, we'd need docs to explain that the user would need to copy the `.example` file to `.env-local-docker` themselves since the app expects it to exist.

As for both the config.json and .env file - just a guess, but maybe the config.json is for JS app specific config and the .env is for local environment config (like docker)

zaccolley:
> I like that - committing the file with individual options is annoying. I think if we change the file to be `.example`, we'd need docs to explain that the user would need to copy the `.example` file to `.env-local-docker` themselves since the app expects it to exist.

Does this work? https://github.com/common-voice/common-voice/pull/3406/files#diff-d9657425e036fb0ad2912c2c24ea793648eb86de5b40aa0844276dd08ed01d87R27

> As for both the config.json and .env file - just a guess, but maybe the config.json is for JS app specific config and the .env is for local environment config (like docker)

Ah that's true, the Docker env file is for everything Docker is running and config.json is just web. Makes sense.

phirework (approved):

PR #3410, opened 2022-01-18
feat(linting): add eslint support

zaccolley:
+ Fix a issue with yarn workplaces
+ Add ESlint with support for
  + Various node linting
  + React (with some accessibility linting)
  + Typescript
  + Prettier


zaccolley:
Interestingly the tests we have started failing based on type errors:

```bash
[1]     TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
[1]     src/lib/api.ts:171:16 - error TS2345: Argument of type 'string | string[] | ParsedQs | ParsedQs[]' is not assignable to parameter of type 'string'.
[1]       Type 'string[]' is not assignable to type 'string'.
```

Why wasn't it failing before?

zaccolley:
I'm not that experienced with Typescript so if either of you think this should be investigated more let me know. I think it's a case of fixing the issue and moving on though I hope. @mozgzh @phirework

package.json
```diff
@@ -46,8 +55,11 @@
     "build-common": "cd common && yarn build && cd ..",
     "build-maint": "cd maintenance && yarn build && cd ..",
     "import-locales": "node scripts/import-locales",
-    "lint": "prettier \"**/*.{ts,tsx}\"",
```
zaccolley, line 49: ESLint now lints for Prettier so we don't need to do this check in the lint command anymore

.eslintrc.js
```diff
@@ -0,0 +1,46 @@
...
+  extends: [
+    'eslint:recommended',
+    'plugin:react/recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:jsx-a11y/recommended',
+    'prettier',
+  ],
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    ecmaFeatures: {
```
zaccolley, line 18: React options

zaccolley:
I'm going to fix the Typescript errors that are blocking this PR

server/src/auth-router.ts
```diff
@@ -74,41 +87,34 @@ if (DOMAIN) {
...
       clientID: CLIENT_ID,
       clientSecret: CLIENT_SECRET,
-      callbackURL:
-        (({
-          stage: 'https://commonvoice.allizom.org',
-          prod: 'https://commonvoice.mozilla.org',
-          dev: 'https://dev.voice.mozit.cloud',
-          sandbox: 'https://sandbox.voice.mozit.cloud',
-        } as any)[ENVIRONMENT] || '') + CALLBACK_URL,
-      scope: 'openid email',
```
zaccolley, line 84: scope isn't a valid parameter and is never used at this stage, its a param for the `passport.authenticate` method. http://www.passportjs.org/packages/passport-auth0/

I assume this was never working and the default is email, do we let people signup with OpenID?

server/src/auth-router.ts
```diff
@@ -48,15 +51,25 @@ router.use(
...
+};
 
+if (DOMAIN) {
+  /*
+    https://github.com/auth0/passport-auth0/issues/53#issuecomment-379463280
+
+    TODO: Explore if we can remove this patch now:
+    https://github.com/auth0/passport-auth0/issues/53#issuecomment-728830811
+  */
+  // @ts-ignore: Property 'authorizationParams' does not exist on type 'StrategyInternal'.
```
zaccolley, line 71: We want to avoid using ts-ignore at all costs really, here I'm using it cause theres a TODO to explore doing a better fix.

server/src/auth-router.ts
```diff
@@ -74,41 +87,34 @@ if (DOMAIN) {
...
+    (profile: any, done: any) => done(null, profile)
   );
 
   passport.use(strategy);
 } else {
-  console.log('No Auth0 configuration found');
+  console.error('No Auth0 configuration found');
+}
+
+function parseState(request: Request) {
```
zaccolley, line 100: The items that are on the `request.query` object are from the `qs` library and so could be an array or a string. To keep Typescript happy we have to check for that edgecase.

server/src/lib/api.ts
```diff
@@ -524,4 +526,21 @@ export default class API {
       await this.model.db.getAccents(client_id, params?.locale || null)
     );
   };
+
+  private getCountQueryParam = (request: Request) => {
```
zaccolley, line 530: The items that are on the `request.query` object are from the `qs` library and so could be an array or a string. To keep Typescript happy we have to check for that edgecase.

I also noticed we don't check if the number is a number so did that too.

web/src/components/layout/layout.tsx
```diff
@@ -26,7 +26,7 @@ import LocalizationSelect from './localization-select';
 import Logo from './logo';
 import Nav from './nav';
 import UserMenu from './user-menu';
-import * as cx from 'classnames';
+import cx from 'classnames';
```
zaccolley, line 29: Typescript couldn't find the types when we imported it with `* as`

zaccolley:
The blocking Typescript changes have been fixed but I want to write some tests that covers that changed code where possible.

server/src/auth-router.ts
phirework: People can sign up using Github, Google, Firefox Accounts, or email, which is split between passwordless (for regular users) and LDAP connected to your Mozilla SSO for Mozilla employees. I would leave scope in here unless you're absolutely certain none of the SSO options require it. 

Also, if you remove the callback URLs I think it'll break login on dev/sandbox.

phirework: Ah nvm, I see where you added it back in below.

web/src/components/layout/layout.tsx
phirework: I wonder why it was fine prior to adding a linter.

phirework (approved):

zaccolley:
I've removed any extra linting changes as requested so these changes are the only ones that blocked the Typescript compliler. `server/src/auth-router.ts`, `server/src/lib/api.ts` and `server/src/lib/clip.ts`. This is ready for review @mozgzh

.eslintrc.js
```diff
@@ -0,0 +1,49 @@
...
+  rules: {
+    // turn on errors for missing imports
+    'import/no-unresolved': 'error',
+
+    // warn us if we're using @ts-ignore etc..
+    '@typescript-eslint/ban-ts-comment': 'warn',
+  },
+  settings: {
+    react: {
+      version: 'detect',
```
mozgzh, line 40: nit: getting a warning:  `Warning: React version was set to "detect" in eslint-plugin-react settings, but the "react" package is not installed. Assuming latest React version for linting.`

Could just use "latest" instead

maintenance/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "maintenance",
+  "private": true,
```
mozgzh, line 3: :rocket:

mozgzh:
Got the linting up and running locally - nicely done! As expected, it complains of a lot of errors. However, I think there might be a config issue due to CV being a monorepo setup. 

Since `eslintrc.js` is in the root dir and the individual apps are pulling in packages in their respective node_modules, eslint isn't matching the imported packages with the correct node_modules.  

For example, after running `yarn lint:web`, the error: `Unable to resolve path to module 'redux` is displayed for `/web/src/stores/user.ts` even though `/web/node_modules/` has redux installed.

zaccolley:
@mozgzh I'm not getting that error, I'm on Yarn `1.22.17`. Could you try `yarn clean && yarn install`?

web/src/components/layout/layout.tsx
zaccolley: I think just some combination of the config files meant Typescript wasn't working correctly which is weird.

mozgzh:
> @mozgzh I'm not getting that error, I'm on Yarn `1.22.17`. Could you try `yarn clean && yarn install`?

You're not getting that linting message after running lint:web?

zaccolley:
> You're not getting that linting message after running lint:web?

No mention of redux anywhere and this is the errors/warnings I get for the file you mentioned.

Screenshot 2022-01-25 at 14 09 02

Could you try `npx jest --cacheClean` then run the tests again. I've had to do that for some things while changing testing config etc.

mozgzh:
Wild, rm'd node_modules and reinstalled and now those errors aren't popping up. Magic.

mozgzh (approved):

PR #3415, opened 2022-01-20
fix(LocalizationSelect): various bugs

### LINKED ISSUE #1388: Keyboard accessibility: Language drop-down
[Environment]:
Windows 10, Chrome 68, Firefox 61, latest Edge

[Steps to reproduce]:
1. Open https://voice.mozilla.org/en/new
2. Press the Tab key to navigate to the language drop-down
3. Open the drop-down using the keyboard
4. After the drop-down is opened press on different letter keys

[Expected results]:
3.T he drop-down can be opened using the keyboard
4. Different languages should be selected

[Actual results]:
3. The drop-down can't be opened using the keyboard
4. The languages are not selected using the keyboard

zaccolley:
Fixes #1388 | Linked to #3422 | Linked to #3423
JIRA ticket: https://mozilla-hub.atlassian.net/browse/OI-2199

---

+ Fixes to existing `LocalizationSelect` components
   + Missing a label - this now added making no visual difference
   + This replaces the `LabeledSelect` component that was confusing
 
 + Added new `LocalizationSelectComplex` component
   + Now keyboard accessible like a `<select>`
   + Styling fixed
     + Added highlight styles that further match a normal `<select>`
     + The sizing of the dropdown button was a bit wonky before
     + Hovering on button would lose the down arrow
     + Was broken on RTL
     
## Screenshots

| LTR (Before) | LTR | RTL |
| - | - | - |
|  |  |  |

## Testing:

+ [x] Add some basic component tests
+ [x] Do some manual testing across devices and browsers
+ [x] Test with screenreader
+ [x] Test with RTL languages


web/src/components/localization-select/localization-select.css
```diff
@@ -0,0 +1,114 @@
...
+        background: white;
+        text-transform: uppercase;
+        font-size: var(--font-size-xs);
+        font-weight: 600;
+        letter-spacing: 0.8px;
+        color: var(--warm-grey);
+        cursor: pointer;
+    }
+
+    & .button {
```
zaccolley, line 34: This way of doing styles is a bit dangerous because we can have classes leaking into other places. Something to think about in future.

web/src/components/localization-select/localization-select.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import { LOCALES, LOCALES_WITH_NAMES } from '../../services/localization';
+import { LabeledSelect } from '../ui/ui';
+
+import './localization-select.css';
+
+interface Props {
+  locale?: string;
+  onLocaleChange: (props: any) => any;
+  isSimple?: boolean;
```
zaccolley, line 11: Can remove this, not used anymore

web/src/components/localization-select/localization-select-complex.tsx
```diff
@@ -0,0 +1,83 @@
+import * as React from 'react';
```
zaccolley, line 1: A lot of this new component is lifted directly from the [`useSelect` docks for downshift](https://github.com/downshift-js/downshift/tree/master/src/hooks/useSelect).

web/src/components/localization-select/localization-select.css
```diff
@@ -0,0 +1,114 @@
+.localization-select {
+    position: relative;
+    font-family: 'Inter', sans-serif;
+
+    &::before {
+        content: '';
+        width: 100%;
+        height: 100%;
+        position: absolute;
+        background: url('./globe.svg') no-repeat left 13px center;
```
zaccolley, line 10: Curious if we could use the `<GlobeIcon />` component and remove the additional icon code

web/src/components/visually-hidden/visually-hidden.css
```diff
@@ -0,0 +1,13 @@
+.visually-hidden {
```
zaccolley, line 1: I should add the reference to GOV.UK's code here as this is where it's stole from

web/src/components/ui/ui.tsx
```diff
@@ -91,10 +103,14 @@ const LabeledFormControl = React.forwardRef(
           props.disabled ? 'disabled' : '',
         ].join(' ')}
         {...props}>
-        <span className="label">
-          {required && '*'}
-          {label}
-        </span>
+        {isLabelVisuallyHidden ? (
```
zaccolley, line 106: Here if the label is hidden we use the new `VisuallyHidden` component. Otherwise we render the label as expected. I've wrapped the * with `aria-hidden` as I don't want that to be read out my assisitive technology.

web/src/components/localization-select/localization-select.css
phirework: Can you expand a bit more on what risk you're envisioning?

zaccolley: For example, here you might assume the element with the class `.button` in `LocalizationSelect` is only styled by this code here, but it's also styled by a file in `ui.css` which considering we don't link to that at any point in this component is confusing.

mozgzh (approved):
Looks good, huge upgrade over the existing implementation! Looking forward to QAing in dev

PR #3420, opened 2022-01-22
Fixed malformed link in About page

bluedaze:
## Pull Request Form

Malformed link tag in new About page content as stated in this issue here:

https://github.com/common-voice/common-voice/issues/3416#issue-1110076624


zaccolley:
Thank you this is great!

PR #3421, opened 2022-01-23
fix incomplete closing tag ctaLink in about-playbook-what-is-language-content-3

petercpg:
### Type of Pull Request

- [x] Other

In the latest batch of exported string there is one string `about-playbook-what-is-language-content-3` has a incomplete closing HTML tag `</ctaLink` instead of `</ctaLink>`. This commit fixes en locale, as well as those locales already translated but did not have the tag corrected in translation. 

* me myself.


zaccolley:
Thank you!

PR #3423, opened 2022-01-24
feat(accessibility): add component to hide things visually

zaccolley:
Add `VisuallyHidden` component. This hides content visually but allows assitive technology to access it still.
```js
<VisuallyHidden>This is hidden weee!</VisuallyHidden>
```

Sometimes we want something to not show visually but availablefor assistive technologies such as screenreaders.

These styles are from GOV.UK's work, known for their accessibility work.

### DIFF
--- a/web/src/components/visually-hidden/visually-hidden.css
+++ b/web/src/components/visually-hidden/visually-hidden.css
@@ -1,5 +1,5 @@
 /*
-  stolen from GOV.UK
+  based on code from GOV.UK
   https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/helpers/_visually-hidden.scss
 */
 .visually-hidden {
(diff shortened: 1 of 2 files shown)


web/src/components/visually-hidden/visually-hidden.css
```diff
@@ -0,0 +1,17 @@
+/*
+  stolen from GOV.UK
+  https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/helpers/_visually-hidden.scss
```
zaccolley, line 3: This is licensed under MIT which is allowed in our MPL

mozgzh: Maybe "Based on" is a little less on the nose :thinking:

mozgzh (approved):
:rocket:

PR #3426, opened 2022-01-26
[Blocked][WIP] feat(stats-card): remember locale options 

zaccolley:
+ Refactor localStorage hook to be used in for `StatsCard`
+ Implement the hook into `StatsCard` for locale changes
+ Fix linting in files that PR touches

### DIFF
--- a/web/src/components/pages/dashboard/stats/stats-card.tsx
+++ b/web/src/components/pages/dashboard/stats/stats-card.tsx
@@ -4,10 +4,14 @@ import { useState, useEffect } from 'react';
 import LanguageSelect, {
   ALL_LOCALES,
 } from '../../../language-select/language-select';
+import { useLocalStorageState } from '../../../../hooks/store-hooks';
 
 import './stats-card.css';
 
+const DEFAULT_LOCALE_OPTION = ALL_LOCALES;
+
 export default function StatsCard({
+  id,
   className,
   title,
   iconButtons,
@@ -17,6 +21,7 @@ export default function StatsCard({
   scrollable,
   currentLocale,
 }: {
+  id?: string;
   className?: string;
   title: string;
   iconButtons?: React.ReactNode;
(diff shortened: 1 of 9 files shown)


web/src/components/pages/dashboard/stats/stats.tsx
```diff
@@ -37,22 +40,19 @@ const StatsPage = ({ allGoals, dashboardLocale }: Props) =>
           key="contribution"
           title="contribution-activity"
           currentLocale={dashboardLocale}
-          tabs={['you', 'everyone'].reduce(
```
zaccolley, line 40: Found this code too complicated for what gains it gives us

web/src/components/pages/dashboard/stats/stats-card.tsx
```diff
@@ -17,6 +21,7 @@ export default function StatsCard({
   scrollable,
   currentLocale,
 }: {
+  id?: string;
```
zaccolley, line 24: Thinking maybe this should be `storageKey`? Matches the other usage of it. Doesn't make much sense for the API of StatsCard though. A unique ID is what it is for StatsCard...

web/src/components/pages/dashboard/stats/stats-card.tsx
```diff
@@ -26,11 +31,15 @@ export default function StatsCard({
...
-  const [locale, setLocale] = useState(ALL_LOCALES);
+  const [locale, setLocale] = useLocalStorageState(DEFAULT_LOCALE_OPTION, id);
   const [selectedTab, setSelectedTab] = useState(Object.keys(tabs)[0]);
-  useEffect(() => setLocale(currentLocale ? currentLocale : ALL_LOCALES), [
-    currentLocale,
-  ]);
+  const isDefaultOptionSelected = locale === DEFAULT_LOCALE_OPTION;
+
+  useEffect(() => {
+    if (isDefaultOptionSelected && currentLocale) {
```
zaccolley, line 39: We would overrite the localstorage with the router's locale without this check.

web/src/hooks/store-hooks.ts
```diff
@@ -41,18 +41,43 @@ export function useIsSubscribed() {
...
+      // remove item from storage if it's the default option
+      if (value === defaultValue) {
+        window.localStorage.removeItem(key);
+        return;
+      }
+
       window.localStorage.setItem(key, JSON.stringify(value));
     } catch (e) {
-      console.warn(`A sessionStorage error occurred ${e.message}`);
+      // silently fail if no localStorage or JSON stringify fails
```
zaccolley, line 80: If a browser doesn't support localStorage we don't really need to know, the hook will instead fallback to not remembering state on refresh which is degraded experience but not worth creating logs for

web/src/hooks/store-hooks.ts
```diff
@@ -41,18 +41,43 @@ export function useIsSubscribed() {
...
+    }
   });
 
   useEffect(() => {
+    if (!key) {
+      return;
+    }
+
     try {
+      // remove item from storage if it's the default option
```
zaccolley, line 72: If we were to update the default values in our code and we don't remove here we could have instances where a client is out of sync and think it's not at default.

zaccolley:
Duplicate of #3427

PR #3427, opened 2022-01-26
feat(stats-card): remember locale options

zaccolley:
JIRA ticket: https://mozilla-hub.atlassian.net/browse/OI-2186

+ Refactor localStorage hook to be used in for `StatsCard`
+ Implement the hook into `StatsCard` for locale changes
+ Fix linting in files that PR touches

### DIFF
--- a/web/src/components/pages/dashboard/challenge/leaderboard-card.tsx
+++ b/web/src/components/pages/dashboard/challenge/leaderboard-card.tsx
@@ -41,7 +41,7 @@ export default function LeaderboardCard({
   const leaderboardRef = useRef(null);
   return (
     <StatsCard
-      key="leaderboard"
+      id="challenge-leaderboard"
       className="leaderboard-card"
       title={title}
       challenge={true}
--- a/web/src/components/pages/dashboard/stats/leaderboard-card.tsx
+++ b/web/src/components/pages/dashboard/stats/leaderboard-card.tsx
@@ -1,8 +1,4 @@
-import {
-  Localized,
-  withLocalization,
-  WithLocalizationProps,
-} from '@fluent/react';
+import { Localized } from '@fluent/react';
 import * as React from 'react';
 import { useRef, useState } from 'react';
 import { connect } from 'react-redux';
@@ -11,13 +7,16 @@ import API from '../../../../services/api';
 import { trackDashboard } from '../../../../services/tracker';
 import { Locale } from '../../../../stores/locale';
 import StateTree from '../../../../stores/tree';
-import { User } from '../../../../stores/user';
+import {
+  User,
+  VISIBLE_FOR_NONE,
+  VISIBLE_FOR_ALL,
+} from '../../../../stores/user';
 import URLS from '../../../../urls';
 import { LocaleLink, LocalizedGetAttribute } from '../../../locale-helpers';
 
 import {
   BookmarkIcon,
-  CheckIcon,
   CrossIcon,
   EyeIcon,
   EyeOffIcon,
@@ -28,7 +27,6 @@ import {
 } from '../../../ui/icons';
 import { Avatar, Toggle } from '../../../ui/ui';
 import StatsCard from './stats-card';
-import { isProduction } from '../../../../utility';
 
 import './leaderboard.css';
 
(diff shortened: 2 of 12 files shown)


web/src/components/pages/dashboard/stats/leaderboard-card.tsx
```diff
@@ -57,22 +55,6 @@ const FetchRow = (props: React.HTMLProps<HTMLButtonElement>) => (
   </li>
 );
 
-const RateColumn = withLocalization(
```
zaccolley, line 60: This code is never used

web/src/components/pages/dashboard/stats/stats.tsx
```diff
@@ -33,26 +36,23 @@ const StatsPage = ({ allGoals, dashboardLocale }: Props) =>
 
       <div className="cards">
         <StatsCard
+          id="stats-contribution"
           scrollable
-          key="contribution"
           title="contribution-activity"
           currentLocale={dashboardLocale}
-          tabs={['you', 'everyone'].reduce(
```
zaccolley, line 40: Found this code too complicated for what gains it gives us

web/src/components/pages/dashboard/stats/stats-card.tsx
```diff
@@ -17,6 +21,7 @@ export default function StatsCard({
   scrollable,
   currentLocale,
 }: {
+  id?: string;
```
zaccolley, line 24: Thinking maybe this should be storageKey? Matches the other usage of it. Doesn't make much sense for the API of StatsCard though. A unique ID is what it is for StatsCard...

web/src/hooks/store-hooks.ts
```diff
@@ -41,18 +41,43 @@ export function useIsSubscribed() {
...
   });
 
   useEffect(() => {
+    if (!key) {
+      return;
+    }
+
     try {
+      // remove item from storage if it's the default option
+      if (value === defaultValue) {
```
zaccolley, line 73: We would overrite the localstorage with the router's locale without this check.

web/src/hooks/store-hooks.ts
```diff
@@ -41,18 +41,43 @@ export function useIsSubscribed() {
...
+        window.localStorage.removeItem(key);
+        return;
+      }
+
       window.localStorage.setItem(key, JSON.stringify(value));
     } catch (e) {
-      console.warn(`A sessionStorage error occurred ${e.message}`);
+      // silently fail if no localStorage or JSON stringify fails
```
zaccolley, line 80: If a browser doesn't support localStorage we don't really need to know, the hook will instead fallback to not remembering state on refresh which is degraded experience but not worth creating logs for

web/src/components/pages/dashboard/stats/leaderboard-card.tsx
```diff
@@ -229,14 +211,6 @@ const Leaderboard = connect<PropsFromState>(
   { forwardRef: true }
 )(UnconnectedLeaderboard);
 
-const FilledCheckIcon = () => (
```
zaccolley, line 232: Unused code

web/src/components/pages/dashboard/challenge/leaderboard-card.tsx
```diff
@@ -41,7 +41,7 @@ export default function LeaderboardCard({
   const leaderboardRef = useRef(null);
   return (
     <StatsCard
-      key="leaderboard"
```
zaccolley, line 44: `key` is not supported as a prop, it's used interally in React for rendering things like lists. Renamed it to `id` for this reason.

web/src/components/pages/dashboard/stats/stats-card.tsx
```diff
@@ -26,11 +31,16 @@ export default function StatsCard({
...
+  const [locale, setLocale] = useLocalStorageState(DEFAULT_LOCALE_OPTION, id);
   const [selectedTab, setSelectedTab] = useState(Object.keys(tabs)[0]);
-  useEffect(() => setLocale(currentLocale ? currentLocale : ALL_LOCALES), [
-    currentLocale,
-  ]);
+  const isDefaultOptionSelected = locale === DEFAULT_LOCALE_OPTION;
+
+  // handle site wide locale change
+  useEffect(() => {
+    if (isDefaultOptionSelected && currentLocale) {
```
zaccolley, line 40: Without this check, locale changes overwrite our localstorage change as it would expect a page refresh to revert to the default option.

PR #3441, opened 2022-02-02
feat(script): check for unused localization

zaccolley:
Working on a script to let us check when localization is used in the codebase.

This is more complicated because some localization is injected in from the sentance collector, I assume because then Pontoon can pick it up in this repo instead of that repo.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -14,3 +14,6 @@ server/js/
 server/upload/
 web/dist/
 web/index.html
+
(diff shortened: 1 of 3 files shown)


MichaelKohler:
> This is more complicated because some localization is injected in from the sentance collector, I assume because then Pontoon can pick it up in this repo instead of that repo.

Yes, that is a correct assumption.

PR #3457, opened 2022-02-09
fix(stats): moving between languages stored filters

zaccolley:
I noticed when moving between the languages on the stats page we would incorrectly store the stats filters.

image

This PR fixes this issue.

### DIFF
--- a/web/src/components/pages/dashboard/stats/stats-card.tsx
+++ b/web/src/components/pages/dashboard/stats/stats-card.tsx
@@ -31,13 +31,16 @@ export default function StatsCard({
   scrollable?: boolean;
   currentLocale?: string;
 }) {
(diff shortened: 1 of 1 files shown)


web/src/components/pages/dashboard/stats/stats-card.tsx
```diff
@@ -31,13 +31,16 @@ export default function StatsCard({
   scrollable?: boolean;
   currentLocale?: string;
 }) {
-  const [locale, setLocale] = useLocalStorageState(DEFAULT_LOCALE_OPTION, id);
+  const [locale, setLocale] = useLocalStorageState(
+    DEFAULT_LOCALE_OPTION,
+    `${id}${currentLocale}`
```
zaccolley, line 36: We now store a value per stats card/tab before it didn't account for the language tab.

mozgzh (approved):

PR #3458, opened 2022-02-09
Implement new language variants

mozgzh:
## Frontend

JIRA ticket: https://mozilla-hub.atlassian.net/browse/OI-2044

+ Some fixing up of existing components
+ Adds ability to select a language variant is available.
+ Currently has a placeholder empty endpoint to be implemented in #3458
+ Add improved accessibility `ExpandableInformation` component to replace broken component (`profile-toggle`)
+ Add help text when a variant is shown in UI

## Backend

...

## Database

Add two tables for variants

+ `variants` holds the literally variant types and associated labeling
+ `user_client_variants` holds the relation between a user and the variant


server/src/lib/model/db/utils/getDiff.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Get the difference of two arrays
+ */
+export const getDifferenceInIds = (
+  requestedIds: number[],
+  savedIds: number[]
+): IdDifferences => {
+  const idsToBeAdded: number[] = [];
+  const idsToBeRemoved: number[] = [];
+
+  requestedIds?.map(id => {
```
zaccolley, line 16:
```suggestion
  requestedIds.forEach(id => {
```

We don't need `?` cause we will always have an array? And I think a forEach makes more sense here as it isnt being used to create a new array

server/src/lib/model/db.ts
```diff
@@ -894,6 +891,36 @@ export default class DB {
...
+      LEFT JOIN locales ON v.locale_id = locales.id
+       ${locale ? 'WHERE locale_id = ?' : ''}
+      `,
+      locale ? [await getLocaleId(locale)] : []
+    );
+
+    // return structured variants obj or empty {} if query is empty
+    const mappedVariants = variants
+      ? variants.reduce((acc: any, curr: any) => {
+          if (!acc[curr.token]) {
```
zaccolley, line 907: Should this be `acc[curr.lang]`?

It looks like this code creates:

```js
{
  'cy-north': [
    { id: 1, token: 'cy-north', name: 'North-Western Welsh' }
  ]
}
```

.eslintrc.js
```diff
@@ -34,6 +34,9 @@ module.exports = {
 
     // warn us if we're using @ts-ignore etc..
     '@typescript-eslint/ban-ts-comment': 'warn',
+
+    // Old code uses any a lot, just show a warn instead
+    '@typescript-eslint/no-explicit-any': 'warn',
```
zaccolley, line 39: I think this is already 'warn':

Screenshot 2022-02-15 at 12 29 50

Could you try reinstalling your node modules?

server/src/lib/model/user-client.ts
```diff
@@ -437,6 +530,7 @@ const UserClient = {
       updateDemographicsPromise,
       this.claimContributions(clientId, clientIds),
       languages && updateLanguages(clientId, languages),
```
zaccolley, line 532: Wondering here instead of adding another line for variants we rename `updateLanguages` to `updateAccents` and then add a function like:

```js
/**
 * Updates languages for a given user
 *
 * @param clientId clientId as string
 * @param languages array of UserLanguage
 * @returns void
 */
async function updateLanguages(clientId: string, languages: UserLanguage[]) {
  return Promise.all([
    updateAccents(clientId, languages),
    updateVariants(clientId, languages),
  ]);
}
```

server/src/lib/model/db/migrations/20220207181915-add-variants.ts
```diff
@@ -0,0 +1,71 @@
+const VARIANTS = [
```
zaccolley, line 1: Need to remove this test data

.eslintrc.js
mozgzh: It could be that my IDE was using a cached TS server. Does it hurt to have it explicitly in config?

zaccolley: Nope doesn't matter

zaccolley:
Pushed up an update for the label copy: https://github.com/common-voice/common-voice/pull/3458/commits/60acd5c54a3e4059234cc882b1a1808029b774ae

zaccolley:
> Could there be an option here for "My variant is not listed here" or "I can't find my variant in the list" ?

@ftyers Good idea, I'll post it on the JIRA ticket so EM can give some input there too

mozgzh:
Going to merge this in dev to get some more eyes for QA

PR #3460, opened 2022-02-10
Oi 2044 add variant to profile page 2

zaccolley:
Refactoring the `UserClient`:

+ `UserAccentLocale` to `UserLanguage`
+ `locales` to `languages` - I think this represents better what it stores.
+ Refactor saving of accents to make it clearer its for accents and not locales.

Separate commit for linting.


server/src/lib/model/user-client.ts
```diff
@@ -63,41 +64,43 @@ const generateAccentToken = (accent: string) => {
   return accent
     .trim()
     .toLowerCase()
-    .replace(/[\.,;\?\!'"]/g, '')
-    .replace(/[\s-_\(\)\[\]\<\>`]/g, '-')
+    .replace(/[.,;?!'"]/g, '')
+    .replace(/[\s-_()[\]<>`]/g, '-')
```
phirework, line 68: What was the thinking behind removing the escapes?

zaccolley: So it the linter picked it up as https://eslint.org/docs/rules/no-useless-escape

Looking into it more it looks like:
+ anything inside of a regex set `[]` doesn't need to be escaped: https://stackoverflow.com/questions/10593387/when-do-i-need-to-escape-characters-within-a-regex-character-set-within only exception is the `]`
+ `<` and `>` doesn't need to be escaped

Can change it back as it makes no difference to code, just style at this point

phirework: Huh, TIL! good to know, this is good to go. Thanks :)

common/locale.ts
```diff
@@ -6,9 +6,11 @@ export type Accent = {
   clientId?: string;
```
zaccolley, line 6: This file should be called 'user-language.ts' i think

PR #3461, opened 2022-02-10
WIP feat(variants): add UI for variant

zaccolley:
Not tested
Not hooked up to server correctly
For reference only


web/src/components/ui/icons.tsx
```diff
@@ -481,24 +468,6 @@ export const DiscourseIconCode = () => (
   </svg>
 );
 
-export const DownIcon = () => (
```
zaccolley, line 484: This was a duplicate of `ChevronDown`

web/src/components/expandable-information/expandable-information.tsx
```diff
@@ -0,0 +1,28 @@
...
+import './expandable-information.css';
+
+interface Props {
+  summaryLocalizedId: string;
+  children: React.ReactNode;
+}
+
+const ExpandableInformation = ({ summaryLocalizedId, children }: Props) => {
+  return (
+    <details className="expandable-information">
```
zaccolley, line 15: If we use the `<details>` component we don't need to use React to handle any state management, it works out of the box for accessibility too

PR #3476, opened 2022-02-21
Fix error pages

zaccolley:
### Type of Pull Request

- [ ] Bulk sentence upload 

*
*

- [ ] Other

*
*

*
*


web/src/components/layout/layout.tsx
```diff
@@ -248,10 +248,7 @@ class Layout extends React.PureComponent<LayoutProps, LayoutState> {
       featureStorageKey,
```
zaccolley, line 248: Because we no longer redirect to `/404` we can't use pathnames to determine classNames

PR #3477, opened 2022-02-22
feat(languages): new request pages

zaccolley:
WIP


server/src/lib/api.ts
```diff
@@ -520,4 +523,43 @@ export default class API {
...
+    }
+
+    const googleReCAPTCHA = new GoogleReCAPTCHA();
+    const isSuccessfulReCAPTCHA = await googleReCAPTCHA.verify(
+      reCAPTCHAClientResponse
+    );
+
+    if (!isSuccessfulReCAPTCHA) {
+      return response.status(500).send('Incorrect reCAPTCHA');
+    }
```
mozgzh, line 544: Since reCaptcha doesn't have to do with the actual response of the API (its essentially validation), it might make sense breaking it out into a reusable middleware and have it throw an error to the default error handler instead of handling the validation here?

server/src/lib/api.ts
```diff
@@ -520,4 +523,43 @@ export default class API {
...
+      response.json({
+        id: info?.messageId,
+        emailPreviewURL: info?.emailPreviewURL,
+        email,
+        languageInfo,
+        languageLocale,
+      });
+    } catch (e) {
+      console.error(e);
+      return response.status(500).send('Something went wrong emailing');
```
mozgzh, line 562: Ideally throw the error and let the default error handler in `server.ts` handle it

server/src/lib/api.ts
```diff
@@ -529,7 +532,23 @@ export default class API {
...
+    try {
+      const info = await this.email.sendLanguageRequestEmail({
+        email,
+        languageInfo,
+        languageLocale,
+      });
+
+      response.json({
+        id: info?.messageId,
+        emailPreviewURL: info?.emailPreviewURL,
```
zaccolley, line 555: Remove in property in production

server/src/lib/email.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    const ses = new AWS.SES({
+      apiVersion: '2010-12-01',
+      region: AWS_REGION,
+      ...AWS_SES_CONFIG,
+    });
+
+    return nodemailer.createTransport({
+      SES: { ses, aws: AWS },
+      sendingRate: 30, // message per second throttling
```
zaccolley, line 43: Explore this number

server/src/lib/api.ts
```diff
@@ -91,7 +95,12 @@ export default class API {
 
     router.get('/language/accents/:locale?', this.getAccents);
     router.get('/language/variants/:locale?', this.getVariants);
-    router.post('/language/request', this.sendLanguageRequest);
+    router.post(
+      '/language/request',
+      googleReCAPTCHAMiddleware,
```
mozgzh, line 100: nit: function name doesn't need to have "middleware" since it is a middleware (kinda like calling `const dataVariable = <blah>`)

zaccolley: Good spot thank you

PR #3478, opened 2022-02-23
fix(profile): don't allow empty accents values

zaccolley:
This PR stops users from entering empty accents on the profile page when adding languages.

Thanks @mozgzh

### DIFF
--- a/web/src/components/pages/profile/info/languages/input-language-accents/input-language-accents-input.test.tsx
+++ b/web/src/components/pages/profile/info/languages/input-language-accents/input-language-accents-input.test.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { axe, toHaveNoViolations } from 'jest-axe';
-import { RenderResult } from '@testing-library/react';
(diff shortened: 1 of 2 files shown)


mozgzh (approved):
awesome :rocket:

phirework:
Hey I have some concerns here - the reason empty accents were allowed was because: 

1. Prior to user-submitted accents being a feature, users who added languages to their profiles would often add language without an accent, and all of that legacy data needed to be migrated into the new system
2. On the basis of data minimization, users should be allowed to determine what level of granularity of data they want to provide. And I'm concerned that disallowing empty entries (i.e. just a language with no accent) would either disincentivize adding languages altogether, or create noise with lots of people just writing "n/a" or irrelevant info, especially given how many speakers of dominant accents don't think of themselves as having an accent at all. 

Is there a specific design/UX reason why you wanted to disable empty accents?

zaccolley:
@phirework so this PR stops someone from entering something like '    ' as a custom accent, you don't have to set any accents if you don't want

PR #3480, opened 2022-02-23
fix(profile): accents UI on RTL languages have incorrect spacing

zaccolley:
+ Fixes some spacing issues for the profile pages accent list on RTL languages
+ Move the 'Close' X to the right on LTR and to the right on RTL

**Before**
Screenshot of broken UI for RTL, the padding is broken

**After**
Screenshot of fixed UI, spacing looks correct


phirework (approved):
Looks good, thanks!

PR #3483, opened 2022-02-24
fix(speak/listen): better loading of data and better error handling

zaccolley:
+ Do not load sentences and clips needed for /speak or /listen pages on any page load. Now loads when needed
+ Add a spinner loading on /speak, /listen page when loading
+ Adds error messaging for when getting random sentences/clips error
+ Tidy up styling/markup for error messages
+ Detect erroring clips and remove them from clips list
+ Do not request clips again if we have a XMLHttpRequest error

### DIFF
--- a/web/src/components/pages/contribution/listen/listen.tsx
+++ b/web/src/components/pages/contribution/listen/listen.tsx
@@ -62,6 +62,7 @@ interface PropsFromState {
 }
 
 interface PropsFromDispatch {
+  loadClips: typeof Clips.actions.refillCache;
   removeClip: typeof Clips.actions.remove;
   vote: typeof Clips.actions.vote;
   addAchievement: typeof Notifications.actions.addAchievement;
@@ -107,6 +108,11 @@ class ListenPage extends React.Component<Props, State> {
     }
 
     return null;
+  }
+
+  componentDidMount(): void {
+    const { loadClips } = this.props;
+    loadClips();
   }
 
   componentWillUnmount() {
@@ -347,27 +353,30 @@ class ListenPage extends React.Component<Props, State> {
               </>
             }
             pills={clips.map(
-              ({ isValid }, i) => (props: ContributionPillProps) => {
-                const isVoted = isValid !== null;
-                const isActive = clipIndex === i;
-                return (
-                  <Pill
-                    className={isVoted ? (isValid ? 'valid' : 'invalid') : ''}
-                    onClick={null}
-                    status={isActive ? 'active' : isVoted ? 'done' : 'pending'}
-                    {...props}>
-                    {isActive ? (
-                      <VolumeIcon />
-                    ) : isVoted ? (
-                      isValid ? (
-                        <CheckIcon />
-                      ) : (
-                        <CrossIcon />
-                      )
-                    ) : null}
-                  </Pill>
-                );
-              }
+              ({ isValid }, i) =>
+                (props: ContributionPillProps) => {
+                  const isVoted = isValid !== null;
+                  const isActive = clipIndex === i;
+                  return (
+                    <Pill
+                      className={isVoted ? (isValid ? 'valid' : 'invalid') : ''}
+                      onClick={null}
+                      status={
+                        isActive ? 'active' : isVoted ? 'done' : 'pending'
+                      }
+                      {...props}>
+                      {isActive ? (
+                        <VolumeIcon />
+                      ) : isVoted ? (
+                        isValid ? (
+                          <CheckIcon />
+                        ) : (
+                          <CrossIcon />
+                        )
+                      ) : null}
+                    </Pill>
+                  );
+                }
             )}
             reportModalProps={{
               reasons: [
@@ -427,6 +436,7 @@ const mapStateToProps = (state: StateTree) => {
 };
 
 const mapDispatchToProps = {
+  loadClips: Clips.actions.refillCache,
   removeClip: Clips.actions.remove,
   vote: Clips.actions.vote,
   addAchievement: Notifications.actions.addAchievement,
--- a/web/src/components/pages/contribution/listen/listen.tsx
+++ b/web/src/components/pages/contribution/listen/listen.tsx
@@ -11,13 +11,11 @@ import URLS from '../../../../urls';
 import {
   CheckIcon,
   CrossIcon,
-  MicIcon,
   OldPlayIcon,
   ThumbsDownIcon,
   ThumbsUpIcon,
   VolumeIcon,
 } from '../../../ui/icons';
-import { LinkButton } from '../../../ui/ui';
 import ContributionPage, {
   ContributionPillProps,
   SET_COUNT,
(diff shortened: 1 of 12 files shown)


web/src/services/api.ts
```diff
@@ -105,7 +105,7 @@ export default class API {
     return this.fetch(`${this.getLocalePath()}/sentences?count=${count}`);
   }
 
-  fetchRandomClips(count: number = 1): Promise<Clip[]> {
+  async fetchRandomClips(count = 1): Promise<Clip[]> {
```
zaccolley, line 108: Adding `async` here is actually important because I'm not sure the errors were being caught otherwise

web/src/stores/clips.ts
```diff
@@ -53,93 +60,105 @@ export namespace Clips {
...
+    refillCache:
+      () =>
+      async (
+        dispatch: Dispatch<RefillCacheAction | LoadAction | LoadErrorAction>,
+        getState: () => StateTree
+      ) => {
+        const state = getState();
+
+        // don't load if no contributable locale
+        if (!contributableLocales.includes(state.locale)) {
```
zaccolley, line 78: We moved this logic from https://github.com/common-voice/common-voice/pull/3483/files#diff-c47fbf66be6c67df8a04429035911604d242969370b34d120f39cb668e569f7eL31

web/src/components/pages/contribution/listen/listen.tsx
```diff
@@ -282,26 +288,14 @@ class ListenPage extends React.Component<Props, State> {
...
-                    </LinkButton>
-                  </div>
-                </div>
-              )
+              <ListenErrorContent
+                isLoading={isLoading}
+                hasLoadingError={hasLoadingError}
+                isMissingClips={isMissingClips}
+                isDemoMode={this.demoMode}
+              />
```
phirework, line 298: Where is `isLoading` being passed in? It's not one of the props that the `<ListenPage>` component gets in `app.tsx`

web/src/stores/clips.ts
```diff
@@ -53,93 +60,105 @@ export namespace Clips {
...
+
+              return {
+                id: clip.id,
+                glob: clip.glob,
+                sentence,
+                audioSrc: clip.audioSrc,
+              };
+            }),
+          });
+          await Promise.all(clips.map(({ audioSrc }) => fetch(audioSrc)));
```
phirework, line 113: I think this is going to expose an issue that wasn't introduced by this PR but that will be made more obvious by it. The clips were originally loaded immediately on loading any page because `/api/v1/clips` doesn't actually return the clip files, just the public URLs, and the clips themselves aren't loaded until the URLs are returned to the store and they're fetched one by one. So clips were being loaded in immediately so that all of this downloading could happen in the background, so that they would be ready when users hit `/listen`. If clips weren't ready when users hit 'play', there would be all sorts of janky behaviour like the animation going wildly, unsupportedSource errors, etc. 

That was true before and is still true now, but I think moving the load to `/listen` (which makes logical sense) will expose that behaviour to more users, and the presence of a loader will make users think that the clips _have_ loaded when they haven't and be even more confused.  Based on local testing, it seems like the `refillCache` will return even if the fetch operations themselves error or time out. I don't think solving for that should block this from deploying, esp in terms of getting the site back to operations ASAP, but I wanted to flag it as something that will need a better solution soon. (Esp as AWS is also experiencing outages, which will mess with our clip loading.)

phirework (dismissed):
Looks good, thanks for the quick work! Left a couple of comments but nothing blocking, happy to discuss further on Slack.

web/src/components/pages/contribution/listen/listen.tsx
zaccolley: This comes from where we map the state to the props:

https://github.com/common-voice/common-voice/pull/3483/files/b909af38c3c131b89104b55d77bc37e6104e58eb#diff-b8b5bd64a65dc3163107a35c1a7bc61a03195781b640d8a8dccfc4fc2ec41160R404

web/src/stores/clips.ts
```diff
@@ -53,93 +115,83 @@ export namespace Clips {
...
-              id: clip.id,
-              glob: clip.glob,
-              sentence,
-              audioSrc: clip.audioSrc,
-            };
-          }),
-        });
-        await Promise.all(clips.map(({ audioSrc }) => fetch(audioSrc)));
-      } catch (err) {
-        if (err instanceof XMLHttpRequest) {
```
zaccolley, line 98: I think the idea with this is if one of clips was a bit dodge we would just request a whole new batch of clips but I think this could end up with us in a infinite loop if AWS is down?

web/src/stores/clips.ts
zaccolley: This is really useful thank you, I added two commits which I think could help. Could you review those for me?

zaccolley: This could make the page a bit slower as we have to wait for all of them to be loaded enough to be able to be played. So really a better solution would be to do this error checking for each clip, but I think this is better than what we currently have?

web/src/stores/clips.ts
```diff
@@ -1,16 +1,71 @@
...
+  });
+}
+
+async function checkClipsForErrors(clips: Clip[]) {
+  return await Promise.all(
+    clips.map(async clip => {
+      let hasError = false;
+
+      try {
+        // can play clip
```
zaccolley, line 30: Rather than doing a `fetch` and downloading the whole file, we can use the Audio API to detect as soon as they are ready to 'play', this should be a bit faster and also gives us the benefit of checking whether they're valid audio files too

web/src/stores/clips.ts
```diff
@@ -1,16 +1,71 @@
...
 import StateTree from './tree';
 import { User } from './user';
 import { Clip } from 'common';
 
+async function getCanPlayAudio(audioSrc: string) {
+  return new Promise(resolve => {
+    const audio = new Audio(audioSrc);
+
+    audio.addEventListener('error', () => {
+      resolve(false);
```
mozgzh, line 13: should this be `reject()`

zaccolley: It could be for sure, the function does return whether or not it can play audio so I think it erroring is an expected state

web/src/stores/clips.ts
phirework: No, `bucket.ts` has provisions that checks clips one by one, so that if one of the clips is a dodge it'll try to pick up a new one, up to a 1.5x multiplier of the default number of clips required: https://github.com/common-voice/common-voice/blob/main/server/src/lib/bucket.ts#L122-L124 (I'm not sure if that explanation makes sense, the code is probably clearer)

If AWS is down across the board and none of the clips are returning, that would be a different problem - there should probably be some max # of retries?

zaccolley: Do you know why we'd expect XMLHttpRequest errors? My hunch is if it errors we should tell the user and they can come back/refresh

phirework: I don't, sorry.

phirework (approved):

web/src/stores/clips.ts
phirework: This is great, thanks so much!

web/src/stores/clips.ts
phirework: I think that's a really good interim solution though, thanks for putting this together!

PR #3487, opened 2022-02-24
Improve getRandomClips performance

mozgzh:
This PR is to enable caching for a heavy query that's involved with finding clips to show to a user. By breaking up the query into multiple parts, utilizing caching where it makes sense and ensuring BI is in code (instead of SQL), the DB server can rest easier, ideally, dramatically reducing reads and total rows processed.


phirework (approved):
🚢 it!

PR #3497, opened 2022-03-01
fix(lodash): importing lodash incorrectly

zaccolley:
Was doing some brief Webpack digging. This does seem to be [the best way to import lodash](https://itnext.io/lodash-es-vs-individual-lodash-utilities-size-comparison-676f14b07568).

- Add `yarn build:audit` to launch the Webpack Bundle Analyzer
- Fix lodash dependancies importing lots more lodash than we needed

### Bundle size (Gzipped)

| Before | After | Percentage change |
|-|-|-|
| 283.34 KB | 259.97 KB | -8.24804% ✅ |

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "noImplicitAny": true,
     "sourceMap": true,
+    "allowSyntheticDefaultImports": true,
     "lib": ["es2017", "dom", "esnext.asynciterable"]
   }
 }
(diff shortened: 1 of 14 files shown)


tsconfig.base.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "noImplicitAny": true,
     "sourceMap": true,
+    "allowSyntheticDefaultImports": true,
```
zaccolley, line 5: We need this to be able to do `import pick from 'lodash.pick'`

https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports

web/src/components/welcome-modal/welcome-modal.tsx
```diff
@@ -2,8 +2,9 @@ import * as React from 'react';
 import { useEffect, useState } from 'react';
 import { Redirect } from 'react-router';
 import { Link } from 'react-router-dom';
-import { pick } from 'lodash';
```
zaccolley, line 5: Importing lodash functions like this was causing us to have a much bigger dependancy than we needed.

mozgzh (approved):

PR #3499, opened 2022-03-01
fix(images): huge images loaded

zaccolley:
While we shouldn't be loading demo content at all on the homepage this is a good fix regardless.


web/src/components/demo-pages/kiosk/create-account.tsx
```diff
@@ -36,14 +36,13 @@ const getCreateAccountComponents = (): PageContentType => {
...
-            <p />
-          </Localized>
+          <h2>
+            <Localized id="demo-account-card-header" />
+          </h2>
+          <p>
+            <Localized id="demo-account-card-body" />
+          </p>
           <div id="demo-account-sign-up">
-            <img src={require('./assets/qr-code-2.svg')} id="qr-code" />
```
zaccolley, line 46: QR code was too small to be read, deleted

mozgzh (approved):

PR #3500, opened 2022-03-01
Add compression package for express

mozgzh:
Currently, Express is responsible for serving static assets (include images, js, css, etc). These large files can make use of gzip compression to shrink transfer size and reduce the required amount of user data to load the site. 

The solution is to use a first-class Express compression package which automatically sniffs headers and compresses assets where allowed. 

I've also removed a superfluous header that wasn't being utilized.


zaccolley (approved):
Nice

PR #3513, opened 2022-03-10
fix(datasets-page): various changes (old)

### LINKED ISSUE #1877: Dataset page description: design is not l10n-friendly
https://github.com/mozilla/voice-web/blob/4ab882b4b15bb2033819d97cfbe5743fd1b81fd2/web/src/components/pages/datasets/dataset-info.css#L95

The CSS using first-line to create a line of heading text, which looks good in English with fixed width, but caused a weird line-breaking effect in some language:

English: "We are building"
We are building

Dutch: We bouwen een... (means "We are building one" -- not as good as English but okay.)
We bouwen een

Chinese: 我們正在打造一... (means "We are building o..." -- the last char is not even finished a word) 
我們正在打造一

Though localizers can try a different sentence to avoid this issue, it would be better to take this into consideration when doing design.

zaccolley:
+ Fix and simplify backgrounds
+ Add in an actual heading replacing the summary of page (Fixes #1877)
+ Various styling fixes

Screenshot 2022-03-10 at 16 37 26


zaccolley:
Replaced by https://github.com/common-voice/common-voice/pull/3515

PR #3515, opened 2022-03-14
Fix various problems on Datasets page

zaccolley:
+ Correctly load datasets information on datasets page instead of embedding all of them into the main JavaScript bundle
+ Refactor datasets page components
+ Allow `Spinner` to be used inline in website for loading states
+ Add in an actual heading replacing the summary of page (Fixes https://github.com/common-voice/common-voice/issues/1877)
+ Simplify and remove datasets/dashboard pages from the demo
+ Fix and simplify backgrounds, various styling fixes

### Changes in bundle size in KB Gzipped (with percentage change)

| | All | Main | Datasets Page |
|-|-|-|-|
| **Before** | 322.8 | 269.96 | 5.4
| **After** |  245.32 (-24%)   | 181.21 (-39%)  | 12.24 (+127%)

image


zaccolley:
I actually want to put some tests on these pages

mozgzh (approved):
Generally it looks good, but because so much has changed its a little hard to keep track of. I'd recommend merging this for a separate deploy such that it'd be easier to test on dev/stage/prod.

zaccolley:
I agree

PR #3546, opened 2022-03-15
Use GitHub's CodeQL code analysis

zaccolley:
Seems a good idea for security, curious what this picks up


mozgzh:
This is interesting. Currently, its catching default params used in local but not in other envs. I'm just worried the extra noise might hinder ease of deploys/focus

zaccolley:
Fair I can close for now

PR #3550, opened 2022-03-15
Webpack fixes

zaccolley:
We currently have our CSS and JavaScript breaking on deploys because we don't have hashes in the filenames. A user then caches `bundle.js` or `index.css`. This PR stops these issues if there's incompability with the server/javascript/styles.

+ Add hashes to CSS and JavaScript files
+ Split out node_modules into `vendor` file
+ Update Webpack dependancies

**Before**

All (322.8 KB)
bundle.js (269.96 KB)

**After**

All (321.71 KB)
main.be33d544594ad371ac94.js (137.27 KB)
vendors.49192381f5e41301b5fc.js (134.55 KB)


web/webpack.config.js
```diff
@@ -38,13 +41,6 @@ module.exports = () => {
       },
     }),
 
-    function () {
```
zaccolley, line 41: Replaced by the inbuilt `new webpack.ProgressPlugin()`

mozgzh (approved):
Nice, this should help out a lot

PR #3614, opened 2022-03-23
Add target sentences column and data

mozgzh:
This branch brings changes to how sentence criteria are defined for the platform.
Instead of having the static 5000 sentence criteria hardcoded, it will be added to the database and utilized from queries.
Work includes:
- Create new column in locales and add sentence criteria target.
- Modifying import-locales.js to make use of db
- Modifying /language_stats endpoint to return `sentence_target_count` per locale

`/language_stats` now looks like the following: 
```{
  "inProgress": [
    {
      "locale": "ace",
      "localizedPercentage": 16,
      "sentencesCount": {
        "current_count": 87,
        "target_sentence_count": 750
      }
    }...
```

### DIFF
--- a/server/src/lib/model/db/migrations/20220323130651-locale-target-sentence-criteria.ts
+++ b/server/src/lib/model/db/migrations/20220323130651-locale-target-sentence-criteria.ts
@@ -248,28 +248,29 @@ const locales: Locale[] = [
 
 export const up = async function (db: any): Promise<any> {
   // console.log(locales);
-
-  const targetColumnQuery = await db.runSql(
-    `SELECT target_sentence_count FROM locales`
+  const checkCol = await db.runSql(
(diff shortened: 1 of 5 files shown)


server/src/lib/model/db/migrations/20220323130651-locale-target-sentence-criteria.ts
```diff
@@ -0,0 +1,276 @@
...
+    sentenceTarget: 2000,
+  },
+  {
+    localeToken: 'yo',
+    sentenceTarget: 2000,
+  },
+];
+
+export const up = async function (db: any): Promise<any> {
+  // console.log(locales);
```
zaccolley, line 250: commented out code

zaccolley (approved):

PR #3634, opened 2022-03-25
Create Contributions Dashboard Page for User Clips

mozgzh:
This is a MVP that introduces basic user controls for clips they submitted. The idea is that users can see and action on their submitted clips individually. Currently, there are no ways a user can see their submitted clips, other than through a "takeout", which is when all user-related data is provided to download.

Currently, only basic pagination and visibility controls are built - that is, users cannot yet delete, or modify clips, just view them. Ideally, that would also be built out. If this ultimately becomes something thats desirable, more time could be invested to build it out with more thought.

image
When a user has no clips

image
When a user has more clips, pagination is displayed and clips are loaded dynamically (not all at once)


zaccolley (commented):
Just commenting to get this out of my 'to review' because I can't remove myself as a reviewer lol

jessicarose:
Hello and my most effusive apologies for the long wait you've had in having this PR looked at. I agree that giving users more insights into their contributions would be incredibly useful and I'm really excited by the thinking you've outlined here. I'm so sorry that the project has dramatically changed in architecture, tooling and scope in the intervening years so the work here would need extensive support from the team to be brought up to date and that we don't currently have the capacity. 

I've documented this work internally for conversations around how we look at contributors' viewing and working with their own sentence and speech contributions for planning and discussion. I'll close this for now with my continued thanks, to clear this from reviewers and your plate for now.

Issue #3636, opened 2022-03-26
Future of Import Locales and Sentence Collector export


MichaelKohler:
https://github.com/common-voice/common-voice/blob/main/package.json is missing the `mysql` dependency, leading to a broken Sentence Collector export: https://github.com/common-voice/sentence-collector/runs/5703458747?check_suite_focus=true . Even with that fixed, the Sentence Collector export would probably still fail as it does not have access to the MySQL credential secrets.

What is the plan for the future for this script? Is it still meant to be run during export? If so, you'd need to add the credentials with the right env vars to the Sentence Collector repo. If not, then we need to come up with another way to periodically run this script.

MichaelKohler:
`const { getConfig } = require('../server/js/config-helper');` is also rather unfortunate for that kind of setup. Would love to hear what the end goal is here :)

mozgzh:
Ah interesting, I didn't realize import-locales was being called from SC's actions and exported outside the repo - I assumed it was being called internally within CV before deploy, and would have access to certain env variables. As for why these changes are necessary, there's a need to have `CONTRIBUTABLE_MIN_SENTENCES` differ per language (instead of just 5000) and have the frontend also reflect said changes. See #3614 for more info regarding the merge for backend changes and #3477 for frontend changes. 

In the medium-term, ideally  `contributable.json` would be replaced with a db query that's able to dynamically return the list of contributable locales and generally, everything starts moving in the db instead of across static files when it makes sense.
 
Just brainstorming here, but I wonder if it might make sense to remove that section for the GH action (the one that runs import-locales) and have that run on app instantiation or scheduled job instead. That's assuming the rest of the export isn't dependent on changes from the import-locales script. I'd love to get more insight into the export and general feedback.

MichaelKohler:
@mozgzh thanks for the answer. Generally there is absolutely no other dependency from the Sentence Collector export side. It was simply added there as it didn't run anywhere else and at that point it made sense to run it after exporting. For now I've removed that step. This means that currently `contributeable.json` won't be updated anymore and there is no other mechanism in place to know that we'd need to do manually. So yes, it would be good if you could integrate this either in the deploy process or a cron on the CV side. I'll leave it up to you to decide what way forward you'd like to go :)

PR #3643, opened 2022-03-30
Run import-locales script on build

mozgzh:
Since `import-locales` now requires a database (and connection) to do its job, I've moved the script into the build of the app. This is more or less a test to see if this is feasible. Addresses #3636


zaccolley:
One thing I'll say is this can take a long time to complete

mozgzh:
> One thing I'll say is this can take a long time to complete

This is something that needs to be checked w/ this PR, essentially is it feasable to run everytime we deploy or should it be brought out into a separate job. The main issue w/ the job approach is that its likely a substantial rewrite, and it'd need to ensure the job is synced across each instance.

mozgzh:
Okay I did some more sleuthing and discovered some of the magic sauce that makes up CV, some of which directly relates to this script fix. 

Essentially, `import-script` is responsible for:
- Building the list of all languages that feature on the platform (the list on /languages page) by creating `all.json`
    - Queries Pontoon to get all translated languages along w/ translation stats which form the list (english injected manually)
- Building the "right to left" language file `rtl.json`
- Dynamically injecting the list of languages in the English `messages.ftl` file which is subsequently translated. 
- Building  the native name of each language file  `native-names.json`
    - List constructed by parsing every `messages.ftl` file in /web/locales/** directory, searches for locale code in its own `messages.ftl`
- Building `contributable.json` which lists all languages which users can interact with (record clips etc) on the platform

The challenges with adding the `target_sentence_count` column to the locales table and making use of it, is that the current way we're deciding if languages are in progress or not is very fluid and 100% dependent on Pontoon, not to mention occurs completely outside the platform (the script is never ran by the platform). When I created my work around to use the new column, I didn't realize how circular the language creation process is, or that the script is literally what builds critical data for CV.

I think going forward, the solution should be to break up the import as much as possible, and replacing static json files with cached queries.

Instead of creating `all.json`, create a fnc that on startup hits Pontoon for the language lists, compares to the existing db and modifies accordingly. Then replace `contributable.json` with a cached query with the same logic as the script. The big chunk of work is refactoring code to make use of a query/API call instead of reading a static file. 

Sorry for the wall of text, but it took a sometime to figure it all out and writing helps keep the thoughts in order! @zaccolley I'd love to hear any of your thoughts on the topic, or maybe other ways to reconcile with some of the quirkiness that is this script!

PR #3649, opened 2022-03-31
Fix random sentences only coming back with 1

zaccolley:
Our validator coerced the query variable into a number but this isn't reflected in the Typescript types.

A proper fix is to use AJV's Typescript features but for now this will do.


server/src/lib/api.ts
```diff
@@ -503,21 +506,4 @@ export default class API {
       await this.model.db.getVariants(client_id, params?.locale || null)
     );
   };
-
-  private getCountQueryParam = (request: Request) => {
-    const { count } = request.query;
-
-    if (typeof count !== 'string') {
```
zaccolley, line 510: count is now a number and fails here

PR #3654, opened 2022-04-05
Make condition more robust

mozgzh:
Moves around when the query is made and allows for new languages to be included (by defaulting to a fallback target).

### DIFF
--- a/scripts/import-locales.js
+++ b/scripts/import-locales.js
@@ -5,7 +5,6 @@ const { parse } = require('@fluent/syntax');
 const fetch = require('node-fetch');
 const { promisify } = require('util');
 const { getConfig } = require('../server/js/config-helper');
-console.log('test');
 const TRANSLATED_MIN_PROGRESS = 0.75;
 const CONTRIBUTABLE_MIN_SENTENCES = 5000;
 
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -49,15 +50,14 @@
...
+    "start:prod": "yarn build:common && yarn import-locales && node server/js/main.js",
     "start:maint": "cd maintenance && yarn dev",
     "test": "yarn build:common && concurrently --kill-others-on-fail \"cd web && yarn test\" \"cd server && SERVER_CONFIG_PATH='../config.json' yarn test\""
   },
   "husky": {
     "hooks": {
       "pre-commit": "pretty-quick --staged"
     }
-  },
-  "packageManager": "yarn@1.22.18"
```
zaccolley, line 62: Question: Why are we removing this? Seems useful as Yarn 1 and Yarn 2/3 are quite different

package.json
```diff
@@ -26,6 +26,7 @@
     "eslint-plugin-react": "7.29.4",
     "eslint-plugin-react-hooks": "4.3.0",
     "husky": "7.0.4",
+    "mysql2": "2.3.3",
```
zaccolley, line 29: This is nice cause we need to migrate to this elsewhere in the codebase!

zaccolley (approved):
Looks good to me, nothing blocking. As a note, this file imo needs to broken up into more functions cause it's quite hard to read.

mozgzh:
Totally agree. Ideally this file doesn't exist - all its functionality should be contained within the platform, controlled through scheduled jobs. CV has multiple sources of truth (database, many many static files) and also gets around using jobs through obscure Github Actions which is not super great.

package.json
mozgzh: Whoops, nice catch - didn't think I committed that!

scripts/import-locales.js
```diff
@@ -102,41 +101,72 @@ async function importPontoonLocales() {
...
+                        'utf-8'
+                      )
+                      .split('\n').length
+                  : count,
+              0
+            );
+          const currentLang = locales[name];
+
+          // get target sentence count from db if exists, else use default
+          const currentTargetSentenceCount =
```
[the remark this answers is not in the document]
mozgzh: Not supported in Node v12 (This is a js file that's run outside tsc) 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility

PR #3656, opened 2022-04-06
Simplify Development Processes

mozgzh:
Main changes:
- Expose DB port locally (makes connecting to IDE/Work Bench easy)
- Stop building images on docker hub on merges to main/stage/production since we never ship them (we ship tagged images)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,6 +2,8 @@ name: Build and Push Docker container
 
 on:
   push:
+    branches:
+      - main
     tags:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -2,11 +2,6 @@ name: Build and Push Docker container
 
 on:
   push:
-    branches:
-      - main
```
zaccolley, line 6: Main is used for `dev.commonvoice.allizom.org/`

PR #3658, opened 2022-04-07
Load contributable language from database instead of static files

mozgzh:
Because all files in `locales/*.json` are being loaded in via `require()`, _Node_ is caching their values (not Redis or anything). That means, instead of the newly minted versions of the file (created through `import-locales`) being used, an old un-tracked version (not obviously) is being used. 

##  Changes

+ Remove all `locales/*.json` file dependencies on the backend, which are replaced with cached queries.
+ Replaces some of the startup logic for the backend, fetching the languages directly from Pontoon instead of relying on the import script. 
+ Use Webpack to copy existing files generated by `import-locales.js` into the `dist/` directory
+ We can use the new `getAllLanguages` endpoint instead of the `all.json` and `contributable.json` files. These are now not being generated by `import-locales.js`
+ Because we do not have the languages hardcoded into the client anymore we have to load them into our state Redux
+ Every instance of using the hardcoded imports now uses Redux state and has helper hooks to get that data
+ This also changes how our error pages and routing work - the top level providers have been refactored and seperated into different files
+ Because we can have errors happen when we don't know the localisation a fallback simple English error page is introduced with `<ErrorBoundary>`
+ The Discourse languages have been hardcoded into the client
+ Updated existing tests to handle the data coming from Redux instead of imported files


common/language.ts
```diff
@@ -6,6 +6,14 @@ export type Accent = {
   clientId?: string;
 };
 
+// a single accent object
+export type Language = {
+  id: number;
+  name: string;
+  target_sentence_count: number;
```
zaccolley, line 13: can we make these:

I think these should be camelCase, could we make them:

```js
targetSentenceCount: number;
totalSentenceCount?: number;
```

zaccolley: actually could we use the existing 'SentenceCount' here?

web/src/components/app.tsx
```diff
@@ -238,173 +187,90 @@ let LocalizedPage: any = class extends React.Component<
...
-                }
-              />
-            </Switch>
-          </Router>
-        </Provider>
-      </Suspense>
-    );
-  }
-}
+// TODO what happens if stuff above error boundary errors?
```
zaccolley, line 255: Need to solve this...

web/src/components/layout/layout.tsx
```diff
@@ -211,7 +213,9 @@ class Layout extends React.PureComponent<LayoutProps, LayoutState> {
           </header>
         </div>
         <NonProductionBanner />
-        <Content location={location} />
+        <main id="content">
```
zaccolley, line 216: We don't need to make these changes anymore, can revert this

web/src/stores/sentences.ts
```diff
@@ -168,8 +168,15 @@ export namespace Sentences {
     }
   }
 
+  const localeSentences = ({ locale, sentences }: StateTree) => {
+    if (!sentences[locale]) {
+      return DEFAULT_LOCALE_STATE;
```
zaccolley, line 173: This feels wrong, I think we should 'load' a default locale state into our state tree rather than do this.

web/src/components/error-boundary/catch-errors-with-sentry.ts
```diff
@@ -0,0 +1,18 @@
+import * as Sentry from '@sentry/react';
+
+import { isProduction, isStaging } from '../../utility';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export default function catchErrorsWithSentry(error: Error, errorInfo: any) {
+  // don't log errors in development
+  if (!isProduction() && !isStaging()) {
```
mozgzh, line 8: Fantastic!

mozgzh:
Since this PR is now really large, once you're in a comfortable spot to get some eyes on it, can you deploy it to `sandbox`?

zaccolley:
Need to handle if the network request errors, and through an error on frontend

zaccolley:
We need a timeout fallback if the languages don't load, on sandbox we have an issue where languages_all is not returning anything and we just see a spinner forever.

PR #3661, opened 2022-04-11
Revert "adding sw variant sentences "

zaccolley:
Reverts common-voice/common-voice#3640

@Witty-Kitty this is a temporary measure to ensure we can manually tag these with variant metadata if possible as discussed on Slack


zaccolley:
We have already released this sentences, we cannot revert this now

PR #3670, opened 2022-04-22
Revert invalidating cache

mozgzh:
Reverts the caching logic, such that caches aren't rebuilt on deploy. Tying cache to releases forces a complete cache rebuild and has a large performance penalty on the platform. If there are multiple releases within cache expiry, the redis instance is quickly overburdened by a massive increase in memory required (this is because we aren't flushing old cache, just build new cache on top of the old). 

A (less than ideal) solution is to just rename the cache that is being specifically worked on, thus forcing just that cache to be rebuilt and not causing performance issues.

### DIFF
--- a/server/src/lib/lazy-cache.ts
+++ b/server/src/lib/lazy-cache.ts
@@ -1,24 +1,17 @@
 import { redis, redlock, useRedis } from './redis';
 
-import { getConfig } from '../config-helper';
-
 type Fn<T, S> = (...args: S[]) => Promise<T>;
 
 function isExpired(at: number, timeMs: number) {
   return Date.now() - at > timeMs;
 }
 
-function memoryCache<T, S>(
-  cacheKey: string,
-  cachedFunction: Fn<T, S>,
-  timeMs: number
-): Fn<T, S> {
+function memoryCache<T, S>(cachedFunction: Fn<T, S>, timeMs: number): Fn<T, S> {
   const caches: {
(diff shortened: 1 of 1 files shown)


server/src/lib/lazy-cache.ts
```diff
@@ -51,8 +44,7 @@ function redisCache<T, S>(
   timeMs: number
 ): Fn<T, S> {
   return async (...args) => {
-    const { RELEASE_VERSION } = getConfig();
-    const key = cacheKey + JSON.stringify(args) + (RELEASE_VERSION || '');
+    const key = JSON.stringify(args);
```
zaccolley, line 47: Confusingly the key for in memory cache was different to the Redis one before we made them the same, but we need this to be:

```js
const key = cacheKey + JSON.stringify(args);
```

zaccolley (changes requested):
Needs a change for it to be an actual revert

zaccolley:
Makes sense, I guess if we had flushed Redis on deploy any other way this would have happened too.

mozgzh:
I wonder if we need the "memory only" cache functionality anymore - the platform has direct dependencies on Redis (e.g. session management, and managing jobs). It would likely simplify this hard-to-read part of the code. What do you think @zaccolley?

zaccolley:
@mozgzh yeah I agree, I think we should remove the memory caching

zaccolley (approved):

PR #3673, opened 2022-04-26
Replace reCAPTCHA with rate limiting

zaccolley:
[JIRA Ticket](https://mozilla-hub.atlassian.net/browse/OI-2297)

+ Remove reCAPTCHA from `/languages/request` page
+ Adds a rate limiter for Express routes called `rateLimiterMiddleware`
+ Use `rateLimiterMiddleware` on the language request endpoint
+ Remove `console.error`s from tests

## After release

+ Remove `CV_GOOGLE_RECAPTCHA_SITE_KEY` and `CV_GOOGLE_RECAPTCHA_SERVER_KEY` from environments to avoid confusion.
+ Delete reCAPTCHA credentials from 1Password and from their admin pages


server/src/config-helper.ts
```diff
@@ -3,6 +3,10 @@ import { S3, SSM } from 'aws-sdk';
 import { SESClientConfig } from '@aws-sdk/client-ses';
 import { config } from 'dotenv';
 
+import Logger from './lib/logger';
```
mozgzh, line 6: Since we're likely to make use of Logger a lot, and it all generally relies on the same logic, I think we can make Logger a singleton. This would mean there'd only be on instance of Logger regardless of how many times its referenced (lowering memory usage). And to use it, call a static method (e.g. `Logger.log(name:string, log:Log)`). Or we could create new classes that extend the Log class (e.g. `AuthRouterLogger extends Logger`) centrally, in a new `/loggers` directory if we anticipate using the same log instance across files. That would be a bit of memory overhead as well, but allows Loggers to be shared. 

If we go one of the aforementioned routes, and it's a lot of work, I can create a ticket and make the suggested changes when theres time and keep this PR the same so we can ship it sooner

mozgzh:
Looks great. Love the addition of rate limiting! I'm excited to implement on ideally whole API going forward :rocket: One thing to note is if we are currently setup to handle IP forwarding from Nginx to Express (https://expressjs.com/en/guide/behind-proxies.html) in our current setup. And food for thought stuff, I have no preference, for local development, would we want to get around the limiter, or have a flag that can turn it off (on by default)?

PR #3682, opened 2022-04-29
Upgrade to Node 16

mozgzh:
Upgrade Node version from 12 to 16


jessicarose:
I'll go ahead and close this, as we got there in the end and the current node version is 18. Thank you so much for your patience!

Issue #3728, opened 2022-06-07
[Corrupted dataset] Very low variability in Catalan


Huguet57:
The Catalan sentence dataset is corrupted by a big dataset of the same words over and over again: https://github.com/common-voice/common-voice/blob/main/server/data/ca/frases_agenda.txt. Aproximately 50% of the sentences to record are with repeated words and can bring a big bias into the algorithms trained.

What can be done in these situations? Can this dataset be erased? Or are now the datasets sealed once it's in production?

Huguet57:
@c-armentano es podrien trobar datasets més variats per substituir aquest?

Huguet57:
From the paper: "Collecting text for reading aloud is the second step of adding a new language to Common Voice. For languages with more than 500,000 Wikipedia articles, text sentences are extracted from Wikipedia using community provided rule-sets per language."

@c-armentano did you use the [CV's Sentence Extractor](https://github.com/common-voice/cv-sentence-extractor) to extract the sentences of the wikipedia datasets?

c-armentano:
We are working to find new sentences, but it's difficult. Yesterday a [new PR](https://github.com/common-voice/common-voice/pull/3719) with +50k sentences was approved.
About wikipedia, it's [already done](https://github.com/common-voice/common-voice/blob/main/server/data/ca/wiki.ca.txt).
Regarding this sentences, it would be great to remove them without removing the ones who have been already recorded. Is it possible?

Huguet57:
50.000 frases aniran perfecte! Sobre no perdre totes les que s'han gravat de la base de dades corrupta, si fem ara que no surtin més, tindran menys veus que la resta de datasets així que estarien doble esbiaixats. Jo votaria per perdre les frases i les gravacions d'aquest dataset.

50k new sentences will be of great help. However, on not losing all the recorded data of the corrupted dataset, if we remove them from showing up now, in the long run they will have less recordings than the ones from other datasets (therefore, being doubly-biased).

I would vote to lose all setences and recordings of this dataset. As seen [here](https://github.com/common-voice/common-voice/blob/main/docs/SENTENCES.md#flagging-problematic-sentences), modifying the text files should be enough. I will now open a pull request to erase the file.

jmontane:
I don't agree with "corrupted" term, may be "biased" is a better one.

[...]

PR #3729, opened 2022-06-09
[Corrupted dataset] Delete biased dataset in Catalan

Huguet57:
Recording a sentence in Catalan at this moment, you get 80% sentences from this dataset and they are all highly repeated that can create bias to the algorithms. As mentioned in issue #3728.

For instance, some subsentences' counts:
- "vint-i-tres de maig" in 16171 sentences of a total of 160665 (10%)
- "Per Sant Jordi" in 13831 sentences of a total of 160665 (8,6%)
- "El deu d'agost" in 13051 sentences of a total of 160665 (8,1%)
- "El vint-i-nou de febrer" in 14611 sentences of a total of 160665 (9%)
- ...


c-armentano:
Hi!
I don't think removing the recorded clips would be a good solution. In my opinion, the best option would be removing the unrecorded sentences while keeping the recorded ones, as @jmontane stands [here](https://github.com/common-voice/common-voice/issues/3728). Is there a way to do so?

Huguet57:
Hi @Heyhillary, what's the status on the discussions?

mozgzh:
As mentioned in #3728 the CV isn't keen to remove recorded clips, which the exact details in the linked thread. Thanks for creating the issue and PR and sparking a good discussion around the topic! :rocket:

Issue #3775, opened 2022-08-10
[BUG] migration failed to run during server deployment


alenpaulvarghese:
**Describe the bug**
The migration seems to be failing during the deployment of the server. I checked twice already on multiple days and still face this issue.

**To Reproduce**
1. Clone the repository
2. rename the config 
3. docker-compose up

**Expected behavior**
The server should have started without any error.

Os - Gnu/Linux (Ubuntu) x86_64
Docker version - 20.10.12
Docker compose version - 20.10.12

**Build Log**

```
3:30:50 AM - Starting compilation in watch mode...
web      | [CO] 
web      | [FE] Webpack: Rebuilding...
web      | [CO] 
web      | [CO] 3:30:53 AM - Found 0 errors. Watching for file changes.
web      | [BE] 
web      | [BE] 3:31:04 AM - Found 0 errors. Watching for file changes.
web      | [BE] $ cd .. && node server/js/main.js
web      | [BE] Could not load config.json, using defaults (error message: ENOENT: no such file or directory, open './config.json')
web      | [BE] No Auth0 configuration found
web      | [BE] APPLICATION -- starting
web      | [BE] Cache is redis
db       | 2022-08-10T03:31:06.187592Z 2 [Warning] Following users were specified in CREATE USER IF NOT EXISTS but they already exist. Corresponding entry in binary log used default authentication plugin 'mysql_native_password' to rewrite authentication information(if any) for them: 'voicecommons'@'%'
db       | 
web      | [BE] APPLICATION -- performing Maintenance
web      | [BE] Running migrations
web      | [BE] APPLICATION -- listening at http://localhost:9000
web      | [BE] unhandled promise rejection Error: ER_BAD_FIELD_ERROR: Unknown column 'undefined' in 'field list'
web      | [BE]     at Query.Sequence._packetToError (/code/node_modules/mysql/lib/protocol/sequences/Sequence.js:47:14)
web      | [BE]     at Query.ErrorPacket (/code/node_modules/mysql/lib/protocol/sequences/Query.js:79:18)
web      | [BE]     at Protocol._parsePacket (/code/node_modules/mysql/lib/protocol/Protocol.js:291:23)
web      | [BE]     at Parser._parsePacket (/code/node_modules/mysql/lib/protocol/Parser.js:433:10)
web      | [BE]     at Parser.write (/code/node_modules/mysql/lib/protocol/Parser.js:43:10)
[...]

PR #3795, opened 2022-09-12
Fix when native name is sometimes used instead of ISO code

mozgzh:
On startup, in some circumstances, the native name is saved (when inserting new languages) to the wrong column instead of the ISO code.

Full fix for #3775

Fixed: 
- native names finding their way into the names column in locales
- removing migration that was used to add data to the db that depended on pre-existing data
- moved adding variants to the start up flow


moz-rotimib (approved):

server/src/lib/model/db/import-locales.ts
```diff
@@ -192,12 +292,46 @@ export async function importLocales() {
...
     );
-    console.log('Saving accent data to database');
+    console.log('Saving accents to database');
 
     // Make sure each language has at minimum an "unspecified" accent
     await db.query(`
     INSERT IGNORE INTO accents (locale_id, accent_name, accent_token, user_submitted)
     SELECT id, "", "unspecified", 0 from locales`);
+
+    console.log('Saveing variants to database');
```
moz-rotimib, line 302:
```suggestion
    console.log('Saving variants to database');
```

PR #3800, opened 2022-09-15
Remove localization in spinner after package upgrade throws error

mozgzh:
Errors are no longer ignored when Localization component called outside `LanguagesProvider` component. As a result, since the `Spinner` is called early in the app (on app load) and before `LanguagesProvider` it throws an error.

Quick fix is to remove localization on `Spinner` and just hard-code "Loading" for screen readers.


moz-rotimib (approved):
Works locally for me. I'll create ticket for this and link this PR

PR #3805, opened 2022-09-20
OI-2420 language list on profile not showing name

moz-rotimib:
Fix languages dropdown not showing language name but code.

### DIFF
--- a/web/src/stores/languages.ts
+++ b/web/src/stores/languages.ts
@@ -14,6 +14,7 @@ export interface State {
   nativeNames?: NativeNames;
   rtlLocales?: Locales;
   translatedLocales?: Locales;
+  contributableNativeNames?: NativeNames;
 }
 
 enum ActionType {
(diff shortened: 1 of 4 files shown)


web/src/stores/languages.ts
```diff
@@ -22,11 +23,14 @@ enum ActionType {
...
-  translatedLocales: Locales;
+  payload: {
+    allLocales: Locales;
+    contributableLocales: Locales;
+    nativeNames: NativeNames;
+    rtlLocales: Locales;
+    translatedLocales: Locales;
+    contributableNativeNames: NativeNames;
+  };
 }
```
moz-rotimib, line 34: Fixing this because Redux actions should follow the convention:

```
type ACTION = {
  action: ACTION_TYPE;
  payload: ACTION_PAYLOAD;
}
```

moz-rotimib: @mozgzh I've tested this locally and it looks good. Also, the unit test for the input language component also passes

mozgzh: awesome :rocket:

mozgzh (approved):
:sun_with_face:

PR #3771, opened 2022-08-05
fix stats graph's tooltip layout of the metrics data

fycdev:
### Type of Pull Request

- [ ] Bulk sentence upload 

- [x] Related to a listed issue 

* Fixes #3692 

- [ ] Other

* Adjust CSS of the tooltip in the Stats graph so that the metrics are laid out vertically and stays within the border.

PR #3817, opened 2022-09-27
Fix horizontal overflow in stats tooltip

mrgnr:
## Pull Request Form

The numbers corresponding to "Hours Recorded" and "Hours Validated" were positioned horizontally within the tooltip, causing the text to overflow outside of the tooltip bubble. We now use the default `display` styling for the containing `<div>`, which lays the metrics out vertically to avoid overflow.

### Type of Pull Request

- [ ] Bulk sentence upload 

* https://github.com/common-voice/common-voice/issues/3816
* https://github.com/common-voice/common-voice/issues/3692

- [ ] Other

### Screenshots

common-voice-fix-tooltip-overflow

### DIFF
--- a/web/src/components/pages/home/stats.css
+++ b/web/src/components/pages/home/stats.css
@@ -149,9 +149,6 @@
     }
 
     & .metrics {
-        display: flex;
-        flex-direction: row;
-
         & > * {
             margin-inline-end: 30px;
 


mozgzh:
Thanks for the PR! There's another PR #3771 that's related to this issue. I think the easiest solution is to ensure the tooltip container (with the white background) isn't restricted by width. If you want to give it a try, removing line 273 in index.css would fix the issue by allowing the parent div to wrap around the datapoints/date instead of forcing them to break. https://github.com/common-voice/common-voice/blob/441e644d0368dfd505992b17bc364276408d0dd7/web/src/components/index.css#L273

PR #3825, opened 2022-10-07
OI-2489: Add Google Tag manager to CV pages

moz-rotimib:
Adds Google Tag manager to CV pages

### DIFF
--- a/web/index_template.html
+++ b/web/index_template.html
@@ -75,6 +75,18 @@
         ga('send', 'pageview');
       }
     </script>
+    <script>
+      (function (w, d, s, l, i) {
+        w[l] = w[l] || [];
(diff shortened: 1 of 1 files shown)


web/src/main.ts
```diff
@@ -1,10 +1,17 @@
 import 'focus-visible';
 import * as React from 'react';
 import { render } from 'react-dom';
+import TagManager from 'react-gtm-module';
 import './components/index.css';
 
 declare var require: any;
 
+const tagManagerArgs = {
+  gtmId: process.env.GTMID,
```
moz-rotimib, line 10: I'm curious how we pass env variables on the front end. I've not seen examples of where we use env variables

PR #3826, opened 2022-10-12
OI 2486 Implement partnerships page

moz-rotimib:
### Implement partnerships page.

Adds new route `/partner`


mozgzh:
The PNG (web/src/components/pages/partner/images/waves.png) needs to be compressed (currently 1.8MB) or it'll bloat the size of the app when clients load the site

web/src/components/pages/partner/partner-option-section.css
```diff
@@ -0,0 +1,35 @@
...
+
+.partner-options-container img {
+    align-self: end;
+}
+
+.partner-options-text-container {
+    flex: 0.8;
+}
+
+.partner-options-text-container h3 {
```
mozgzh, line 18: Are there any other font classes (ideally generic) that look have similar css? I'd be good to ensure all our fonts use the same or similar css or make it so this style can be used more generally

web/src/components/pages/partner/partner-option-section.css
```diff
@@ -0,0 +1,35 @@
...
+    font-weight: 400;
+    text-align: left;
+    margin-bottom: 16px;
+}
+
+.partner-options-text-container p {
+    text-align: left;
+    font-weight: 400;
+    @media (--lg-up) {
+        width: 650px;
```
mozgzh, line 29: Just checking that we want to add width the the `p` tag, and not the container?

web/src/components/pages/partner/partner-options.tsx
```diff
@@ -0,0 +1,91 @@
...
+  },
+  {
+    id: 'foundations',
+    elems: {
+      programmaticWork: (
+        <a
+          href="https://foundation.mozilla.org/en/common-voice/in-country-programmes"
+          target="_blank"
+          rel="noreferrer">
+          programmatic work
```
mozgzh, line 34: Will this be translated?

web/src/components/pages/partner/partner.css
```diff
@@ -0,0 +1,168 @@
...
+            }
+        }
+    }
+
+    p {
+        font-weight: 300;
+
+        @media (--xs-down) {
+            font-size: 18px;
+            width: 355px;
```
mozgzh, line 52: Not sure we want to set exact widths considering the size of screen is variable and can be relative to the parent element

web/src/components/pages/partner/partner-options.tsx
moz-rotimib: Yeah, I think it will. Thanks for pointing it out

web/src/components/pages/partner/partner-option-section.css
moz-rotimib: I'm not sure. I've looked at various pages in the application and I can't see any text that looks similar to this

mozgzh (approved):

PR #3833, opened 2022-10-23
Replace incorrect quotes

galcorlo:
### Type of Pull Request

- [ ] Bulk sentence upload 

*
*

- [ ] Other

*
*

*
*


ftyers:
Hi @galcorlo, have you checked to see which of these sentences already have recordings, or do none of them have recordings?

galcorlo:
Hi @ftyers, I downloaded the dataset and I could verify that at least some sentences are there. Having theses sentences with incorrect quotes is not an bloker because the person who is recording his own voice is probably able to fix it, but it's true he has to pay more attention before start to record and probably he has to read the sentence two times, as it happened to me.
Can I do something else for helping you to modify these sentences?

moz-bozden:
I'm afraid we are too late for this.
CV does not get sentences from these `.txt` files anymore. These are sometimes/only used by developers for some testing. Currently the new Write page interfaces are used for sentence collection.

@ftyers, I think we should close this PR.

PR #3842, opened 2022-10-26
OI-2427 Stats API work

mozgzh:
Add statistics endpoint that provides a 12 month overview of the platform.

### DIFF
--- a/docs/API/statistics.md
+++ b/docs/API/statistics.md
@@ -1,26 +1,87 @@
 # Statistics
 
+All statistics provide the increase in metrics over a 12 month period in monthly intervals.
+
+## How statistics are calculated
+
+- If a month has no increase, it is skipped.
+- The date is provided as a key for that month (including a day value).
+- If a date has a day value that is **not** the final day of the month (e.g. 2022-12-15) that was the last time a value was added the interval still includes that whole month
+- _unless_ the month in question is the current month (i.e. if current date = 2022-12-16, the stats for that month are up until current day).
+
+## Endpoints
+
 Statistics endpoints are accessed at `/api/v1/statistics/<stat name>`
 
-## Development
+### Downloads
 
-When running this project locally we don't want to use AWS. We're using the test service [Ethereal](https://ethereal.email/). This means no actual emails are sent but we get a preview instead.
+All downloads of any dataset:
 
-You should see a `📧 Email preview URL: http://example.com` log whenever a email is sent.
+`/api/v1/statistics/downloads`
 
-> ⚠️ When sending emails Ethereal, emails are essentially public on their site. Consider this when testing.
+### Clips
 
-## Production
+All clip contributions (not unique to user or sentence):
+`/api/v1/statistics/clips`
 
-In production (or testing production) we need to have the `CV_AWS_SES_CONFIG` enviornment variable set to something like:
+#### Options
+
+Filter by only the rejected clips by using query parameter:
+
+`?filter=rejected`
+
+### Speaker
+
+All unique speaker contributors:
+
+`/api/v1/statistics/speakers`
+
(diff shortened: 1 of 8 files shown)


server/src/lib/model/statistics.ts
```diff
@@ -0,0 +1,130 @@
...
+export const queryStatistics = async (
+  tableName: TableNames,
+  options?: QueryOptions
+) => {
+  const isDistinict = options?.isDistinict ?? false;
+  let total_count, monthly_count;
+
+  // two basic query paths (queries with distinct have to use group by)
+  if (isDistinict) {
+    // ({ total_count } = (await getTotal(tableName))[0]);
```
moz-rotimib, line 42: If this code is commented out can it be removed?

docs/API/statistics.md
```diff
@@ -0,0 +1,89 @@
...
+All statistics provide the increase in metrics over a 12 month period in monthly intervals.
+
+## How statistics are calculated
+
+- If a month has no increase, it is skipped.
+- The date is provided as a key for that month (including a day value).
+- If a date has a day value that is **not** the final day of the month (e.g. 2022-12-15) that was the last time a value was added the interval still includes that whole month
+- _unless_ the month in question is the current month (i.e. if current date = 2022-12-16, the stats for that month are up until current day).
+
+## Endpoints
```
moz-rotimib, line 12: I think it would also be helpful to prefix the endpoints with the appropriate HTTP verb. I'm guessing some of these endpoints expect the GET verb but I'm not sure that's the case for all of them

mozgzh: Good catch, I've added HTTP values to the specific endpoints and example response

server/src/lib/model/statistics.ts
```diff
@@ -0,0 +1,239 @@
...
+    FROM
+      ${tableName} d
+    ${conditional ? 'WHERE ' + conditional : ''}
+  `);
+  return rows;
+};
+
+export const getStatistics = lazyCache(
+  'get-stats',
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
moz-rotimib, line 218: It seems the `_filter` parameter is not used anywhere in this function. So is it ok to remove it and the eslint comment?

server/src/lib/model/statistics.ts
```diff
@@ -0,0 +1,239 @@
...
+  async (tableName: TableNames, options?: QueryOptions, _filter?: string) => {
```
moz-rotimib, line 219: It seems the _filter parameter is not used anywhere in this function. So is it ok to remove it and the eslint comment?

server/src/lib/model/statistics.ts
```diff
@@ -0,0 +1,239 @@
...
+  const monthlyRunningTotals = monthlyIncrease.reduce((obj: any, row) => {
+    const diff = yearlySum - currentSum;
+    currentSum += row.total_count;
+    obj[row.date] = diff;
+    return obj;
+  }, {});
+
+  return {
+    yearly_sum: yearlySum,
+    total_count: totalCount,
```
moz-rotimib, line 87: **Question/Thought**: It seems `totalCount` is not really used anywhere else in this function besides being passed in as a parameter and returned from the function. So I was wondering if it was better to return it from the `getStatistics` function (after line 219). If there is another reason for it to be passed into this function I'm fine with leaving it as it is now.

mozgzh: I felt it made more sense to have one function responsible for building the response (it builds the response object and changes the key name). It'd theoretically make testing simpler (there is no testing currently however) because it removes a side effect of building in two places

moz-rotimib (approved):

PR #3850, opened 2022-11-02
OI-2492

moz-rotimib:
## OI-2492 Metadata expansion on Speak UI individual clips

- Adds a check box for accepting terms and privacy before submitting. Previously, this was done via a modal - the modal is still visible when the user tries to abort before uploading 5 clips
- Remove the share button from the recording pill

https://user-images.githubusercontent.com/113194143/199770020-ac38bc92-9a42-4ebe-8e64-da5f8ab6e7c1.mov

### DIFF
--- a/web/locales/en/messages.ftl
+++ b/web/locales/en/messages.ftl
@@ -414,6 +414,7 @@ why-demographic = Why does this matter?
 why-demographic-explanation-2 = Anonymized user data like age, gender, and accent helps improve the audio data used to train the accuracy of speech recognition engines. Your username and email will never be associated with your submitted data, and you can choose whether to make your username public or anonymous.
 accept-privacy = I'm okay with you handling this info as you explain in Mozilla's <privacyLink>Privacy Policy</privacyLink>
 accept-privacy-title = Privacy Policy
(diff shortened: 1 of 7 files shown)


mozgzh:
Looks good! One thought, should the button be disabled or change colour based on the state of "privacyAgreedChecked", ie if the ToS aren't checked, grey out, once agreed, blue?

moz-rotimib:
> Looks good! One thought, should the button be disabled or change colour based on the state of "privacyAgreedChecked", ie if the ToS aren't checked, grey out, once agreed, blue?

In the designs Sabrina shared, the button does not change colour depending on the state of "privacyAgreedChecked". If you try to submit the clips without checking the box, you are prevented from doing so until the box is checked

Screenshot 2022-11-18 at 13 23 13

mozgzh:
Cool, that should be fine if its in the design. Last question, how does it look on mobile?

moz-rotimib:
> Cool, that should be fine if its in the design. Last question, how does it look on mobile?

This is how it looks on the iPhone emulator in Firefox


https://user-images.githubusercontent.com/113194143/202707925-611a7adf-3577-44d1-b388-8ff5ef5704ef.mov

PR #3868, opened 2022-11-23
OI-2497 Refactor user accents api

mozgzh:
Allows for anonymous users to provide accent and variant data without creating an account.


moz-rotimib:
LGTM

I haven't tested locally yet but the logic looks good. I'll update this comment when I am able to test

moz-rotimib (approved):

PR #3876, opened 2022-12-01
Metadata Expansion for Speak flow

moz-rotimib:
Allows anonymous users to add metadata about how they speak


web/src/components/pages/contribution/speak/secondSubmissionCTA/secondSubmissionCTA.tsx
```diff
@@ -0,0 +1,86 @@
...
+              src={require('./images/happy-mars@2x.png')}
+              alt="Happy Mars"
+              width={175}
+              height={175}
+            />
+          </div>
+          <div className="header-text-container">
+            <Localized id="second-cta-header-text">
+              <h1 className="header-text">
+                Thank you for contributing your voice!
```
mozgzh, line 41: Is there a reason we're hardcoding the text here?

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -475,17 +494,61 @@ class SpeakPage extends React.Component<Props, State> {
...
+    // and the user does not have an account
+    if (hasSeenFirstCTA !== 'true' && !user.account) {
+      this.setState({ shouldShowFirstCTA: true });
+      window.sessionStorage.setItem(SEEN_FIRST_CTA, 'true');
+    }
+
+    // display second CTA screen if it has not been seen it before
+    // and the user does not have an account
+    if (
+      hasSeenFirstCTA === 'true' &&
```
mozgzh, line 534: Question: is this condition checking for the string true or the boolean true?

moz-rotimib: It's checking for the string true. You can only write string values to Session Storage

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -475,17 +494,61 @@ class SpeakPage extends React.Component<Props, State> {
...
+      hasSeenSecondCTA !== 'true' &&
+      !user.account
+    ) {
+      this.setState({ shouldShowSecondCTA: true });
+      window.sessionStorage.setItem(SEEN_SECOND_CTA, 'true');
+    }
+
+    // Reset for unauthenticated users who have seen the first and second CTA so they can see new clips to record
+    if (
+      !user.account &&
```
mozgzh, line 544: since `!user.account` is a condition thats being checked 3 times in the same function, it might make it easier to read to break up the chain of if statments like: 
```javascript
if (!user.account){
  if(otherCondition){}
  else if (otherCondition){}
//restartState etc
}```

File: web/src/components/pages/contribution/speak/speak.tsx (hunk 2 above)
Comment (mozgzh) [reply web/src/components/pages/contribution/speak/speak.tsx:534]: Mixing the type and meaning of 'true' and true is kinda confusing, it might be worth casting the string 'true' into a real boolean once, and then the conditions  will look like all the other ones

File: web/src/components/pages/contribution/contribution.tsx [hunk 4]
```diff
@@ -440,92 +375,115 @@ class ContributionPage extends React.Component<Props, State> {
...
-                    Math.abs(
-                      Math.min(Math.max(selectedPill, 1), pills.length - 2) - i
-                    ) > 1
-                      ? { display: 'none' }
-                      : {},
-                })
-              )}
+          )}
+
+          {shouldShowFirstCTA || shouldShowSecondCTA ? (
```
mozgzh, line 439: if there a lot of bespoke conditions, it can make the code easier to read to create the conditions as separate variables and then use the variable in the condition

eg
```
const shouldShowCTA = shouldShowFirstCTA || shouldShowSecondCTA;
if(showShouldCTA){}
```

mozgzh: This kind of approach can then be applied consistently across all conditions/components

mozgzh:
Since there a lot of conditionals, its a little hard to follow when things are happening. I'd say a lot of my notes follow some of this logic from [this blog post](https://blog.jetbrains.com/idea/2017/09/code-smells-if-statements/). Essentially encapsulating the common conditions into separate,reusable and obvious variables/functions.

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -475,17 +495,58 @@ class SpeakPage extends React.Component<Props, State> {
...
     this.setState({
       showPrivacyModal: false,
       showDiscardModal: !this.state.showDiscardModal,
     });
   };
 
+  private handleSubmit = (evt: React.SyntheticEvent) => {
+    const { user } = this.props;
+
+    const hasSeenFirstCTA = window.sessionStorage.getItem(SEEN_FIRST_CTA);
```
mozgzh, line 517: How about something like ```
// utils.js
const castTrueString = (strValue: string) => strValue === 'true' ? true : false;

const hasSeenFirstCTA = castTrueString(window.sessionStorage.getItem(SEEN_FIRST_CTA));

```

then in the conditions below we don't need to check the actual string value since its a boolean.

moz-rotimib: Yeah that's a really good suggestion 👌 as an improvement I think we can drop the ternary expression because the value returned will always be `false` if the string passed in is not equal to the string true and `true` if it is.

mozgzh (approved):
:rocket:

laubonghaudoi:
@mozgzh @moz-rotimib Hi, is there any reason why the profile settings for disabling "ready to do 5 more" is removed? We are having the same issue as in https://discourse.mozilla.org/t/giving-progress-notification-after-every-5-listened-sentences-is-annoying/113347/2 and this "ready to do 5 more" prompt has been slowing down our work greatly.

PR #3880, opened 2022-12-07
Oi 2539 random clips count validation

moz-dfeller:
The private method `getCountFromQuery` from clip.ts is being removed and replaced by a middleware that validates the count query parameter. The reason is that the clip class should not be responsible for query parameter validation.

### DIFF
--- a/server/src/lib/validation/random-clips-count.ts
+++ b/server/src/lib/validation/random-clips-count.ts
@@ -1,0 +1,13 @@
+import { AllowedSchema } from 'express-json-validator-middleware';
+
+export const randomClipsCountSchema: AllowedSchema = {
+  type: 'object',
+  properties: {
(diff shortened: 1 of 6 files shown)


server/src/lib/validation/random-clips-count.ts
```diff
@@ -0,0 +1,13 @@
+import { AllowedSchema } from 'express-json-validator-middleware';
+
+export const randomClipsCountSchema: AllowedSchema = {
```
mozgzh, line 3: Since the schema can actually be used across different routes, it might be better to use a more generic name for it, like `clipSchema`.

mozgzh:
Looks great, left one comment about the naming. :rocket:

mozgzh (approved):
:rocket:

PR #3882, opened 2022-12-09
Oi 2532 dataset download links to old versions have permission denied error

mozgzh:
Enable older datasets to be displayed and downloaded on frontend. This will allow users to download any previously released dataset.

### DIFF
--- a/server/src/lib/model/db/migrations/20221209022521-enable-dataset-downloads.ts
+++ b/server/src/lib/model/db/migrations/20221209022521-enable-dataset-downloads.ts
@@ -1,0 +1,21 @@
+export const up = async function (db: any): Promise<any> {
+  await db.runSql(`
+    UPDATE datasets
+    SET is_deprecated = 0
+    WHERE release_dir in (
+      'cv-corpus-1',
+      'cv-corpus-2', 
+      'cv-corpus-3',
+      'cv-corpus-4-2019-12-10',
+      'cv-corpus-5.1-2020-06-22',
+      'cv-corpus-6.1-2020-12-11',   
+      'cv-corpus-7.0-2021-07-21',
+      'cv-corpus-8.0-2022-01-19',
+      'cv-corpus-9.0-2022-04-27'
+      )
+  `);
+};
+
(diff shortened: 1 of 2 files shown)


server/src/lib/model/db/migrations/20221209022521-enable-dataset-downloads.ts
```diff
@@ -0,0 +1,21 @@
...
+      'cv-corpus-6.1-2020-12-11',   
+      'cv-corpus-7.0-2021-07-21',
+      'cv-corpus-8.0-2022-01-19',
+      'cv-corpus-9.0-2022-04-27'
+      )
+  `);
+};
+
+export const down = function (): Promise<any> {
+  return null;
```
moz-dfeller, line 20: What about the down migration where we set `is_deprecated` back to 1? Is that needed?

web/locales/en/messages.ftl
```diff
@@ -746,7 +747,6 @@ language = Language
 download-dataset-header = Download the Dataset
 download-delta-explainer = We’ve made some changes. Delta Segments just contain the most recent clips since the last release. <deltaLink>Read more about this work</deltaLink>.
 download-dataset-tag = Select the desired language dataset and choose the version you wish to download.
-download-dataset-access = Looking for previous dataset releases? Email <emailLink> { $email } </emailLink>
```
moz-dfeller, line 749: Is it enough to delete the key in the `en` file and change is propagated to the other locales or how does this work?

moz-dfeller (approved):

server/src/lib/model/db/migrations/20221209022521-enable-dataset-downloads.ts
mozgzh: Generally since this is a "content" update, not a schema update,  I didnt create an equivalent `down` as the values are both correct (true or false)

web/locales/en/messages.ftl
mozgzh: I do believe so, can be confirmed once pontoon is updated

PR #3885, opened 2022-12-15
Expand the stats API so users can query by year 

moz-dfeller:
All the stats endpoints can be queried by year now. The current year is used, if the `?year={year}` query parameter is omitted. The `year` query parameter can also be combined with existing query parameters, e.g:

GET `/api/v1/statistics/clips?filter=rejected&year=2021` HTTP/1.1

QA:
- Check the different stats endpoints, e.g. the one mentioned above, and compare whether the `yearly_sum` adds up to the `total_count`[^1]
- Use combined query parameters as well

[^1]: The `speakers` endpoint is the exception, refer to the [documentation](https://github.com/common-voice/common-voice/blob/main/docs/API/statistics.md) for more information

### DIFF
--- a/common/statistics.ts
+++ b/common/statistics.ts
@@ -1,0 +1,9 @@
+export type Filters = 'rejected' | 'hasEmail';
+
+export type QueryOptions = {
+    groupByColumn?: string;
+    isDistinct?: boolean;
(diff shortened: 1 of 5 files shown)


mozgzh (approved):
:fireworks:

PR #3900, opened 2023-01-09
OI-2480: Fix sign up link not going to the footer when clicked

moz-rotimib:
Fix sign up link not going to the footer

### DIFF
--- a/web/src/components/layout/footer.tsx
+++ b/web/src/components/layout/footer.tsx
@@ -21,10 +21,10 @@ const LocalizedLocaleLink = ({ id, to }: { id: string; to: string }) => {
   );
 };
 
-export default React.memo(() => {
+const Footer = React.memo(() => {
   const [locale] = useLocale();
   return (
-    <footer>
+    <footer id="footer">
       <div id="help-links">
         <LocaleLink to={URLS.FAQ} onClick={() => trackNav('faq', locale)}>
           <SupportIcon />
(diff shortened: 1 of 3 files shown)


moz-dfeller (approved):

web/src/services/api.ts
```diff
@@ -265,7 +265,7 @@ export default class API {
     });
   }
 
-  subscribeToNewsletter(email: string): Promise<void> {
+  subscribeToNewsletter(email: string): Promise<{ token: string }> {
```
moz-dfeller, line 268: should be void again, since we don't return the token anymore, right? :)

```suggestion
  subscribeToNewsletter(email: string): Promise<void> {
```

web/src/components/pages/about/get-involved.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+  it('renders GetInvolved component', () => {
+    renderWithProviders(<GetInvolved />);
+    expect(screen.getByText('How do I stay in touch?')).toBeDefined();
+
+    // email signup button should be visible because basket token is undefined
+    expect(screen.queryByTestId('email-signup-button')).toBeDefined();
+    expect(screen.queryByTestId('discourse-button')).toBeDefined();
+  });
+
+  it('hides email signup button if basket token is available', () => {
```
moz-dfeller, line 17: the naming of the tests is not consistent anymore, since we don't have the basket token

moz-dfeller (changes requested):

moz-dfeller (approved):
Awesome 🚀

PR #3950, opened 2023-02-10
OI-2612 Add migration for sentence collector sentences and votes

moz-dfeller:
Adding two new tables for the sentence collector migration


MichaelKohler (approved):
Didn't cross-check every field but generally looks good to me.

PR #3951, opened 2023-02-10
OI-2630 Add endpoint post a single sentence

moz-dfeller:
Adds an endpoint to submit a sentence. The sentence validation is mostly adopted from the [Sentence Collector Repo](https://github.com/common-voice/sentence-collector). A big thank you for that! @MichaelKohler 

## Note:
Must be merged together or after #3950


MichaelKohler (commented):
Added some comments but a very high-level look doesn't sound any alarm.

Overall, is the plan to move all the code to this new folder setup as well as fp-ts? While possibly easier to maintain overall on a day-job, one-off contributions might get quite complicated as fp-ts is not super straightforward to get into. But then that heavily depends on the plan on community contributions in the future.

server/src/application/sentence-collector/repository/sentence-repository.ts
```diff
@@ -0,0 +1,37 @@
...
+const insertSentence = (db: Mysql) => (sentence: Sentence): TE.TaskEither<ApplicationError, unknown> => {
+  return TE.tryCatch(
+    () =>
+      db.query(
+        `
+            INSERT INTO sc_sentences (sentence, source, locale_id, client_id)
+            VALUES (?, ?, ?, ?);
+        `,
+        [
+          sentence.sentence,
```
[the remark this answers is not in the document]
moz-dfeller: I renamed the type to `PendingSentenceSubmission`. The sentence is being submitted to be reviewed/validated by the community before it is incorporated into the CV sentences pool or rejected. Therefor the sentence is in a pending state. So I guess the name makes sense for now until a better naming comes across ;)

server/src/core/sentence-collector/validation/validation.ts
```diff
@@ -0,0 +1,86 @@
...
+  ca,
+  ckb,
+  en,
+  eo,
+  ig,
+  it,
+  kab,
+  ko,
+  lo,
+  'ne-NP': ne,
```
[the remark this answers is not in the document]
moz-dfeller: Fixed! Unfortunately we can't use hyphenated variable names in TS.

server/src/application/sentence-collector/repository/sentence-repository.ts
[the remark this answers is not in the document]
MichaelKohler: I think that makes sense, yes. Thanks!

moz-dfeller:
> Added some comments but a very high-level look doesn't sound any alarm.
> 
> Overall, is the plan to move all the code to this new folder setup as well as fp-ts? While possibly easier to maintain overall on a day-job, one-off contributions might get quite complicated as fp-ts is not super straightforward to get into. But then that heavily depends on the plan on community contributions in the future.

This was actually a point I was thinking about for a long time. There were basically two options to make the code base more maintainable, extendable and improve code reusability: 
1. go the OOP route
2. go a more functional route

In general I would say that OOP is more accessible to beginners, but this goes out the window as soon we add DI, Abstractions, Composition over Inheritance, Facades, Factories, Delegation Pattern, Strategy Pattern, and the list goes on. There is also a lot of jargon in the FP community. In the end I believe that it would be easier for most to add functionality by creating a function that takes a type A and returns a type B and plug it into a pipeline where types match up. Once there are more features built in the new architecture, people can use them as reference to create more features.

To the point of fp-ts. It doesn't have a great documentation (unfortunately) but it's not so different from learning other libraries like React, Redux or RxJS, which are also quite functional.

I was really torn between going OOP or FP but for the reasons I just mentioned, I decided to go with FP in the end.

server/src/core/sentence-collector/validation/languages/en.ts
```diff
@@ -0,0 +1,41 @@
...
+    fn: (sentence: string) => {
+      const words = tokenizeWords(sentence)
+      return words.length < MIN_WORDS || words.length > MAX_WORDS
+    },
+    error: `Number of words must be between ${MIN_WORDS} and ${MAX_WORDS} (inclusive)`,
+  },
+  {
+    type: 'regex',
+    regex: /[0-9]+/,
+    error: 'Sentence should not contain numbers',
```
moz-rotimib, line 23: ### Question

I just wanted to ask why a sentence can't have numbers. Is a sentence like:

> She has 2 cats

not valid because of the number "2"?

moz-dfeller: Yes, that would mean that the sentence
> She has 2 cats

would be rejected from submission. I am not entirely sure why that rule is necessary, though. Maybe @ftyers can help.

MichaelKohler: We have had that from the beginning basically. The reasoning back then was that for non-single-digit numbers there can be quite some ambiguity and unclarity on how to read it. For example 2020 could be read as:

- two zero two zero
- two thousand and twenty
- twenty twenty

I'm not sure how much sense this still makes, but that was the reasoning back then.

moz-rotimib: Thanks for the explanation 👍

MichaelKohler:
@moz-dfeller please note https://github.com/common-voice/sentence-collector/pull/664 which will need to be ported here as well. Do you want to include it in this PR or should I add a task for myself to port it after this is merged?

moz-dfeller:
@MichaelKohler thanks for pointing that out. I will update the PR with the changes 👍

moz-dfeller:
Changes are included in #3963

PR #3963, opened 2023-03-01
OI-2651 integrate back end changes for sc endpoints

moz-dfeller:
This PR bundles all changes on the back-end needed to provide the necessary endpoints for the initial version of the integrated sentence collector


server/src/api/pending-sentences/handler/add-pending-sentence-vote-handler.ts
```diff
@@ -0,0 +1,25 @@
...
+import { pipe } from 'fp-ts/lib/function'
+import { task as T, taskEither as TE } from 'fp-ts'
+import { StatusCodes } from 'http-status-codes'
+import { addPendingSentenceVoteCommandHandler } from '../../../application/pending-sentences/use-case/command-handler/add-pending-sentence-vote-command-handler'
+import { AddPendingSentenceVoteCommand } from '../../../application/pending-sentences/use-case/command-handler/command/add-pending-sentence-vote-command'
+import { createPresentableError } from '../../../application/helper/error-helper'
+
+export default async (req: Request, res: Response) => {
+  const command: AddPendingSentenceVoteCommand = {
+    clientId: req.client_id || null,
```
moz-rotimib, line 11: QUESTION: Should it be possible for users that are not logged in to vote on sentences? I think only logged-in users should be able to review sentences because on the frontend we are hiding the Review option for users that are not logged in

server/src/core/sentences/validation/languages/am.ts
moz-rotimib, line 1: I think it will be helpful to have an error type in the validation object so we know what type of error we are getting from the backend. Especially for languages that don't send error messages in English. (This applies to all the other validation files in the folder).

The error type could be an enum of the various errors that we could send and then in the validation object we could have something like:

```
{
    type: 'regex',
    regex: /[0-9]+ | [\u1369-\u137C]+/,
    error: "Some error",
    errorType: ERROR.IncorrectSpelling
}
```

What do you think?

moz-rotimib:
I think it will be helpful to have an error type in the validation object so we know what type of error we are getting from the backend. Especially for languages that don't send error messages in English. (This applies to all the other validation files in the folder).

The error type could be an enum of the various errors that we could send and then in the validation object we could have something like:

{
    type: 'regex',
    regex: /[0-9]+ | [\u1369-\u137C]+/,
    error: "Some error",
    errorType: ERROR.IncorrectSpelling
}

What do you think?

moz-dfeller:
Closes: #3961 #3960 #3959 #3950

PR #4014, opened 2023-04-25
chore: update docker image debian from stretch to bullseye

moz-dfeller:


moz-rotimib (approved):
So one line was the solution 😅

moz-dfeller:
> So one line was the solution 😅

As so often

PR #4067, opened 2023-06-02
Will not override existing value of is_used

raivisdejus:
### Type of Pull Request

- [ ] Bulk sentence upload 

* https://github.com/common-voice/common-voice/issues/4061

- [ ] Other

The `is_used` column of the `sentences` table can be used to hide sentences from the UI, but the current codebase would override any values of this column during the sentence import. This PR will fix that, the value of `is_used` will be preserved.

This will allow removing problematic sentences from the Common Voice UI without removing their recordings.
The use case of this is to remove offensive sentences from the system retaining their recordings.

### DIFF
--- a/server/src/lib/model/db/import-sentences.ts
+++ b/server/src/lib/model/db/import-sentences.ts
@@ -124,8 +124,7 @@ async function importLocaleSentences(
                 .join(', ')}
               ON DUPLICATE KEY UPDATE
                 source = VALUES(source),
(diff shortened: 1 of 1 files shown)


moz-dfeller:
Hey @raivisdejus 
I guess that makes sense! Thank you very much for your effort and sorry for the delay!

moz-dfeller (approved):

PR #4078, opened 2023-06-18
Will hide problematic latvian sentences

raivisdejus:
### Type of Pull Request

- [ ] Bulk sentence upload 
- [x] Other

This PR should be merged after https://github.com/common-voice/common-voice/pull/4067

PR will hide from the UI sentences that could be offensive to some people. Recordings are ok and should be kept, we want to just hide these sentences from further recordings.


server/src/lib/model/db/migrations/20230518103600-hide-problematic-latvian-sentences.ts
```diff
@@ -0,0 +1,104 @@
+export const up = async function (db: any): Promise<any> {
+  return db.runSql(`
+    UPDATE sentences SET is_used = FALSE WHERE text in (
```
moz-dfeller, line 3: We don't want to overwrite sentences from other locales by accident. So it would be best to add the latvian locale id as constraint as well.

```suggestion
  return db.runSql(`
    UPDATE sentences SET is_used = FALSE WHERE locale_id=(SELECT id from locales WHERE name='lv') AND text in (
```

moz-dfeller (changes requested):
Could you please

server/src/lib/model/db/migrations/20230518103600-hide-problematic-latvian-sentences.ts
raivisdejus: True, better safe than sorry

raivisdejus:
@moz-dfeller Added a check for `lv` locale

moz-dfeller (approved):

PR #4136, opened 2023-07-31
OI-2740: Show toast to user for rate limit exceeded error

moz-rotimib:
- Shows toast error message to users if they exceed the rate limit while reviewing sentences.
- Also refactors Review component by moving some logic to a custom hook

Screenshot 2023-08-08 at 08 46 53


moz-dfeller (approved):
Good job. I like the refactor :)

PR #4137, opened 2023-08-01
Will mark approved sentences as valid

raivisdejus:
### Type of Pull Request
* https://github.com/common-voice/common-voice/issues/4133

A migration to make all currently approved sentences validated and usable for recording
This is a temporary solution while we wait for a more permanent solution.


moz-dfeller:
Hi @raivisdejus , thank you for the migration! I was running it myself on the production machine (after testing ofc). It was a total of 17861 sentences that are now in the corpus.

Thanks again for the great work. I could probably use that migration as a starting point for futures scheduled sentences updates 👍

raivisdejus:
@moz-dfeller I think if this is run as part of the quarterly dataset release, the community will also be fine. 
Thanks for unblocking us and keeping care of Common Voice!

PR #4141, opened 2023-08-08
OI-2744 Delete profile voice recordings toggle flipped logic in rtl languages

moz-rotimib:
Fix default position for delete toggle for rtl languages.

https://github.com/common-voice/common-voice/assets/113194143/3f69674c-54c0-4e1d-befb-c83674ca3237


moz-rotimib:
@kwikwag can you confirm from the video if the delete toggle is fixed for RTL languages like Hebrew?

PR #4146, opened 2023-08-15
Will hide problematic sentences

raivisdejus:
### Type of Pull Request

* https://github.com/common-voice/common-voice/issues/4138

This PR will hide sentences that have `?` in the middle of some words in sentence text.

Check for locale on first query is omitted on purpose as this issue is affecting multiple locales and it is hard to imagine some language where `?` in the middle of the word is a valid grammatical structure. If necessary, I would be happy to adjust the PR.

Checking Latvian sentences noticed also that some sentences do not have any text, so this PR will also hide those.

To run a query manually one would use 
```
UPDATE sentences SET is_used = FALSE
    WHERE (created_at > '2023-05-01 00:00:00' AND text REGEXP '.*\\?[a-zāčēģīķļņšūžâîûüöç].*') OR text = ''
```
and
```
UPDATE sentences SET is_used = FALSE
WHERE locale_id=(SELECT id from locales WHERE name='lv')
  AND source REGEXP '.*\\?[a-zāčēģīķļņšūžâîûüöç].*'
```

### DIFF
--- a/server/src/lib/model/db/migrations/20230815103600-hide-problematic-sentences.ts
+++ b/server/src/lib/model/db/migrations/20230815103600-hide-problematic-sentences.ts
@@ -1,0 +1,10 @@
+export const up = async function (db: any): Promise<any> {
+  return db.runSql(`
+    UPDATE sentences SET is_used = FALSE
+    WHERE text REGEXP '.*\\\\?[a-z].*' OR text = ''
+  `);
+};
+
+export const down = async function (): Promise<any> {
+  return null;
+};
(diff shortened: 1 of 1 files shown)


HarikalarKutusu:
I wish we had a datetime stamp on entires... Belease be aware of the following constructs, which are many in out text-corpus for example:

> "Where will you go?", she asked.

raivisdejus:
>"Where will you go?", she asked.

This is safe as there is no letter after the `?`  Checked this case.

Only grammatically incorrect sentences without space after the `?` would get disabled like `Where will you go?He was asking.`

There `created_at` column for the sentences, will add that as parameter

HarikalarKutusu:
Sorry @raivisdejus, I was on mobile and didn't check the code at that time. I never used REGEXP in SQL but `[a-z]` might not cover all letters in Latin-1. AFAIK, as of now we think only non-Latin-1 characters are affected. So, A-Z âîû üöç ÜÖÇ etc can also be there.

raivisdejus:
@HarikalarKutusu Added extra letters to the query.

moz-dfeller (approved):
Thank you very much. I tested the where clause before hand. Looks good to me!

PR #4171, opened 2023-09-15
feat: add cli tool to manually import bulk submission

moz-dfeller:


server/src/api/cli/import-sentences.ts
```diff
@@ -0,0 +1,48 @@
...
+import { AddBulkSentencesCommand } from '../../application/sentences/use-case/command-handler/command/add-bulk-sentences-command'
+import * as fs from 'fs'
+import { AddBulkSentencesCommandHandler } from '../../application/sentences/use-case/command-handler/add-bulk-sentences-command-handler'
+import { pipe } from 'fp-ts/lib/function'
+import { taskEither as TE, task as T } from 'fp-ts'
+
+const importSentences = async (args: any, options: any) => {
+  const config = getConfig()
+  const languagesResponse = await fetch(
+    `http://localhost:${config.SERVER_PORT}/api/v1/languages`
```
moz-rotimib, line 14: Will this url be correct in production? With localhost I mean

moz-dfeller: I think so. The server is listening on the `SERVER_PORT`, so it should be correct. The confusion might be that this is supposed to be called locally via the cli like so `node import-sentences.js -f filepath -l en -cid randomstring`. This is just a temporary solution to import bulk submissions manually.

moz-rotimib (approved):

PR #4216, opened 2023-10-23
OI-2745: Fix for changing language when active on a contribution session

moz-rotimib:
Fix for changing language when active on a contribution session


moz-dfeller (approved):
🚀

PR #4225, opened 2023-11-03
OI-2805 Upgrade React & Typescript

moz-rotimib:
Upgrades React & Typescript


moz-rotimib:
Closing this because upgrading React will involve upgrading react router which has breaking changes that we need to figure how to implement. So for now we are going to upgrade Typescript first.

PR #4231, opened 2023-11-10
OI-2817: Add tracking for events on the write page

moz-rotimib:
Adds tracking for specific events on the Write page


web/src/services/tracker.ts
```diff
@@ -146,6 +148,25 @@ export function trackChallenge(
   track('Challenge', action);
 }
 
+export function trackSingleSubmission(
+  action: 'toggle-button-click' | 'submit',
+  locale: string
+) {
+  track('SingleSubmission', action, locale);
+}
```
moz-dfeller, line 156: Does `track` return something? Because then you could just do it like this:
```suggestion
export trackSingleSubmission = (
  action: 'toggle-button-click' | 'submit',
  locale: string
) => track('SingleSubmission', action, locale);
```
This way the caller can decide if they want to throw away the result or not. The same goes ofc for the other function.

moz-dfeller (commented):
Looks good! Just a small question :)

moz-dfeller (approved):

PR #4240, opened 2023-11-20
OI-2807: 'fix' typescript issue in user store

moz-rotimib:
Fix Typescript namespaces linting issue.

### DIFF
--- a/web/src/stores/user.ts
+++ b/web/src/stores/user.ts
@@ -7,6 +7,25 @@ export const VISIBLE_FOR_NONE = 0;
 export const VISIBLE_FOR_ALL = 1;
 export const VISIBLE_FOR_TEAM = 2;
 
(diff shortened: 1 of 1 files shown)


web/src/stores/user.ts
```diff
@@ -7,6 +7,25 @@ export const VISIBLE_FOR_NONE = 0;
...
+    validateTally: 0,
+    isSubscribedToMailingList: false,
+
+    userClients: [],
+    isFetchingAccount: true,
+    account: null,
+  };
+}
+
+// eslint-disable-next-line @typescript-eslint/no-namespace
```
moz-rotimib, line 28: "Fixing" this by disabling the eslint rule because the alternative is a lot of effort for something that does not give us any immediate value. Perhaps we can revisit this in the future

PR #4248, opened 2023-11-23
OI 2797: Add E2E tests for Listen page

moz-rotimib:
Adds e2e tests for Listen page


moz-rotimib:
@moz-dfeller I guess we can merge this PR as it is now (if it is ok) but running these tests as part of our CI should be a separate ticket. What do you think?

PR #4250, opened 2023-11-23
OI-2855: Show last updated timestamp on languages page

moz-rotimib:
Show last updated timestamp on languages page

### DIFF
--- a/web/src/components/pages/languages/languages.tsx
+++ b/web/src/components/pages/languages/languages.tsx
@@ -6,6 +6,7 @@ import {
 } from '@fluent/react';
 import * as React from 'react';
 import { useState, useEffect } from 'react';
+import { format } from 'date-fns';
 import { BaseLanguage, LanguageStatistics } from 'common';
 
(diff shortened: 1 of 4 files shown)


web/src/components/pages/languages/languages.tsx
```diff
@@ -256,6 +257,7 @@ const LanguagesPage = ({ getString }: WithLocalizationProps) => {
   // on mount
   useEffect(() => {
     loadData().then(({ localeMessages, languageStatistics }) => {
+      console.log({ languageStatistics });
```
moz-dfeller, line 260: Debug statement?

moz-dfeller (changes requested):

web/src/components/pages/languages/languages.tsx
moz-rotimib: Good catch. I'll remove that now

moz-dfeller (approved):
:rocket:

PR #4264, opened 2023-12-05
OI 2844: Add donate button on the download modal page segment

moz-rotimib:
Add donate button to the datasets download page


moz-dfeller (changes requested):
Just fixing the test, rest looks good to me :)

ftyers (commented):
Only one question: For the `why-email` we have the question in bold, but for the `why-donate` we don't have the question in bold, there is a bold part of the sentence at the beginning. In other languages this might not be placed at the beginning. Would the bolding still work if it were at the end of the sentence?

ftyers:
> > Only one question: For the `why-email` we have the question in bold, but for the `why-donate` we don't have the question in bold, there is a bold part of the sentence at the beginning. In other languages, this might not be placed at the beginning. Would the bolding still work if it were at the end of the sentence?
> 
> @ftyers I think @ moz-michelleuv can answer this when she gets back. However, I want to clarify when you say that we don't have the question in bold for the `why-donate` - in the screenshot, the question at the end is in bold text. So I want to clarify if you're referring to this question or something else.
> Screenshot 2023-12-05 at 16 05 18

In this part,

imatge

The "Why email?" is in bold, which will come at the beginning in all languages (because it's a separate sentence). My point is that by bolding:

- **We are X,** Y, and Z.

In some languages that would be realised as:

- **X**, Y, and Z-**we.are**

The question is if that is intentional, or do we want to change how the bolding/emphasis works.

PR #4304, opened 2024-01-08
Oi 2879 enable the ability to retrieve data for full delta releases

moz-dfeller:


server/src/core/datasets/types/dataset.ts
```diff
@@ -0,0 +1,35 @@
...
+  valid_clips_duration: number
+  average_clips_duration: number
+  total_users: number
+  size: number
+  checksum: string
+  release_date: string
+  name: string
+  release_dir: string
+  download_path: string
+  splits: DatasetSplits
```
moz-rotimib, line 15: I was wondering if the `splits` field should be optional since it's possible for some datasets not to have it.

moz-rotimib (approved):

PR #4326, opened 2024-01-18
OI 2878: Implement metadata viewer for mobile

moz-rotimib:


web/src/components/pages/datasets/metadata-viewer/types.ts
```diff
@@ -0,0 +1,41 @@
...
+  download_path: string
+  splits?: Splits
+}
+
+export type Splits = {
+  gender: Gender
+  age: Age
+}
+
+type Age = {
```
moz-dfeller, line 23: If you export `type Age` you could use it as type for your keys in the `AGE_MAPPING` type. That should give you some more type safety.

web/locales/en/messages.ftl
```diff
@@ -813,6 +822,11 @@ dataset-version = Version
 sha256-checksum-copied = SHA256 Checksum Copied!
 sha256-checksum-copied-error = Failed to copy SHA256 Checksum
 clipboard-not-supported = Clipboard not supported
+no-information-available = No information available
+# dataset metadata
+dataset-metadata-sex = Sex
+# dataset metadata
+dataset-metadata-age = Age
```
moz-rotimib, line 829: Hello @ftyers there's already a translation key that maps to `Age` [here](https://github.com/common-voice/common-voice/blob/e6d43f9bbdfb3c2c0ea2f00357f8715bc836a92c/web/locales/en/messages.ftl#L393) but the key wouldn't make sense to use in the context of the datasets page.

Is it ok to create a new key that maps to `Age` or does it make sense to use the existing key?

ftyers: This is a good question. This is such a short string, and most translators will get it from the translation memory that I think there is no problem duplicating the string. But I would add a comment for translators that this is the age of the contributor, not the age of the dataset.

moz-rotimib: Thanks!

PR #4362, opened 2024-02-05
OI-2905: Add sentence domain dropdown and styling

moz-rotimib:


web/src/components/pages/contribution/sentence-collector/write/sentence-input-and-rules/constants.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const SENTENCE_DOMAIN_MAPPING: Record<SentenceDomain, string> = {
+  agriculture: 'Agriculture',
+  automotive: 'Automotive',
+  finance: 'Finance',
+  food_service_retail: 'Food, Service and Retail',
+  general: 'General',
+  healthcare: 'Healthcare',
+  history_law_government: 'History, Law and Government',
+  language_fundamentals: 'Language Fundamentals (e.g. Digits, Letters, Money)',
```
ftyers, line 11: In general, I think that this is ok, but I might make the explanations a separate string. E.g. there is only an explanation for one `language_fundamentals`, but I think there could potentially be explanations for various.

ftyers (commented):
Looks fine, only why does only "Language fundamentals" have an `(e.g. ...)` part, and could that part be in a separate string?

PR #4379, opened 2024-02-22
OI-2919: Add sex or gender select option to first post submission CTA

moz-rotimib:
Screenshot 2024-02-22 at 16 33 28


web/locales/en/messages.ftl
```diff
@@ -1612,7 +1612,7 @@ partnerships-our-partners = Our Partners
 
 # FIRST POST SUBMISSION CTA
 first-cta-header-text = Thank you for donating your voice clips!
-first-cta-subtitle-text = Would you like to share some information about how you speak?
+first-cta-subtitle-text = Would you like to share some information about yourself? It helps developers build more inclusive technology
 why-donate = Why are you asking?
 why-donate-explanation-1 = This information is optional to provide - you should only add it if you feel comfortable. When we release the dataset, we pseudonymise all personal information, and make sure information like your name and email are not connected to demographic information like age or sex.
```
ftyers, line 1617: *pseudonymise* is perhaps a difficult term to translate. Could we come up with a clearer way of saying this? Something like "we replace personal information with placeholders..." (although I'm not too keen on "placeholder" here either), could we be more specific about the process?

web/locales/en/messages.ftl
```diff
@@ -1612,7 +1612,7 @@ partnerships-our-partners = Our Partners
 
 # FIRST POST SUBMISSION CTA
 first-cta-header-text = Thank you for donating your voice clips!
-first-cta-subtitle-text = Would you like to share some information about how you speak?
+first-cta-subtitle-text = Would you like to share some information about yourself? It helps developers build more inclusive technology
```
ftyers, line 1615: Maybe "By telling us about yourself you will help develoeprs build more inclusive technology", (make it explicit who is doing the helping and how they are helping). But the original is ok too...

ftyers (commented):
A couple of minor suggestions...

web/locales/en/messages.ftl
emlewjong: I think that's better

web/locales/en/messages.ftl
emlewjong: legal likes pseud. sorry!

PR #4396, opened 2024-03-12
Update sentence bands for luo/dav/kln/azz

ftyers:
Update the sentence bands to 750 sentences for `kln`, `dav`, `luo` and `azz`.


moz-rotimib (commented):
LGTM

PR #4407, opened 2024-03-22
OI-2970: update metadata viewer to show all sex/gender options

moz-rotimib:


moz-rotimib:
> The changes look fine, but it's a bit unclear to me what "No information" means, because below that header there appears to be information. Is it "no age information" ?

Yeah @ftyers it means "no age information". It's in AC for the ticket

PR #4425, opened 2024-04-02
OI-2975: Option to select variant in profile front end

moz-rotimib:
Screenshot 2024-04-02 at 17 31 45


ftyers (commented):
In general, maybe this should be the two options and then a longer explanation?  (like for the variant and accent field).

Should we let people know that not all sentences/clips are tagged with a variant, so that by choosing only their variant they might not get all clips in that variant?

moz-rotimib:
> In general, maybe this should be the two options and then a longer explanation? (like for the variant and accent field).
> 


- There's a longer explanation before the radio options (check screenshot) if you're referring to something else please kindly clarify

Screenshot 2024-04-03 at 14 46 39

- > Should we let people know that not all sentences/clips are tagged with a variant, so that by choosing only their variant they might not get all clips in that variant?

   There's a [design](https://www.figma.com/file/L3lo7IRaD9m4XvsXkjCGr1/cv-platform-production?node-id=142%3A1179&mode=dev) for that which will be implemented in an upcoming ticket @ftyers

moz-dfeller (approved):
👍

PR #4427, opened 2024-04-04
OI-2971: Add Variant option to sentence submission UIs

moz-rotimib:
- Adds select component to select sentence variant on Write page
- Fix issue with arrow keys and enter keys not working with sentence domain combobox
- CSS fixes


web/locales/en/messages.ftl
```diff
@@ -1794,6 +1793,42 @@ nature_environment = Nature and Environment
 news_current_affairs = News and Current Affairs
 # Sentence Domain dropdown option
 technology_robotics = Technology and Robotics
+sentence-variant-select-label = Sentence Variant
+sentence-variant-select-placeholder = Select a variant (optional)
+sentence-variant-select-all-variants = General language / all variants
+
+## LANGUAGE VARIANT CODES
```
ftyers, line 1800: Do we want these hardcoded? They might be localised, and they might change...

ftyers (commented):
Some minor comments...

web/locales/en/messages.ftl
```diff
@@ -1794,6 +1793,42 @@ nature_environment = Nature and Environment
 news_current_affairs = News and Current Affairs
 # Sentence Domain dropdown option
 technology_robotics = Technology and Robotics
+sentence-variant-select-label = Sentence Variant
+sentence-variant-select-placeholder = Select a variant (optional)
+sentence-variant-select-all-variants = General language / all variants
```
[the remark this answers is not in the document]
emlewjong: Good thought - suggest "General language / multiple variants"

web/locales/en/messages.ftl
ftyers: That's good, is it possible to transclude it/import it in any way (to avoid having the error-prone) manual copy/paste. Maybe it is already like that, and just I'm missing part of the story.

moz-dfeller: I believe we do that with some other data at some point. I think we can just update this `messages.ftl` file on start-up and paste the variants from the database.

moz-rotimib: I've created a ticket for this - https://mozilla-hub.atlassian.net/browse/OI-3062

PR #4463, opened 2024-04-29
OI-3083: Update sentence domain dropdown list

moz-rotimib:
Screenshot 2024-04-29 at 16 26 59


moz-rotimib:
Hey @moz-dfeller I made some changes to `common/taxonomies.ts` and `server/src/core/domains/domain.ts` so I was wondering if the BE part of this ticket should be released before the FE because I think the BE still uses the old ids defined in the files I mentioned.

moz-dfeller (approved):
LGTM, just need to create the DB migration and we can merge all together 🚀

Issue #3779, opened 2022-08-25
[BUG] Wording without i10n on dashboard/goals page


HarikalarKutusu:
**Describe the bug**
It is reported that there are some wordings that do not go through localization. Please look at the following Discourse post for screenshots:

https://discourse.mozilla.org/t/cant-locate-some-ui-strings/103752/8

**To Reproduce**
1. Use a non-English setting, login and go to dashboard/goals https://commonvoice.mozilla.org/tr/dashboard/tr/goals
2. Check the screenshots on the post

E.g.: https://commonvoice.mozilla.org/tr/dashboard/tr/goals

**Additional context**
Found two of them in the code, their locations are also in that Discourse post.

robovoice1:
On mobile i found these:
Also in selected german language displaying the weekly achieved  goals for german,  and no goals set for french (also with selected german ui)

![Screenshot_20220917-191558](https://user-images.githubusercontent.com/92784377/190869403-f0022537-d4c1-4c80-873b-430123f1d064.png)

![Screenshot_20220917-191639](https://user-images.githubusercontent.com/92784377/190869421-a949a7cb-083c-4029-8392-3571d395230d.png)

👍

robovoice1:
![Screenshot_20220917-194945](https://user-images.githubusercontent.com/92784377/190870054-314069b4-ce9b-4fbd-8ed1-2162f97c6a0c.png)

robovoice1:
Also in languages page 
https://github.com/common-voice/common-voice/issues/3853

robovoice1:
More not translated english leftovers in german cv displaying my german goals

![Screenshot_20221112-232652](https://user-images.githubusercontent.com/92784377/201497682-70bf93ba-ee9a-41e9-9271-24548900d99c.png)

robovoice1:
https://github.com/common-voice/common-voice/issues/3888

adina-v:
Another unlocalized string under the "Donate" tab in the "Datasets" page:
![image](https://github.com/common-voice/common-voice/assets/96519317/88c5c249-a479-4f6a-8a46-cfac6d298d8d)

robovoice1:
Also tapping on spenden (donate) in de language, the next page is english.
![Screenshot_20240120-161654](https://github.com/common-voice/common-voice/assets/92784377/b7ce53b1-8e0b-4f58-8102-babd8a3b951b)
[...]

PR #4494, opened 2024-05-31
Fix untranslated strings in Custom Goals #3779

dpaleino:
### Type of Pull Request

- [ ] Bulk sentence upload 

* https://github.com/common-voice/common-voice/issues/3779
* https://github.com/common-voice/common-voice/issues/4315

- [ ] Other

* Use localized strings instead of hardcoded ones


HarikalarKutusu:
I'm not sure how it can be done given the graphic / layout, such constructs are always problematic in localization:
```
<CircleProgress ...> of daily goal
```
becomes (in Turkish):
```
Günlük hedefin <CircleProgress ...>'si
```
And the suffix would change with different numbers. Similar problems are valid for many languages (Amharic: `የዕለታዊ ግብ 50%`, Bengali: `দৈনিক লক্ষ্যের 50%`, ...)...
In my designs, I always try to prevent them.
Maybe removing the proposition and changing it to `<CircleProgress ...> (weekly goal)` in the original might help.

The alternative (using graphics as part of Fluent), will be bad for the layout.

dpaleino:
> Maybe removing the preposition and changing it to `<CircleProgress ...> (weekly goal)` in the original might help.

Thank you for the feedback, it definitely makes sense!

Also, we can use the weekly-goal and daily-goal strings, which are already available in translations.

I updated the commit with your suggested change.

web/src/components/pages/dashboard/goals/custom-goal.tsx
```diff
@@ -194,6 +194,7 @@ export default function CustomGoal({
   dashboardLocale: string;
 }) {
   const { history, location } = useRouter();
+  const l10n = useLocalization();
```
HarikalarKutusu, line 197: I think the correct use is:
```
const {l10n} = useLocalization();
```
This is what is exposed.
See: https://github.com/projectfluent/fluent.js/blob/main/fluent-react/src/use_localization.ts

HarikalarKutusu:
The existing PR is half-done because of wrong use of the Fluent related libraries. At that time I tried to show some directions to @dpaleino in a Matrix thread, but it is not yet finished.

web/src/components/pages/dashboard/goals/custom-goal.tsx
dpaleino: Thank you, fixed now

dpaleino:
@HarikalarKutusu @moz-kathyreid please have a look again, I fixed the wrong import

PR #4605, opened 2024-08-29
add sentence band migration for inhunuma languages

ftyers:
This is a sentence band migration for the languages being collected next week. Each of them should be band A.


moz-dfeller (approved):
Thanks @ftyers ! Looks good to me 😉

PR #4638, opened 2024-10-08
Update Circassian language names in "en"

HarikalarKutusu:
## Pull Request Form

- [x] Other

Although having separate ISO codes, many linguists and ALL Circassians name the language as "Circassian" and look at the Adyghe `ady` and Kabardian `kbd` as variants of the same language. Or, some name the language as Adyghe and consider Kabardian as a variant.  In any case, especially in diaspora countries (where their population is much larger than homeland) the Circassians live mostly together, and the language is called Circassian. But, the word "Circassian" (_Çerkesçe_ in Turkish) does not exist anywhere on CV.

But, internationally, Adygea is also named "West Circassian" and Kabardian is also named as "East Circassian".

This small PR changes the language names by adding West/East specs to the language names. I think this would solve or ease the problem.

For review of @ftyers


moz-dfeller:
Taken care of in #4697 
Feel free to review @HarikalarKutusu :)

PR #4649, opened 2024-10-16
613 implement sticky navigation menu  update nav links

moz-rotimib:


web/locales/en/pages/home.ftl
```diff
@@ -85,3 +85,16 @@ help-contribute = You can help build a diverse, open-source dataset by creating
 login-company = Log In / Sign Up with { $company } email
 profile-not-required = Having a profile is not required to contribute though it is helpful, see why below.
 read-more-about = Read more on our About page
+
+# Nav items
+read-sentences = Read Sentences
```
moz-dfeller, line 90: Can we put the navigation items that correspond to specific pages, e.g. `read-sentences` inside the `pages/contribute/speak.ftl`? This way we ensure that the navigation items are translated when we have a new language launch.

web/src/utility.ts
```diff
@@ -186,3 +186,6 @@ export const invalidSmallBatchSentencesToTSVString = (
 
   return tsvString
 }
+
+export const typedObjectKeys = <T extends object>(object: T): (keyof T)[] =>
```
moz-dfeller, line 190: Interesting!

PR #4655, opened 2024-10-20
[FIX] re-record message layout for RTL

Goudarz:
* [FIX] re-record message layout for RTL

**Screenshots**

Before:
Before

After:
After


moz-rotimib:
Thanks for your contribution 🎉

PR #4662, opened 2024-10-22
Add Circassian (ady & kbd) variants

HarikalarKutusu:
This PR adds variants for Circassian languages (ady & kbd) for different/major countries & scripts.

Here is it in table form:

image

Reviewer @ftyers 

Edit-1: Added Kabardian reference to variant
Edit-2: Changed transliteration bcp-47 code from `<lc>-Latn-TR-t-<lc>-cyrl-tr` to  `<lc>-Latn-TR-t-<lc>-cyrl` as some of the sentences would be common to all variants and those will also be transliterated to Latin-Turkish.


moz-dfeller (approved):
Thank you very much @HarikalarKutusu 🚀

PR #4710, opened 2024-12-05
616 add section for developers

moz-rotimib:


HarikalarKutusu:
Small fix: It should be STT (not SST)...

developers-section-subtitle = Datasets for ASR, SST, TTS, and other NLP contexts - created through community participation.

jgrant29:
@HarikalarKutusu thanks for pointing this out!  We will fix it asap.

PR #4729, opened 2024-12-26
Add Svan (sva) variants

HarikalarKutusu:
This PR is for @razmikb to speed up things during holidays...

image

Ref: https://en.wikipedia.org/wiki/Svan_language#Dialects

It has been worked with local linguists and decided to put subdialects as accents (as they actually are). They will follow in another PM.

For review by @ftyers :)


moz-dfeller:
Thanks @HarikalarKutusu 👍

PR #4730, opened 2024-12-26
Add Svan (sva) preset accents

HarikalarKutusu:
This PR is for @razmikb ...

Ref: https://en.wikipedia.org/wiki/Svan_language#Dialects

For review by @ftyers :)


moz-dfeller:
Thanks @HarikalarKutusu ! I was handling the issue in #4732 because I wanted to avoid using migrations for accents since it causes issues when setting up a dev environment for the first time.

PR #4786, opened 2025-02-06
Allow long text to fit sentence area

### LINKED ISSUE #4758: [UX-BUG/FR] Longer sentences do not fit into cards causing UX & recording problems
I'm dealing with Circassian (ady & kbd) and the language(s) contain 60+ sounds thus the alphabet is large. So one sound is represented by multiple symbols. I don't have any statistical/calculated values as they are not released yet, but I roughly assumed 150 symbols for 10 sec recordings will be appropriate. Plus there are many symbols which are displayed using wide characters.

This results in longer strings which do not fit into provided area, thus a vertical scroller appears and this causes UX issues as people need to scroll while recording and there may be disruptions in speech because of that etc - especially where people use touchpads. We are trying to support them by providing solutions (zoom in/out, change window size etc.) for now, but this is not a solution of course, especially with older and non-tech-savvy people we are dealing in these _heritage languages_.

This happens if there are 140-150 symbols for 10 sec limited recordings, and if we increase the limit to 15 secs (which we are planning) sentences will definitely not fit.

There might be two approaches:
1. **LONG TERM:** For that part use **fluid typography** using some CSS3 so that regardless of the length of the text it will be resized to fit (inside min-max limits).
2. **SHORT TERM FIX:** By default, on a 1080p screen the font is 32px, so dropping it a bit (e.g. **26-28px**) would solve most of the problems.

Alternative: Put A+ / A- buttons to change font size (which I personally hate).

Additionally, the padding surrounding text is large.

I propose a **quick fix** using no. 2 and planning for/testing no. 1 (which might need some work to test on multiple devices/screen sizes).

Edit - Screenshots:

ADY, 14 words, 134 symbols:

Image

KBD, 14 words, 151 symbols:
Image

HarikalarKutusu:
This fixes #4758 for most usual 1080p screens by using smaller font and smaller left/right padding, letter-spacing, and line-height values.

Note that:
- The default 14 words/sentence (estimated from English) should be about 1.5x14 = 21 words for 15 sec recordings.
- Although the recording limit is increased the frontend is not adapted for longer texts.
- Frontend is most probably tested against English texts, where average word length is ~5 symbols. On the other hand we calculated ~7 symbols/word average for Kabardian (similar for Adyghe). Also many symbols are wide, like Ж Д Ш Ю

This is tested for Kabardian ON THE BROWSER for a 21 word sentence for 15 sec recordings. Here are the results:

For Speak:
image

For Listen with variant pill:
image

### DIFF
--- a/web/src/components/pages/contribution/contribution.css
+++ b/web/src/components/pages/contribution/contribution.css
@@ -614,14 +614,14 @@
 
         @media (--md-up) {
             padding: 0 25px;
-            font-size: var(--font-size-xl);
+            font-size: var(--font-size-lg);
             font-weight: normal;
-            letter-spacing: 1.3px;
-            line-height: 1.63;
+            letter-spacing: 1.2px;
+            line-height: 1.5;
         }
 
         @media (--lg-up) {
-            padding: 0 100px;
+            padding: 0 50px;
         }
     }
 
--- a/web/src/components/pages/contribution/sentence-collector/review/review.css
+++ b/web/src/components/pages/contribution/sentence-collector/review/review.css
@@ -122,14 +122,14 @@
 
             @media (--md-up) {
                 padding: 0 25px;
-                font-size: var(--font-size-xl);
+                font-size: var(--font-size-lg);
                 font-weight: normal;
-                letter-spacing: 1.3px;
-                line-height: 1.63;
+                letter-spacing: 1.2px;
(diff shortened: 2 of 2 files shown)


moz-dfeller:
Hey @HarikalarKutusu thanks for your effort 🚀

PR #4787, opened 2025-02-06
feat: add platform to language request

moz-dfeller:
Adding the platform field to the language request email.

### DIFF
--- a/server/src/lib/validation/send-language-request.ts
+++ b/server/src/lib/validation/send-language-request.ts
@@ -1,8 +1,8 @@
-import { AllowedSchema } from 'express-json-validator-middleware';
+import { AllowedSchema } from 'express-json-validator-middleware'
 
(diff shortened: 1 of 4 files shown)


server/src/lib/validation/send-language-request.ts
```diff
@@ -14,5 +14,9 @@ export const sendLanguageRequestSchema: AllowedSchema = {
     languageLocale: {
       type: 'string',
     },
+    platform: {
+      type: 'string',
+      enum: ['common-voice', 'spontaneous-speech'],
```
moz-rotimib, line 19: @moz-dfeller I've asked for confirmation from Michelle but I think it might be possible for users to select both common voice and spontaneous speech as the platforms so maybe there should be a third option or is it possible to indicate somehow that both options were selected?

P.S. I think the `common-voice` enum should be updated to `scripted-speech`

PR #4803, opened 2025-02-21
Migrate Circassian common sentences to <lc>-Cyrl variants

HarikalarKutusu:
This is a follow-up for #4791 for Circassian languages (`ady` and `kbd`).

If the variant field is empty, they get "moved" to `<lc>-Cyrl` to distinguish from others in profile.

Until now, except recently added `<lc>-Latn-TR-t-<lc>-Cyrl` sentences, there are other sentences without variant info, so it is safe.

EDIT: The first commit has already been merged, but here will silently be processed. I hope it is OK.

### DIFF
--- a/server/src/lib/model/db/language-data/variants.ts
+++ b/server/src/lib/model/db/language-data/variants.ts
@@ -192,6 +192,11 @@ export const VARIANTS: Variant[] = [
   },
   {
     locale_name: 'ady',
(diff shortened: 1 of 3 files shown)


moz-dfeller:
ℹ️ There was an issue with the migrations which was fixed here: d4747c69b9d682d28caa1ede0a77b4a60e3c3e5e

HarikalarKutusu:
Thank you! Greatly appreciated 💯...

Issue #3907, opened 2023-01-16
[BUG] The string activity-needed-calculation-plural is unnecessarily complex


eemeli:
**Describe the bug**
Right now, this string breaks Pontoon in Welsh, as lifting the selectors to the message's top level results in 4^6 = 1296 variants. It's also unnecessarily complex, given that its input values are and have always been hard-coded:
https://github.com/common-voice/common-voice/blob/d19ad06c6b470b06af0fd2eefe3f0e84cca09235/web/src/components/pages/dashboard/goals/custom-goal-steps.tsx#L197-L204

**To Reproduce**
1. Go to https://pontoon.mozilla.org/ru/common-voice/web/locales/en/messages.ftl/?string=213590
2. Observe that an excessive number of edit boxes are presented.
3. Go to https://pontoon.mozilla.org/cy/common-voice/web/locales/en/messages.ftl/?string=213590
4. Observe that the browser freezes, and eventually asks the user to stop the script.

**Expected behavior**
The message should not presume that the fixed inputs are actually variable. Here's a valid alternative that's much easier to translate:
```fluent
activity-needed-calculation-plural =
  10,000 hours is achievable in just over 6 months if 1,000 people record 45 clips a day.
```

**Screenshots**
<img width="442" alt="Screenshot 2023-01-16 at 11 38 47" src="https://user-images.githubusercontent.com/617000/212657384-87bc4450-cacf-4532-9c0b-aed803d96e16.png">

**Additional context**
mozilla/pontoon#2703

It should not count as a feature that this particular string is one we tend to use to stress-test Fluent, as it's the most complex one in all of Mozilla's localizations.

eemeli:
This string is now getting special treatment in Pontoon so that it at least doesn't break the UI for localisers.

eemeli:
Just to let you know, under mozilla/pontoon#3538 we're changing how Pontoon serialises message files (in the next month or two).

If this message isn't fixed (e.g. by merging #3538), then the Arabic translation of this string will explode into a bit over 2000 lines. See here: https://gist.github.com/eemeli/e8ca98bd78ef8c92781afe5b9092f968

HarikalarKutusu:
Just to be clear:

Common Voice cannot work without this sentence, right?
Simply dividing the sentence into parts, or just removing it as a whole would break everything, right?
Or would it be misleading to just give a sample statement like given above "10,000 hours is achievable in just over 6 months if 1,000 people record 45 clips a day."?
[...]

PR #4811, opened 2025-02-28
Fix: Remove variables and revise wording in goals page which crashes in some languages

HarikalarKutusu:
Fixes: #3907 

In goals page, there is a complex structure which crashes the page in some languages because of unnecessarily added plurals/variables to the i10n sentence, which creates a thousand possibilities for Fluent. But the variables are not variables, just passed as constants.

Also the calculations are based on 10k hours what could be valid in 2018, but not nowadays. And it also depends on many other variables as explained here:
https://github.com/common-voice/common-voice/issues/3907#issuecomment-2690459311

So, this PR just makes it a single fixed value string with more recent requirements.

### DIFF
--- a/web/locales/common-voice/en/pages/dashboard.ftl
+++ b/web/locales/common-voice/en/pages/dashboard.ftl
@@ -52,20 +52,7 @@ help-reach-hours-general-pluralized =
     } in a language with a personal goal
 set-a-goal = Set a goal
 cant-decide = Can't decide?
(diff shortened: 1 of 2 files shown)


moz-dfeller:
Looks great, less complex. If you are okay with that @ftyers , I'm happy to merge it.

ftyers (approved):
Looks good to me!

PR #4842, opened 2025-03-24
Create 202503241812000-update-sentence-bands.ts

ftyers:
Update the sentence bands for new languages.


ftyers:
I updated the PR with the full list of languages.

PR #4843, opened 2025-03-26
Feat : Add less more to datasets page

HarikalarKutusu:
- This implements the FR https://github.com/common-voice/common-voice/issues/4303 
- It puts a "see more/less" button at the top of the table. We kept similar styling like in languages page, but used a contrasting gray.
- By default it shows the latest 6 records, when clicked it shows all records.
- It only works on desktop views. On smaller width screens (mobile) it is not displayed (hidden) and the whole list is shown. We assume people would not download datasets on phones - but might like to see what is available. For desktops where the windows width is small, one might just increase the width (see [related comment](https://github.com/common-voice/common-voice/issues/4529#issuecomment-2705427276) in another issue).

Checked on local dev environment (Win 11, WSL Ubuntu 22.04.x, Docker)

These are for different views:

**1080p screen full width**
image

**Medium width**
image

**Mobile widths**
image

### DIFF
--- a/web/locales/common-voice/en/pages/datasets.ftl
+++ b/web/locales/common-voice/en/pages/datasets.ftl
@@ -104,5 +104,5 @@ download-contribute-menu-tooltip = Download our voice datasets
 # MENU ITEM ARIA LABELS
 download-contribute-menu-aria-label = Download options for accessing Common Voice data
 
-datasets-show-more = See More
-datasets-show-less = See Less
+datasets-show-more = Show All Datasets
+datasets-show-less = Show Latest Datasets
--- a/web/src/components/pages/datasets/dataset-corpus-download-table.css
+++ b/web/src/components/pages/datasets/dataset-corpus-download-table.css
@@ -1,98 +1,110 @@
-
 .dataset-table {
+  width: 100%;
+  border-collapse: collapse;
+  @media (--md-down) {
+    &.table {
+      border-collapse: separate;
+      border-spacing: 0 1em;
+    }
+  }
+  thead,
+  tbody {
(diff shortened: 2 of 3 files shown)


ftyers:
Shouldn't the "see more" be at the bottom? It feels really backwards to me to have it at the top. Or change it to "Show all datasets" rather than "see more".

moz-bozden:
@ftyers, good points. Here are my chain of thoughts about these at that time:

### See more

I replicated "See More" in the languages page, which would prevent adding a new Pontoon key... But it can be changed of course (with counterpart "See latest datasets")...

image

### Location of the "See More"

At that time I tried both. When at the bottom, it gets shifted out of the screen, so, to "see less" one should scroll down a lot. And when the user presses it, this time the list gets shorter, and gets out of the screen, so the user must scroll-up this time...
IMO it is bad UX. It should work like an accordion. Maybe adding a down-chevron on the right will explain the idea (we also have it in mobile view to show the details)?

Like this example: https://getbootstrap.com/docs/5.0/components/accordion/#example

Whatever you choose... But the accordion example with chevron and new wording can be better I think...

ftyers:
I think this is ok (the chevron):

> IMO it is bad UX. It should work like an accordion. Maybe adding a down-chevron on the right will explain the idea (we also have it in mobile view to show the details)?

It could also be:

> Show older datasets

moz-bozden:
Here is how it looks/works now:

https://github.com/user-attachments/assets/e6825c29-3257-4dc4-b866-b2d8e28d54ab

ftyers:
Fantastic thanks! :D

PR #4847, opened 2025-03-26
Fix variant names in speak listen

HarikalarKutusu:
- Fixes sentence variant display in both SPEAK and LISTEN pages
- Instead of uppercase variant token it now displays a combination of `name (token)` as they come from database (no case manipulation).
- Fixes several reports:
  - https://github.com/common-voice/common-voice/issues/4790
  - https://github.com/common-voice/common-voice/issues/4840
  - https://github.com/common-voice/common-voice/issues/4806 (second part)

Example:

image

### DIFF
--- a/web/src/components/pages/contribution/contribution.tsx
+++ b/web/src/components/pages/contribution/contribution.tsx
@@ -450,7 +450,7 @@ class ContributionPage extends React.Component<ContributionPageProps, State> {
                           </div>
                         ) : null}
                         {sentence?.variant && (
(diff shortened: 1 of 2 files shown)


moz-kathyreid:
@HarikalarKutusu Bülent, I wanted to say a HUGE thank you for all of your efforts on this piece. I have booked some time to review when I'm on shift again, which is in about 6 hours' time. Really, this is a lot of work and we really appreciate it.

moz-rotimib (approved):

PR #4850, opened 2025-03-27
Feat text input cleaning

HarikalarKutusu:
This fixes several raw text related problems for new sentences and reports.
- Introduces simple central functions
- All free-form text input (sentences, sentence sources, reports' other options) are cleaned before entering the database.
- The `sentence_id` is calculated from the cleaned sentence

Notes/warnings:
- It does not handle any existing data. We still need cleaning in bundler for metadata `.tsv` files so they have correct structure (columns) coming from past malformed data.
- In dev environments, where sentences are imported from files, there might be cases where the `sentence_id`s / even sentences not match, because this time they will be cleaned of invisible characters like TAB or CR/LF (word separating characters).
- Some existing regex cleaning has been replaced by this, but the piped one (server/src/core/sentences/cleaning/multiple-sentences.ts) is untouched.

Limited testing on dev environment - checked from local database (Dockerized):
- Entering a single sentence with TAB chars in sentence and source => OK
- Entering a small batch sentences with (malformed sentence and source) => OK
- Reporting with "other", multiple lines and TAB characters => OK

### DIFF
--- a/server/src/application/text-cleaner.ts
+++ b/server/src/application/text-cleaner.ts
@@ -1,0 +1,10 @@
+export const cleanText = (s: string): string => {
+  return s.trim().replace(/\s/gi, ' ').replace(/\s+/g, ' ')
+}
+
+export const cleanTextArray = (ss: string[]): string[] => {
+  for (let i = 0; i < ss.length; i++) {
+    ss[i] = ss[i].trim().replace(/\s/gi, ' ').replace(/\s+/g, ' ')
+  }
+  return ss
+}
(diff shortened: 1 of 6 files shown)


server/src/application/text-cleaner.ts
```diff
@@ -0,0 +1,10 @@
+export const cleanText = (s: string): string => {
+  return s.trim().replace(/\s/gi, ' ').replace(/\s+/g, ' ')
+}
+
+export const cleanTextArray = (ss: string[]): string[] => {
```
moz-dfeller: We're not calling `cleanTextArray` anywhere in the code, so this can be removed.

server/src/application/text-cleaner.ts
```diff
@@ -0,0 +1,10 @@
+export const cleanText = (s: string): string => {
+  return s.trim().replace(/\s/gi, ' ').replace(/\s+/g, ' ')
```
moz-dfeller: Should be same as this:
```suggestion
  return s.trim().replace(/\s+/g, ' ')
```

PR #4854, opened 2025-03-28
Fix source field not shown in review

HarikalarKutusu:
Fix for https://github.com/common-voice/common-voice/issues/4580
Detailed explanation is [here](https://github.com/common-voice/common-voice/issues/4580#issuecomment-2761791979)


moz-dfeller:
Thanks @HarikalarKutusu 🙏

Issue #4750, opened 2025-01-14
Tagging variants for the Svan language texts


razmikb:
Hi, please assign variant tags to these sentences for the Svan (sva) language. TSV file https://drive.google.com/file/d/17ZzFTEE11bExK8VLB_ooppjmQZFzqbVx/view?usp=sharing

razmikb:
@jessicarose hi, any news on these task? it's been 2 weeks now.

moz-kathyreid:
## Approach for tackling this request 

These are my notes from discussing this request with @moz-dfeller (thank you, Dmitrij). 

### Migration file for updating the database 

* Adding the variant to the database should be written as a migration file in `.ts` format, similar to the migration files that are at: `server/src/lib/model/db/migrations`

### Documentation required

* We also need better documentation for how these migration files should be written to better enable other developers to write them in the future, with a particular focus on testing, because if a migration fails in staging, it is a lot of work to fix. Migrations need to be well-tested. 

### SQL specifics 

* The SQL statement should take the form: 

```
UPDATE sentence_metadata sm
JOIN sentences s ON sm.sentence_id = s.id
SET sm.variant_id = 
	(SELECT variant_id 
	 FROM variants v
	 WHERE v.variant_token = 'sva-lashkh'
	 )
WHERE s.text IN
('', '', '', '') <- this is where the sentence text would go 

```
_Note: there are four different variants given in the sentence file so there need to be four different SQL statements for the three variants - `sva-lashkh`, `sva-lentekh`, `sva-lowerbal` and `sva-upperbal`._

### Similar migration files to use as 

* Looking for similar migration files for sentence variants to use as a foundation: 
  - `20250221120100-set-empty-kbd-variants-to-cyrl.js`
  - `20250221120100-set-empty-kbd-variants-to-cyrl.js`

### Testing approach 

* Run the SQL sentences in isolation on the DB in development 
  - The sentence count for `sva-lashkh` is 73
  - The sentence count for `sva-lentekh` is 118
  - The sentence count for `sva-lowerbal` is 13
  - The sentence count for `sva-upperbal` is 50
  - The total count is 254
* Rebuild the docker container (so we have a fresh DB to run from) 
* Spin up the docker containers, which will run the migrations and fail if there are any errors 
* Add `sva` as a langauge to Profile, add variant 1 
* Check sentences that are ascribed to that variant show when contributing audio
* Repeat for variant 2 and 3

moz-kathyreid:
[...]

PR #4859, opened 2025-04-03
Add variant metadata for `sva` sentences per #4750

moz-kathyreid:
### Type of Pull Request

- [X] Add variant metadata for sentences already in CV

* #4750 

### Testing done 

* Individually ran each SQL query x 4 against a dev version of the db => all syntax correct
* Rebuild container and run `docker compose up` to validate `.ts` file is syntactically correct, _however_ it has no impact on my dev db because the `sva` sentences are not loaded there (were imported in bulk). So I can't guarantee that the update works _as intended_, just that the SQL itself is syntactically correct. 

Many thanks to @razmikb for doing the hard work of assigning each sentence to a variant in the spreadsheet.


moz-dfeller (changes requested):
Hey @moz-kathyreid, thanks for the migration. I tested it and it works somewhat. If you run each individual update statement separately it works but if you try to run all at once it crashes. This is basically what you did by putting all update statements into one query. You have to run each update statement as a separate query, i.e. 
```ts
export const up = async function (db: any): Promise<any> {
  await db.runSql('update one')
  await db.runSql('update two')
  ...
 }
 ```
 then it will work as expected 👍

moz-kathyreid:
Don't review this yet, it's still failing on `docker compose up` on a syntax issue, debugging ...

moz-kathyreid:
This is now ready for review. 

The way I approached this was: 

* Changed all of the single quotes to double quotes to align with our internal standards
* Ran `docker compose up` to test that there were no syntax errors -> there were, so I ... 
* For each block of SQL, I copied and pasted into a separate `.sql` file and ran a linter over the SQL file -> this found a syntax error in one of the blocks
* Ran `docker compose up` again to test the syntax error was resolved. 

I believe this can now be reviewed and merged, and apologies for the stuffing around, I should have been more thorough the first time.

moz-dfeller (approved):
Good job, ready to go 🚀

Issue #4721, opened 2024-12-13
[BUG] Bad source for variant name translations (should get from database)


HarikalarKutusu:
In "write" functionality, new "sentence-variants" feature is enabled. On the other hand the variant names are not shown, only variant bcp-47 codes are shown, because the code tries to look them up from the `.ftl` (translation) files. As they are only intelligible for that language, the team correctly decided to put these names into the database (otherwise we would have a thousand terms for each locale to translate - without any use).

So, it should actually get the "translations" from the database, and should not try to use l10n/localization.

See:
![image](https://github.com/user-attachments/assets/fca7497d-7d76-47e4-bb85-ab69b6526639)

![image](https://github.com/user-attachments/assets/09ca33ee-b6fa-461d-8034-ad012373ac50)

As you might guess, bcp-47 codes are not very intelligible for an average user.

HarikalarKutusu:
Fixed by https://github.com/common-voice/common-voice/pull/4865
Closing

PR #4865, opened 2025-04-09
Use "variant name [variant tag]" format everywhere (review, write, profile)

HarikalarKutusu:
This PR fixes/implements:
- https://github.com/common-voice/common-voice/issues/4840 (this time for real).
- #4721
- Also removes translation warnings in the console.

- [X] TO-CHECK: Effect of variant related settings in the profile => Added a commit to handle that case (see note below)
- [x] TO-DO: Better layout and test with different length sentences => seems OK

@moz-dfeller & @moz-rotimib :
- This might need a bit more refactoring afterwards. I think the variant display interface can be made better (e.g. with hooks or passing the Variant[] as a whole), so that it can be included from a single place. Waiting for comments.
- In WRITE, when selection completed, only the tag is shown in the form field. The `Select` component interface is a bit limiting, but can be fixed in refactoring.

image

Large screen with long textual parts test (seems OK):
image

Same on smaller screens (seems OK):
image

Profile:
image

Write:
image


moz-dfeller (changes requested):
Hey @HarikalarKutusu, looks good for me from the backend perspective. There are FE tests that are failing though and have to be fixed and one merge conflict (due to the recent random sentences update).

HarikalarKutusu:
Missed one test :( `yarn test` gives too much output...

@moz-rotimib, and comment/request is very much welcome.

web/src/components/pages/contribution/sentence-collector/write/sentence-input-and-rules/sentence-input-and-rules.tsx
```diff
@@ -46,15 +47,21 @@ export const SentenceInputAndRules: React.FC<Props> = ({
...
   const isSentenceError =
     error && error?.type !== SentenceSubmissionError.NO_CITATION
   const isCitationError = error?.type === SentenceSubmissionError.NO_CITATION
   const hasVariants = variantTokens && variantTokens.length > 0
-
-  const { l10n } = useLocalization()
+  const variantOptions: string[] = [l10n.getString('sentence-variant-select-multiple-variants')]
+  if (hasVariants) {
+    for (let i=0; i < variantTokens.length; i++) {
+      variantOptions.push(`${variantNames[i]} [${variantTokens[i]}]`)
```
moz-rotimib, line 62: Could you use Array.map here instead? You can then spread the result from the map into the variantOptions array. Generally, we prefer to use Array functions that `map` that don't mutate variables unlike `push` which does.

For example,

```js
const resultFromMap = variantTokens.map(...) // example variable name feel free to use one that makes sense to you

const variantOptions = [l10n.getString('sentence-variant-select-multiple-variants'), ...resultFromMap]
```

web/src/components/select/index.tsx
```diff
@@ -53,7 +55,7 @@ export const Select: React.FC<Props> = ({
         {...getToggleButtonProps()}
         data-testid="select-toggle-btn">
         <span>
-          {selectedItem ? l10n.getString(selectedItem) : placeHolderText}
+          {selectedItem ? (doTranslation ? l10n.getString(selectedItem) : selectedItem) : placeHolderText}
```
moz-rotimib, line 58: You can refactor this to use if/else. Nested ternaries are hard to understand

HarikalarKutusu: It was in the JSX thou, I took the logic out for this.

HarikalarKutusu: Done

moz-rotimib (approved):
LGTM

moz-dfeller (approved):

Issue #4708, opened 2024-12-04
[BUG] multiple issues with docker local development


NiziL:
**Describe the bug**
Cannot start the app on a fresh clone

**To Reproduce**
1. Clone the repository
2. Follow the DEVELOPMENT.md tutorial

**Expected behavior**
Well... I expect it to start :clown_face: 

**Desktop or Mobile (please complete the following information):**
 - OS: Ubuntu 24.04
 - Browser: Firefox 133.0

**Additional context**
So, I've tried to start a local instance of `common-voice`, here is the issues I've been confronted to

- Permission error
  - a permission denied, solved by exporting `UID` and `GID` as specified in #4003 
- Authentication errors
  - a TypeError `Cannot read properties of null` (at [auth-router.ts](https://github.com/common-voice/common-voice/blob/main/server/src/auth-router.ts#L33)) if I don't set up the authentication, while the documentation tells me I could start the app without it
  - the doc is outdated, variables are no longer prefixed by `CV_AUTH0_`, but by `CV_FXA_`. I had to take a look at `server/src/config-helper.ts` to learn this
- DB Migration error
  - My `locales` table is not properly populated, there is only `en` and `unknown`. This leads to SQL errors `Column 'locale_id' cannot be null` on all `update-sentence-bands` files, as there is no `ady`, `kdb` and so on in the `locales` table.

Any help is greatly appreciated :)

HarikalarKutusu:
I'm now trying `docker compose` workflow on Windows 11 through WSL2. My previous dev environment was fully on Windows/bare-metal. I had to solve lots of stuff wrt WSL of course.

> Permission error

I think #4801 solved it, at least in my case.

> Authentication errors

I spent a lot of time to understand why database tables are not created and there is no web server/page. I could only pass that stage when I commented out related line in ``server.ts, which of course caused further problems.

```
// app.use(await setupAuthRouter())
```
I've been using Auth0 in my old setup, but now I have to figure out how Mozilla Accounts work. It is definitely necessary to include that info in the `DEVELOPMENT.md` file.

@NiziL, if you can share your solution (even PR) it will help.

> DB Migration error

- After the comment out, the database tables could be created, migrations applied and my only locale (tr) could be imported. I had one deadlock error.

```
web  | [BE] error when inserting sentence batch from " Bin yıldır Allah'ın emri dendi kadınlara. " to "  ": Error: Deadlock found when trying to get lock; try restarting transaction
[...]

PR #4885, opened 2025-05-01
[FIX] 4708 update AUTH to FXA prefix in env / docs

SallyMcGrath:
### Type of Pull Request

- [ ] Bulk sentence upload 

* #4708
*

- [ ] Other

Fixes _part_ of #4708

In the setup docs it instructs the developer to create an auth application with Auth0 and add the credentials to the env file for docker. 

However in #4684 Auth0 was replaced throughout the application with Mozilla Accounts (FXA) so this cannot work and the site won't run in dev. 

I looked around a bit and while I eventually found this https://mozilla.github.io/ecosystem-platform/relying-parties/tutorials/integration-with-fxa it looks like a mountain to climb as a first step in onboarding! (Especially compared to the absolute breeze of creating an Auth0 app. ) 

I'm sure there's a simple way to get an FXA detailed somewhere. For now, I've just edited the docs to call the auth0 creds FXA. Can I suggest a subsequent ticket to write a step by step for new users on how they are to acquire their creds. 

*
*

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -230,10 +230,12 @@ If you want to work with login-related features (Profile, Dashboard, Goals, ...)
 For Docker, in `.env-local-docker`:
 
 ```Dotenv
-CV_AUTH0_DOMAIN="<domain_here>"
-CV_AUTH0_CLIENT_ID="<client_id_here>"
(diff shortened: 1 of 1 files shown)


moz-kathyreid:
Flagging that this is related to #4973 and should be tackled all at once

Issue #4845, opened 2025-03-26
[BUG]


ffsdmad:
## Describe the bug
Docker containers, including the bundler and web containers, fail to start correctly when running the `docker-compose up` command.

## To Reproduce
1. Run `git clone https://github.com/common-voice/common-voice/`
2. Navigate to the project directory: `cd common-voice`
3. Copy the environment file: `cp .env-local-docker.example .env-local-docker`
4. Run `docker-compose up`

## Expected behavior
I expected all containers, including bundler and web containers, to start and function correctly.

## Screenshots
Screenshots will be provided later.

## Desktop or Mobile (please complete the following information):
- **OS**: Ubuntu 22.04.5 LTS
- **Browser**: Not applicable (this is a backend issue with Docker)
- **Version**: Ubuntu 22.04, Kernel version 6.5.0-44-generic

## Additional Hardware (were you using headphones, an external speaker or an external microphone?):
- None

## Additional context
Here are the logs for the affected containers:

**bundler container log:**
bundler  | npm error code EACCES
bundler  | npm error syscall mkdir
bundler  | npm error path /home/node/code/node_modules
bundler  | npm error errno -13
bundler  | npm error
bundler  | npm error Your cache folder contains root-owned files, due to a bug in
bundler  | npm error previous versions of npm which has since been addressed.
bundler  | npm error
bundler  | npm error To permanently fix this problem, please run:
bundler  | npm error   sudo chown -R 10001:10001 "/home/node"
bundler  | npm error A complete log of this run can be found in: /home/node/_logs/2025-03-26T16_22_24_333Z-debug-0.log
bundler exited with code 243

**web container log:**
web      | warning lru.min@1.1.1: The engine "bun" appears to be invalid.
web      | warning lru.min@1.1.1: The engine "deno" appears to be invalid.
web      | [4/5] Linking dependencies...
web      | warning "workspace-aggregator-447a28d2-afb2-4ac2-b900-3c4e63e9e277 > server > db-migrate-plugin-typescript@2.0.0" has incorrect peer dependency "ts-node@^3.3.0".
web      | warning "workspace-aggregator-447a28d2-afb2-4ac2-b900-3c4e63e9e277 > web > react-balance-text@2.0.1" has incorrect peer dependency "react@^0.14.7 || ^15.0.0 || ^16.0.0-0".
[...]

Issue #4389, opened 2024-03-08
[BUG] Docker-compose up -d --build has problem with bundler


cod3r0k:
I am having an issue when trying to set up Common Voice using Docker Compose.

here is `f7bbcf4aa4ef   common-voice_bundler        0.0.0.0:9001->9001/tcp, :::9001->9001/tcp` logs:

```
npm ERR! code EACCES
npm ERR! syscall mkdir
npm ERR! path /home/node/code/node_modules
npm ERR! errno -13
npm ERR!
npm ERR! Your cache folder contains root-owned files, due to a bug in
npm ERR! previous versions of npm which has since been addressed.
npm ERR!
npm ERR! To permanently fix this problem, please run:
npm ERR!   sudo chown -R 10001:10001 "/home/node"

npm ERR! A complete log of this run can be found in: /home/node/_logs/2024-03-08T16_41_55_510Z-debug-0.log
npm ERR! code EACCES
npm ERR! syscall mkdir
npm ERR! path /home/node/code/node_modules
npm ERR! errno -13
npm ERR!
npm ERR! Your cache folder contains root-owned files, due to a bug in
npm ERR! previous versions of npm which has since been addressed.
npm ERR!
npm ERR! To permanently fix this problem, please run:
npm ERR!   sudo chown -R 10001:10001 "/home/node"

```

It never run and exit

raivisdejus:
I was able to solve permission denied issues with these commands on Ubuntu 24.04:
```
export UID=$(id -u)
export GID=$(id -g)
docker compose up --force-recreate web
```

moz-bozden:
This has been fixed in Docker file and in DEVELOPMENT.md...
Closing...

Issue #4003, opened 2023-04-15
[BUG] Development - Can't start docker container


MichaelKohler:
**Describe the bug**
I tried to set up CV locally from a fresh setup on Fedora. Building the docker container works, however when it comes to running it, the command (from `docker-compose.yml`) `bash -c "/code/docker/wait-for-it.sh s3proxy:80 -- /code/docker/prepare_s3.sh && yarn && yarn start"` fails due to:

```
~/development/common-voice main ⇡ ❯ docker-compose up -d && docker logs -f web
redis is up-to-date
s3proxy is up-to-date
db is up-to-date
Recreating web ... done
bash: /code/docker/wait-for-it.sh: Permission denied
```

The issue here seems to be that, when mounted through the volume in docker-compose.yml, the owner user/group is wrong:

```
drwxr-xr-x.   1 root root  160 Apr 15 17:26 .
drwxr-xr-x.   1 root root  160 Apr 15 17:26 ..
-rwxr-xr-x.   1 root root    0 Apr 15 17:26 .dockerenv
drwxr-xr-x.   1 root root  948 Apr 20  2022 bin
drwxr-xr-x.   1 root root    0 Jul 10  2020 boot
drwxr-xr-x.   1 node node  670 Apr 15 08:00 code
drwxr-xr-x.   5 root root  340 Apr 15 17:26 dev
drwxr-xr-x.   1 root root 1848 Apr 15 17:26 etc
drwxr-xr-x.   1 root root    8 Apr 20  2022 home
drwxr-xr-x.   1 root root   90 Apr 20  2022 lib
drwxr-xr-x.   1 root root   40 Apr 18  2022 lib64
drwxr-xr-x.   1 root root    0 Apr 18  2022 media
drwxr-xr-x.   1 root root    0 Apr 18  2022 mnt
drwxr-xr-x.   1 root root   26 Apr 20  2022 opt
dr-xr-xr-x. 483 root root    0 Apr 15 17:26 proc
drwx------.   1 root root   42 Apr 20  2022 root
drwxr-xr-x.   1 root root   16 Apr 18  2022 run
drwxr-xr-x.   1 root root 1084 Apr 20  2022 sbin
drwxr-xr-x.   1 root root    0 Apr 18  2022 srv
dr-xr-xr-x.  13 root root    0 Apr 15 07:56 sys
drwxrwxrwt.   1 root root  954 Apr 15 17:20 tmp
[...]

PR #4896, opened 2025-05-15
Update config-helper to ignore non-existent config.json

moz-kathyreid:
### Type of Pull Request

- [X] Amendment to logic which loads configuration

* #4845
* #4708 
* #4389 
* #4003

This PR: 

* Changes the logic in the `configuration-helper.ts` and no longer checks for a `config.json` file, which was causing confusion with people setting up their own development environment, as it threw an error if it wasn't found, _even though_ it wasn't required. 
* Removes references to `loadedConfig` which no longer serves a purpose AFAICT
* Adds better semantic logging of what's happening through `console.log`
* Provides better failure message if dotenv file not found 

I do have some questions about things that I don't understand in this file: 

* I don't understand how `injectedConfig` works, or what purpose it serves? 
* With the logging I added, I can see that the config gets loaded _a lot_. I couldn't see an easy way to log _what_ was being loaded from the config at a particular stage via `getConfig()`, but that would be helpful here.

### DIFF
--- a/server/src/config-helper.ts
+++ b/server/src/config-helper.ts
@@ -6,7 +6,15 @@ if (process.env.DOTENV_CONFIG_PATH) {
   const result = config({ path: process.env.DOTENV_CONFIG_PATH })
   if (result.error) {
     console.log(result.error)
-    console.log('Failed loading dotenv file, using defaults')
+    console.log(
+      'ERROR loading config: Failed loading dotenv file. Expected file: ',
+      process.env.DOTENV_CONFIG_PATH
+    )
+  } else {
+    console.log(
+      'Loading config: successfully loaded dotenv file: ',
+      process.env.DOTENV_CONFIG_PATH
+    )
   }
 }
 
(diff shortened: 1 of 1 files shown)


server/src/config-helper.ts
```diff
@@ -122,32 +130,25 @@ const BASE_CONFIG: CommonVoiceConfig = {
...
-    fileConfig = JSON.parse(fs.readFileSync(config_path, 'utf-8'))
+    if (injectedConfig) {
+      console.log('Loading config: reading injectedConfig ...')
+      return injectedConfig
+    }
+
+    if (BASE_CONFIG) {
+      console.log('Loading config: reading BASE_CONFIG ...')
+      return BASE_CONFIG
+    }
```
moz-dfeller, line 148: There is no need to check whether `BASE_CONFIG` exists, because it's always set either with default values or environment variables. So we can we can shorten it here to:
```suggestion
    console.log('Loading config: reading BASE_CONFIG ...')
    return BASE_CONFIG
```

moz-dfeller (changes requested):
Looks good! Thanks for adding better logging 🚀 There's just an improvement we can do here.

moz-dfeller:
> I do have some questions about things that I don't understand in this file:
>* I don't understand how injectedConfig works, or what purpose it serves?
>* With the logging I added, I can see that the config gets loaded a lot. I couldn't see an easy way to log what was being loaded from the config at a particular stage via getConfig(), but that would be helpful here.

So regarding `injectedConfig`. It's used in tests to overwrite the actual config to provide alternative values during testing.

The config gets called a lot because it provides a lot information that is required at every step, e.g. the environment alone (i.e. sandbox, stage, prod) is probably needed in a lot of files. We also don't really want to log what gets called because we don't want to pop secrets into logs that are sent around the globe :)

moz-kathyreid:
Confirming requested changes have been made re `BASE_CONFIG`, ready for re-review

moz-dfeller (approved):
Thanks for the much needed update 👍

Issue #4904, opened 2025-05-22
[TASK] Add Australian English variant `en-AU` to English


moz-kathyreid:
## Request details 

Australian English en-AU Australian English has unique lexical and phonetic features as a variant of English. For example, many place names are derived from Indigenous Australian languages - e.g. Geelong (Djilang) from Waddawurrung, Wodonga from Wiradjuri, Wagga Wagga from Wiradjuri and so on. 

Many slang expressions shorten words - e.g. "I'm heading off this arvo to go to the bottlo to get some bevvies for Tommos" (I am heading off this afternoon to go to the off-license shop to purchase alcohol to take to Mr Thompson's). 

Phonetically, Australian English intonation rises at the end of a sentence so many statements sound like questions. There are also phonetic variations *within* Australian English which are factors of geography and also internal demography - there are variations for example based on migration patterns. Adelaide tends to sound more like British English, while in Western Sydney we have a crucible of phonetic patterns from migration - think something similar to MLE emerging in London. 

This request is to have en-AU added as a Variant of English in Common Voice.

ftyers:
There are curently no variants for English, but there is a pre-set accent for Australian English:

```
  en: {
    us: 'United States English',
    australia: 'Australian English',
    england: 'England English',
    canada: 'Canadian English',
    philippines: 'Filipino',
    hongkong: 'Hong Kong English',
    indian: 'India and South Asia (India, Pakistan, Sri Lanka)',
    ireland: 'Irish English',
    malaysia: 'Malaysian English',
    newzealand: 'New Zealand English',
    scotland: 'Scottish English',
    singapore: 'Singaporean English',
    southatlandtic: 'South Atlantic (Falkland Islands, Saint Helena)',
    african: 'Southern African (South Africa, Zimbabwe, Namibia)',
    wales: 'Welsh English',
    bermuda: 'West Indies and Bermuda (Bahamas, Bermuda, Jamaica, Trinidad)',
  },
```

I don't know if we want to implement variants for English, and if so, what they should be. But imo, this should be done holistically and not piecemeal.

moz-kathyreid:
Excellent point, @ftyers. 

I've been pondering on this all week and have arrived at the following: 

* Given that English is spoken by approx 1.2 billion people on the planet
* And is one of our largest language datasets 
[...]

PR #4905, opened 2025-05-22
Add en-AU variant

moz-kathyreid:
### Type of Pull Request

- [ ] Add a Variant for an existing locale

* #4904 

### Testing done 

* Check that any locale-specific regex don't need to be altered to support the Variant => 
checked `https://github.com/common-voice/common-voice/blob/main/server/src/core/sentences/validation/languages/en.ts` => no rules require changing for `en-AU`
* Check that only the altered files are committted => OK 
* Check that the Variant is added to the DB => OK

```
SELECT l.name, l.native_name, v.variant_name, v.variant_token 
FROM variants v, locales l
WHERE v.locale_id = l.id 
AND l.name = 'en'
```

### DIFF
--- a/server/src/lib/model/db/language-data/variants.ts
+++ b/server/src/lib/model/db/language-data/variants.ts
@@ -481,4 +481,9 @@ export const VARIANTS: Variant[] = [
     variant_name: 'Xopuri',
     variant_token: 'lzz-xopuri',
   },
+  {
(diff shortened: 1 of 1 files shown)


moz-kathyreid:
PLEASE HOLD THIS PR, we will likely take a different approach, see linked Issue.

PR #4920, opened 2025-06-05
Fix where `client_id` is returning `null` and throwing a `TypeError`

moz-kathyreid:
[WIP do not merge I will update details]

### Type of Pull Request

- [ ] Bulk sentence upload 

*
*

- [ ] Other

*
*

*
*

### DIFF
--- a/server/src/lib/api.ts
+++ b/server/src/lib/api.ts
@@ -609,7 +609,10 @@ export default class API {
 
   getContributionActivity = async (req: Request, response: Response) => {
     const { locale } = req.params
-    const { client_id } = req?.session?.user
+    const { client_id } = req?.session?.user || {}
+    if (!client_id) {
+      return response.sendStatus(StatusCodes.BAD_REQUEST)
+    }
     const { from } = req.query
 
     response.json(


server/src/lib/api.ts
```diff
@@ -271,10 +271,10 @@ export default class API {
...
+    // default the client_id to null if it is not present in the session.user object
+    // so it does not raise a TypeError
     const {
-      session: {
-        user: { client_id },
-      },
+      session: { user: { client_id = null } = {} } = {},
       params: { id },
     } = request
     await this.model.db.createSkippedSentence(id, client_id)
```
moz-dfeller, line 283: Throw 400 here, since we need the client_id to create a meaningful entry

server/src/lib/api.ts
```diff
@@ -609,7 +609,10 @@ export default class API {
 
   getContributionActivity = async (req: Request, response: Response) => {
     const { locale } = req.params
-    const { client_id } = req?.session?.user
+    const { client_id } = req?.session?.user || {}
+    if (!client_id) {
+      return response.sendStatus(StatusCodes.BAD_REQUEST)
```
moz-dfeller: No need to throw here, since we can actually serve this request without a client_id

server/src/lib/clip.ts
```diff
@@ -193,12 +193,10 @@ export default class Clip {
...
-      session: {
-        user: { client_id },
-      },
-      headers,
-    } = request
+    // default the client_id to null if it is not present in the session.user object
+    // so that a 400 is returned below
+    const { session: { user: { client_id = null } = {} } = {}, headers } =
+      request || {}
     const sentenceId = headers.sentence_id as string
```
moz-dfeller, line 200:
```suggestion
    const client_id = request?.session?.user?.client_id
    const sentenceId = headers.sentence_id as string
```

server/src/lib/clip.ts
```diff
@@ -193,12 +193,10 @@ export default class Clip {
...
-    const {
-      session: {
-        user: { client_id },
-      },
-      headers,
-    } = request
+    // default the client_id to null if it is not present in the session.user object
+    // so that a 400 is returned below
+    const { session: { user: { client_id = null } = {} } = {}, headers } =
+      request || {}
```
moz-dfeller:
```suggestion
    const { headers } = request 
```

server/src/lib/api.ts
moz-kathyreid: Confirming this is addressed in commit https://github.com/common-voice/common-voice/pull/4920/commits/38fcf9d1cdd1cba7f0f97aef40108c560363314a

server/src/lib/api.ts
moz-kathyreid: Confirming this is addressed in commit https://github.com/common-voice/common-voice/pull/4920/commits/2dd486c854c2a90281a5750803ed2197f8c859ba

moz-kathyreid:
Confirming I have addressed all the review comments, and the additional one we chatted about just now, @moz-dfeller. Now sending for review again.

moz-dfeller (approved):

PR #4931, opened 2025-06-11
1157- Backend to retrieve sentences & clips without variant when variant ones run out

moz-bozden:
When a user sets a variant and "show only sentences/clips from my variant", the system shows "you are out of sentences/clips".

We want to give an option to the user to continue with sentences/clips from any variant (including those without any variant).

This PR handles the backend of this with an `ignoreClientVariant=true` query parameter (default false) which should be called from front-end after a button click.

### DIFF
--- a/server/src/lib/api.ts
+++ b/server/src/lib/api.ts
@@ -192,6 +192,7 @@ export default class API {
 
     // the validator coerces count into a number but doesn't update the type
     const count: number = (request.query.count as never) || 1
+    const ignoreVariant: boolean = Boolean(request.query.ignoreVariant) || false
 
     const userClientVariant = await pipe(
       client_id,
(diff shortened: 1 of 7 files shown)


server/src/lib/api.ts
```diff
@@ -192,6 +192,7 @@ export default class API {
 
     // the validator coerces count into a number but doesn't update the type
     const count: number = (request.query.count as never) || 1
+    const ignoreVariant: boolean = Boolean(request.query.ignoreVariant) || false
```
moz-dfeller: This should be probably called `ignoreClientVariant` as we don't ignore sentences with variants per se, but rather ignore if the client wants to see only variants that they have set in their profile
```suggestion
    const ignoreVariant: boolean = Boolean(request.query.ignoreClientVariant) || false
```

server/src/lib/validation/sentences.ts
moz-dfeller, line 1: Good!

server/src/lib/api.ts
moz-bozden: I agree. I wanted to shorten it, but lost from context.

moz-bozden: Went ahead and also renamed the variables for completeness/clarity.

PR #4933, opened 2025-06-11
4915-fix-small-tag-in-the-upper-menu-dropdowns

moz-rotimib:
- Use regex to remove <small>...</small> text
- Update contribute menu styles and add review questions link


web/src/components/layout/nav/contribute-menu/contribute-menu.css
```diff
@@ -188,6 +189,15 @@
             width: auto;
         }
 
+        p.coming-soon-text {
```
moz-dfeller, line 192: This should not be in this PR I assume?

moz-rotimib: Answered on Slack but also mentioning here for visibility:

>There's a different type of coming soon that we have - for locales that are not yet contributable we show coming soon for the speak and listen links till they are contributable
When they are contributable, the normal text shows

moz-dfeller (approved):

PR #4935, opened 2025-06-13
1166-CC Fix setuptools version to fix bundler problems

moz-bozden:
Simple fix for bundler not working in `docker compose up`, preventing developers and also releases.
This has also been reported here: https://github.com/common-voice/CorporaCreator/issues/129


moz-dfeller (approved):
Thanks @moz-bozden 🚀

PR #4949, opened 2025-06-24
1177-MCV-BE Fix lazycache long lock issue

moz-bozden:
[Internal issue](https://github.com/common-voice/Internal-Common-Voice/issues/1177)
[Internal main issue](https://github.com/common-voice/Internal-Common-Voice/issues/1176)

The Redis based lazyCache implementation had a very long time (3 hours fixed), which prevented more quick updates.
As part of re-visiting the Redis caching in general to optimize query performance by using caching, we first fix this issue.

This adds a new lockDurationMs parameter to lazyCache, which defaults to 3 minutes. Also added some plausible lock duration to the calling functions, to be fine-tuned after fine examination of query performances.


server/src/lib/model.ts
```diff
@@ -144,37 +141,42 @@ export default class Model {
       const languages = await this.db.getAllLanguages()
       return languages
     },
-    DAY
+    TimeUnits.DAY,
+    5 * TimeUnits.SECOND,
```
moz-dfeller: I would not touch the 3 minute default in many cases, especially as 5 seconds locking time is too small for some endpoints that need more than 10s to resolve. This way you would just spawn new requests for the same data before the initial request has even finished, leading to wasted calculations.

moz-dfeller (changes requested):

server/src/lib/model.ts
moz-dfeller: More than 3 min is OK for known problem areas, but less than 3 min is probably not worth the potential wasted resource consumption

moz-bozden: I only dropped below 3 min when it was obvious (some of them this one I tested on local), but you are right, there might be a temporary congestion resulting in delays. I'll up them to 3 min.

moz-bozden: How about this one :) This 1 minute already seemed too low for me... Up it to 3 min?
What do you think?

```
  async findClipsWithFewVotes(
    client_id: string,
    locale_id: number,
    count: number,
    exemptFromSSRL?: boolean
  ): Promise<DBClip[]> {
    // get cached clips for given language
    const cachedClips: DBClip[] = await lazyCache(
      `new-clips-per-language-${locale_id}`,
      async () => {
        return await this.getClipsToBeValidated(locale_id, 10000)
      },
      TimeUnits.MINUTE,
      3 * TimeUnits.MINUTE
    )()
```

moz-dfeller: 3 min is the default so you can just omit the parameter here, i.e.
```ts
  async findClipsWithFewVotes(
    client_id: string,
    locale_id: number,
    count: number,
    exemptFromSSRL?: boolean
  ): Promise<DBClip[]> {
    // get cached clips for given language
    const cachedClips: DBClip[] = await lazyCache(
      `new-clips-per-language-${locale_id}`,
      async () => {
        return await this.getClipsToBeValidated(locale_id, 10000)
      },
      TimeUnits.MINUTE
    )()
```

moz-bozden: I want to be explicit on these, in case somebody changes the default.
What I mean is:

- CacheTime is 1 minute
- LockTime is 3 minutes

for this particular query. Isn't 1 minute already very small?
This is per-user hit per-locale...

moz-dfeller: I see, keeping it explicit is fine. Cache time is low to not have the same clips presented during community events with a couple hundred people participating at the same time.

moz-dfeller (approved):

PR #4950, opened 2025-06-24
1178-MCV-BE Initial lazySetCache implementation

moz-bozden:
[Internal issue](https://github.com/common-voice/Internal-Common-Voice/issues/1178)
[Internal main issue](https://github.com/common-voice/Internal-Common-Voice/issues/1176)

The Redis based lazy cache implementation for sets (numbers and strings), more performant and exposes more functionality.

- This module is currently as-is, without higher level connections for now.
- While integrating, we might need some changes, so consider this WIP.
- After integration to db & model temporary functions in the old lazyCache module will be removed.

### DIFF
--- a/server/src/lib/lazy-set-cache.ts
+++ b/server/src/lib/lazy-set-cache.ts
@@ -1,105 +1,39 @@
 import { redis, redlock } from './redis'
 import { TimeUnits } from 'common'
 
-//
-// Redis set based caching module for simple lists of numbers and integers
-// - More performant, quicker and low memory footprint, so that cache size can be bigger
-// - Adds more functionality and exposes them for application has more freedom
-// - Best use cases:
-//   - SELECT queries which return id fields only (locale_id, sentence_id, etc)
-//   - User skips, user reports, locales needing validation, locales having unrecorded sentences, etc
-//
+/*
+Redis set based caching class for simple lists of numbers and integers
+- More performant/quicker and low memory footprint, so that cache sizes can be bigger
+- Adds more functionality and exposes them for applications so that they can have more freedom
+- Best use cases:
+  - SELECT queries which return id fields only (locale_id, sentence_id, etc)
+  - User skips, user reports, locales needing validation, locales having unrecorded sentences, etc
+
+Example usage:
+
+import { LazySetCache } from './lazy-set-cache'
+import { TimeUnits } from 'common'
+
+async function getAllClipIdsUserSkipped(user_id: number): Promise<any[]> {
+  return await LazySetCache.use(
+    `get-all-skipped-clip-ids:${user_id}`,
+    async () => {
+      return await this.db.getAllClipIdsUserSkipped(user_id)
+    },
+    TimeUnits.DAY
+  )
+}
+
+// E.g. can be called just after sentence additions
+await LazySetCache.addSingleWithExpiry('get-all-language-ids-with-unvalidated-sentences', newId, TimeUnits.HOUR)
+
+// Generic
+await LazySetCache.addManyWithExpiry('keep-alive-xyz', new_list, TimeUnits.HOUR)
+
+*/
 
 function msToSeconds(ms: number): number {
   return Math.floor(ms / 1000)
-}
-
-// Add one item (number|string) to a Redis SET with TTL (resets TTL)
-export async function lazySetCacheAddWithExpiry(
-  key: string,
-  value: number | string,
-  cacheDurationMs: number
-): Promise<void> {
-  await redis.sadd(key, String(value))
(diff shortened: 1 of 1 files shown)


server/src/lib/lazy-set-cache.ts
```diff
@@ -0,0 +1,140 @@
...
+  values: (number | string)[],
+  cacheDurationMs: number
+): Promise<void> {
+  const pipeline = redis.multi()
+  pipeline.del(key)
+  if (values.length > 0) {
+    pipeline.sadd(key, ...values.map(String))
+  }
+  pipeline.expire(key, msToSeconds(cacheDurationMs))
+  await pipeline.exec()
```
moz-dfeller:
```suggestion
  if (values.length === 0) return
  const pipeline = redis.multi()
  pipeline.del(key)
  pipeline.sadd(key, ...values.map(String))
  pipeline.expire(key, msToSeconds(cacheDurationMs))
  await pipeline.exec()
```

server/src/lib/lazy-set-cache.ts
moz-dfeller, line 1: I would probably remove the `lazySetCache` prefix from the functions as you're importing the functions from the `lazy-set-cache` module/file. The caller can either import all or some functions, e.g.
```ts
import * as LazySetCache from './lazy-set-cache.ts'
import { fill } from './lazy-set-cache.ts'
 
LazySetCache.addWithExpiry('key', 'cool', TimeUnits.DAY)
fill('key', ['cool'], TimeUnits.DAY)
// Compared to 
// LazySetCache.lazySetCacheAddWithExpiry('key', 'cool', TimeUnits.DAY)
```
The user importing the `lazy-set-cache` file should understand that the module deals with sets, there shouldn't be a need to explicitly add the prefix to every function. If users are just importing some function, it can become hard to understand, e.g. `fill('key', ['cool'], TimeUnits.DAY)` is not really clear what that is. In this case, you could use
```ts
export const LazySetCache = {
  async function fill(
    key: string,
    values: (number | string)[],
    cacheDurationMs: number
  ): Promise<void> {...}
}
```
so that others have to import `import { LazySetCache } from './lazy-set-cache.ts'` and call `LazySetCache.fill()`

moz-dfeller (changes requested):

server/src/lib/lazy-set-cache.ts
moz-bozden: Thanks for the guard and bad expire reset.

Let us figure out the following case which I was trying to handle - it seems my logic was flawed:

- We expose this and some logic in upper layer decides to forced cache re-fill through this before expiry (e.g. other user's actions changed the DB state, and this user's "unvalidated-sentence-id:client-id" must be refilled forgetting old ones.
- But there are none found (values.length === 0)
- From the previous cache refill we have cache members (until expiry) and if the cache is directly used (regular lazySetCache call) they will still be presented to the user (e.g. causing unnecessary voting in the example).

I think we still need to set the cache to an empty state here - but then after a lazySetCache call, because it is empty, it will again try to refill it - and again it will be empty... That would cause a loop and unnecessary DB access (sometimes costly).

One solution would be NOT to expose this function and only use it through `lazySetCache` and control the whole "empty cache" state there, e.g. by checking the expire and not try to refill until it expires (currently it tries to refill on each call, e.g. after page refresh).

What do you think?

server/src/lib/lazy-set-cache.ts
moz-bozden: > import { LazySetCache } from './lazy-set-cache.ts' 
> LazySetCache.fill()

This method will be most useful , thank you!

server/src/lib/lazy-set-cache.ts
moz-bozden: No, I think the original code is still safer. When the application or main function decides to "fill" and the result set it empty:
- Old values will be reset to empty set (forgotten)
- We will have a new expiry

Now, assume a case with cache-duration = 10 min, lock-duration = 1 min. That would result in 8 additional DB hits...
1. This can be considered as a good thing, in the meantime new ones can become valid and begin to be served - but if nothing is happening on DB, the DB hits will be in vain.
2. We can be adaptive (which I was thinking for adaptive-cache) - We can start with 1 min lock, if still locked double the lock duration to 2, then 4, 8 - up to max = cache-duration (for idle on FE - e.g. nobody adding sentences, but keep trying validate). This would require "key" + "key-lock-dur" (int)...

What do you think?
I propose let's keep it as it is (I converted it to a class) and deal with it later when actually connecting them. This implementation has no connections for now...

server/src/lib/lazy-set-cache.ts
moz-bozden: Converted to a class with methods...

PR #4966, opened 2025-07-03
1187-MCV Change clip-upload process to handle remaining duplicate-clip errors

moz-bozden:
Although we fixed the main reason here (https://github.com/common-voice/common-voice/pull/4944) we still have some remaining errors. This PR changes the upload workflow to handle these, while we can continue to find the real reason:

- Instead of showing retry/cancel after each duplicate-clip problem, the new workflow sends the whole batch (1-5) and reports back at the end with an alert box where the user can only say OK (no retry option, because it will fail again).
- Clip sending will still be retried 3 times internally e.g. temporary for communication errors, but without the Retry button (which you can never pass) there will be 3 errors recorded for duplicate clips.
- This enables a better experience for the user and a better feedback on what is actually happening.
- We will have exactly 3 error record (num/3 actual errors) where we can further research.

Edit: Here are the final errors (JS alerts)...

If none could be uploaded (maybe 1-5)
image

If some could be uploaded (maybe 1-4)
image

### DIFF
--- a/web/locales/common-voice/en/pages/error.ftl
+++ b/web/locales/common-voice/en/pages/error.ftl
@@ -6,9 +6,18 @@ banner-error-slow-link = Status Page
 error-something-went-wrong = Sorry, something went wrong
 error-clip-upload = Upload of this clip keeps failing, keep retrying?
 error-clip-upload-server = Upload of this clip keeps failing at server. Reload the page or try again later.
-error-duplicate-clips =  Only { $uploaded } of { $total } clips could be uploaded. It looks like you've already recorded others. Let's go ahead and try the next set!
 error-title-404 = We couldn’t find that page for you
 error-content-404 = Maybe our <homepageLink>homepage</homepageLink> will help? To ask a question, please join the <matrixLink>Matrix community chat</matrixLink>, monitor site issues via <githubLink>GitHub</githubLink> or visit <discourseLink>our Discourse forums</discourseLink>.
 error-title-503 = We’re experiencing unexpected downtime
 error-content-503 = The site will be back up as soon as possible. For the latest information, please join the <matrixLink>Matrix community chat</matrixLink> or visit <githubLink>GitHub</githubLink> or <discourseLink>our Discourse forums</discourseLink> to submit and monitor site experience issues.
 error-code = Error { $code }
+
+# Warning message shown when none of the clips could be uploaded
+error-duplicate-clips-all =
+  We could not upload { NUMBER($total) ->
+    [one] your clip. It has already been uploaded before.
+    *[other] { $total } clips. They have already been uploaded before.
+  } Let’s continue with the next batch!
+# Warning message shown when only some of the clips could be uploaded (uploaded count will be <5)
+error-duplicate-clips-some =
+  We uploaded { $uploaded } of your clips — The remainder have already been uploaded. Let’s continue with the next batch!
--- a/web/locales/common-voice/en/pages/error.ftl
+++ b/web/locales/common-voice/en/pages/error.ftl
@@ -6,6 +6,7 @@ banner-error-slow-link = Status Page
 error-something-went-wrong = Sorry, something went wrong
 error-clip-upload = Upload of this clip keeps failing, keep retrying?
 error-clip-upload-server = Upload of this clip keeps failing at server. Reload the page or try again later.
+error-duplicate-clips =  Only { $uploaded } of { $total } clips could be uploaded. It looks like you've already recorded others. Let's go ahead and try the next set!
 error-title-404 = We couldn’t find that page for you
 error-content-404 = Maybe our <homepageLink>homepage</homepageLink> will help? To ask a question, please join the <matrixLink>Matrix community chat</matrixLink>, monitor site issues via <githubLink>GitHub</githubLink> or visit <discourseLink>our Discourse forums</discourseLink>.
 error-title-503 = We’re experiencing unexpected downtime
--- a/web/src/components/pages/contribution/speak/speak.tsx
+++ b/web/src/components/pages/contribution/speak/speak.tsx
@@ -407,17 +407,26 @@ class SpeakPage extends React.Component<Props, State> {
 
     this.setState({ clips: [], isSubmitted: true })
 
+    const clip_count = clips.length
+    let duplicate_clip = false
+    let uploaded_count = 0
+
     addUploads([
       ...clips.map(({ sentence, recording }) => async () => {
         let retries = 3
         while (retries) {
           try {
+            const res = await api.uploadClip(
+              recording.blob,
+              sentence.id,
+              this.demoMode
+            )
             const {
               showFirstContributionToast = false,
               hasEarnedSessionToast = false,
               showFirstStreakToast = false,
               challengeEnded = true,
-            } = await api.uploadClip(recording.blob, sentence.id, this.demoMode)
+            } = res
             URL.revokeObjectURL(recording.url)
             try {
               sessionStorage.setItem(
@@ -462,17 +471,21 @@ class SpeakPage extends React.Component<Props, State> {
               tallyRecording()
             }
             retries = 0
+            uploaded_count += 1
           } catch (error) {
             let key = 'error-clip-upload'
-
-            if (error.message.includes('save_clip_error')) {
+            console.log(error)
+            if (error.message.includes('ALREADY_EXISTS')) {
+              duplicate_clip = true
+              key = 'error-duplicate-clip'
+            } else if (error.message.includes('save_clip_error')) {
               key = 'error-clip-upload-server'
             }
 
             retries--
             await new Promise(resolve => setTimeout(resolve, 1000))
 
-            if (retries == 0 && confirm(getString(key))) {
+            if (retries == 0 && !duplicate_clip && confirm(getString(key))) {
               retries = 3
             }
           }
@@ -481,14 +494,23 @@ class SpeakPage extends React.Component<Props, State> {
       async () => {
         trackGtag('submit-clips', { locale })
         refreshUser()
-        addNotification(
-          <>
-            <CheckIcon />{' '}
-            <Localized id="clips-uploaded">
-              <span />
-            </Localized>
-          </>
-        )
+        if (uploaded_count == clip_count) {
+          addNotification(
+            <>
+              <CheckIcon />{' '}
+              <Localized id="clips-uploaded">
+                <span />
+              </Localized>
+            </>
+          )
+        } else if (duplicate_clip) {
+          alert(
+            getString('error-duplicate-clips', {
+              uploaded: uploaded_count,
+              total: clip_count,
+            })
+          )
+        }
       },
     ])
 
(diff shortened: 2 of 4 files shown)


web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -407,17 +407,26 @@ class SpeakPage extends React.Component<Props, State> {
...
+    const clip_count = clips.length
+    let duplicate_clip = false
+    let uploaded_count = 0
+
     addUploads([
       ...clips.map(({ sentence, recording }) => async () => {
         let retries = 3
         while (retries) {
           try {
+            const res = await api.uploadClip(
```
moz-dfeller: There is no need to capture the response in `res` here and then destructure it, if you're not using `res` for anything else. The way it was before is usually the way to go.

web/locales/common-voice/en/pages/error.ftl
```diff
@@ -6,6 +6,7 @@ banner-error-slow-link = Status Page
 error-something-went-wrong = Sorry, something went wrong
 error-clip-upload = Upload of this clip keeps failing, keep retrying?
 error-clip-upload-server = Upload of this clip keeps failing at server. Reload the page or try again later.
+error-duplicate-clips =  Only { $uploaded } of { $total } clips could be uploaded. It looks like you've already recorded others. Let's go ahead and try the next set!
```
moz-dfeller: `Only { $uploaded } of { $total } clips could be uploaded` sounds so pessimistic and ` It looks like you've already recorded others` sounds so vague and disconnected from the current batch.
How about this:
```suggestion
error-duplicate-clips =  We uploaded { $uploaded } out of { $total } clips—it looks like the others are already recorded. Let’s move on to the next batch!
```

moz-dfeller (changes requested):

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -462,17 +471,21 @@ class SpeakPage extends React.Component<Props, State> {
               tallyRecording()
             }
             retries = 0
+            uploaded_count += 1
           } catch (error) {
             let key = 'error-clip-upload'
-
-            if (error.message.includes('save_clip_error')) {
+            console.log(error)
```
moz-rotimib: Remove console.log

web/src/components/pages/contribution/speak/speak.tsx
```diff
@@ -603,6 +631,50 @@ class SpeakPage extends React.Component<Props, State> {
       <>
         <div id="speak-page">
           {noNewClips && isLoading && <Spinner delayMs={500} />}
+          {isSubmitted && true && (
```
moz-rotimib: I don't think this `true` is needed here. If yes then it should be assigned to a variable so it's clear what it means

web/src/components/pages/contribution/speak/speak.tsx
moz-bozden: Yep, I was doing it for debugging, remnant...

moz-bozden: Done.

web/src/components/pages/contribution/speak/speak.tsx
moz-bozden: Done

web/locales/common-voice/en/pages/error.ftl
moz-bozden: The plural wording is a bit more complex/conditional...
We can have 1 to 5 recordings and any number of them can be duplicate.
If 1 clip recorded and failed, or if N clips recorded and all failed => the the wording becomes bad...
We need two wordings, conditionally...

"Let's go ahead and try the next one!" was recommended before when we were talking handling them one by one ( I think "let's move on" feels a bit imperative, also implies "it is not important" or "it was a failure" - as a L2 speaker thou).

Something like:

- N of N failed (singular/plural form in Fluent):
  - We could not upload { $total } clip(s) — it looks like they were already recorded by you. Let's go ahead and try the next batch!
 - M of N failed (M<N) (singular/plural form in Fluent):
   - We uploaded { $uploaded } out of { $total } clip(s) — it looks like the others are already recorded by you. Let's go ahead and try the next batch!
   
Let's wait for input from @ftyers for this one...

ftyers: Maybe something like:

`error-duplicate-clips =  We uploaded { $uploaded } out of { $total } clips—The remainder have already been uploaded. Let’s continue with the next batch!`

Is `$total` always 5? Ideally we want to avoid situations where there are two numbers if one of them remains the same, because otherwise we will end up getting plural forms * N translations where N is the number of numbers in the string. 

The other thing we could do is split it into two strings.

moz-bozden: @ftyers, it will be mostly 5 but not always. The last batch can have remaining (1-5). We request 25 sentences from FE, but BE sends whatever is available (up to 25), it could be 3 for example. This particularly happens with new/low resourced languages and very active users (leads).

I implemented it as two sentences, conditionally shown... Coming with updates.

moz-bozden: Like this?

```
error-duplicate-clips-all =
  We could not upload { NUMBER($total) ->
    [one] your clip. It has already been uploaded before.
    *[other] { $total } clips. They have already been uploaded before.
  } Let’s continue with the next batch!
error-duplicate-clips-some =
  We uploaded { $uploaded } out of { NUMBER($total) ->
    [one] clip
    *[other] { $total } clips
  } — The remainder have already been uploaded. Let’s continue with the next batch!
```

moz-bozden: > otherwise we will end up getting plural forms * N translations where N is the number of numbers in the string.

This happens when all numbers have possible plural forms. E.g. if we had:
```
"Uploaded ${ uploaded } clip(s) out of ${ total } clip(s)."
```
That would result in 4 combinations... Combinations of binary values = 2^N, for 5 variables it will be 32...

ftyers: But for those languages that have e.g. 5 plural forms wouldn't this create the same problem as that other time?

moz-bozden: For example for the one below, in a 5-plural for language, there will be 5 combinations.
```
error-duplicate-clips-some =
  We uploaded { $uploaded } out of { NUMBER($total) ->
    [one] clip
    *[other] { $total } clips
  } — The remainder have already been uploaded. Let’s continue with the next batch!
```

The one we fixed last month (https://github.com/common-voice/common-voice/pull/4811/files) had 4 variables. Combinations for 5 plural forms with 4 variables would be 5^4 = 50,000. That was why it was crashing for some languages :)

```
activity-needed-calculation-plural =
    { NUMBER($totalHours) ->                                         // 1
        [one] { $totalHours } hour
       *[other] { $totalHours } hours
    } is achievable in just over { NUMBER($periodMonths) ->          // 2
        [one] { $periodMonths } month
       *[other] { $periodMonths } months
    } if { NUMBER($people) ->                                        // 3
        [one] { $people } person
       *[other] { $people } people
    } record { NUMBER($clipsPerDay) ->                               // 4
        [one] { $clipsPerDay } clip
       *[other] { $clipsPerDay } clips
    } a day.
```


It is all a matter of wording. When thinking in English and wording it, people tend to create nice, descriptive sentences with multiple variables - but actually you are presenting e.g. statistics. In fact these can be simplified most of the time. For example pre-calculation helps (like I did on the fix you mentioned):
- 3/5 - no plurals
- 60% - no plurals

(Wow, my first localization was in 2000 for an ASP based forum as the internationalization moderator, been 25 years)

moz-bozden: What do you think of the wording below @ftyers ?
- First one is if every single one fails
- Second one is part of a batch fail.

I control it from code. Otherwise the wording gets impossible.

> Like this?
> 
> ```
> error-duplicate-clips-all =
>   We could not upload { NUMBER($total) ->
>     [one] your clip. It has already been uploaded before.
>     *[other] { $total } clips. They have already been uploaded before.
>   } Let’s continue with the next batch!
> error-duplicate-clips-some =
>   We uploaded { $uploaded } out of { NUMBER($total) ->
>     [one] clip
>     *[other] { $total } clips
>   } — The remainder have already been uploaded. Let’s continue with the next batch!
> ```

ftyers: I think maybe the quickest thing is to get feedback from @bandrandr for the Russian case (which is a medium difficulty one) if it works for Russian hopefully it won't be too horrible for Welsh/Irish.

moz-bozden: When I translated to Turkish, I recognized no singular would be needed in second error message.
1 out of one is uploaded would not result in this error.

So they should be (with Turkish translations):

```
error-duplicate-clips-all =
  We could not upload { NUMBER($total) ->
    [one] your clip. It has already been uploaded before.
    *[other] { $total } clips. They have already been uploaded before.
  } Let’s continue with the next batch!
```

Singular: Ses kaydınızı yükleyemedik. Daha önce yüklenmiş. Hadi bir sonraki grupla devam edin!
Plural: { $total } ses kaydınızı yükleyemedik. Daha önce yüklenmişler. Hadi bir sonraki grupla devam edin!

```
error-duplicate-clips-some =
  We uploaded { $uploaded } out of { $total } clips — The remainder have already been uploaded. Let’s continue with the next batch!
```

{ $total } ses kaydından { $uploaded } tanesi yüklendi - Diğerleri daha önce yüklenmiş. Hadi bir sonraki grupla devam edin!

moz-bozden: @moz-rotimib can you please check the Fluent for correctness?

moz-bozden: These translations came from the Circassian group. I think the second plural form is missing...

Case-1
- Singular: Мы не смогли загрузить вашу аудиозапись. Она загружена ранее. Продолжайте со следующей группой! 
- Plural: Мы не смогли загрузить { $total } из ваших аудиозаписей. Их загрузили ранее. Продолжайте со следующей группой!

Case-2
- Из{ $total } аудиозаписей  { $uploaded } загружены. Остальные были загружены ранее. Продолжайте со следующей группой!

ftyers: My suggestion is that we just drop the multiple numbers in the same localisation string, so we could go with:
```
error-duplicate-clips-some =
  We uploaded { $uploaded } of your clips — The remainder have already been uploaded. Let’s continue with the next batch!
```

moz-bozden: That would resolve many Pontoon translation issues, thank you...
I'd say let's implement this and find the real reason behind this bug...
In any case, if we get negative community feedback, we can easily change it.
IMO, anything is better wrt the current UX...

moz-bozden:
I implemented the last proposed sentence options and put the screenshots at the top.
If we can quickly merge and release this today, we will be able "to start to search" for the real reason...
Thank you all for your quick reviews 💯

moz-dfeller (approved):
lgtm

PR #4967, opened 2025-07-03
[WIP] 1208-MCV-BE Add simple stats for not yet validated sentences to api

moz-bozden:
NOTE: This PR is thought to be merged together with the BE part #4968 , but it will still function as API endpoint on its own.

Output will be like the following:

```json
  {
    "id": 350,
    "is_contributable": 1,
    "is_ss_contributable": 0,
    "localizedPercentage": 100,
    "recordedHours": 0,
    "validatedHours": 0,
    "invalidatedHours": 0,
    "speakersCount": 1,
    "sentencesCount": {
      "targetSentenceCount": 5000,
      "currentCount": 12,
      "votableCount": 36            // <== ADDED
    },
    "locale": "tr",
    "lastFetched": "2025-07-01T23:21:18.300Z"
  },
```

### DIFF
--- a/server/src/lib/model/db.ts
+++ b/server/src/lib/model/db.ts
@@ -1082,6 +1082,78 @@ export default class DB {
     }
   }
 
+  // TODO: This is possibly cacheable
+  // Can be used to speed-up findSentencesForReview in application/repository/sentences-repository.ts
+  async getVotableLanguageSentenceCounts(
+    localeId: number
+  ): Promise<{ locale_id: number; count: number }> {
+    const [[row]] = await this.mysql.query(
+      `
+      SELECT COUNT(*) AS votable_sentence_count
+      FROM sentences s
+      WHERE s.locale_id = ?
+        AND s.is_validated = FALSE
+        AND NOT (
+          s.is_validated = TRUE AND s.is_used = FALSE
+        )
+        AND NOT EXISTS (
+          SELECT 1
+          FROM sentence_votes sv
+          WHERE sv.sentence_id = s.id
+          GROUP BY sv.sentence_id
+          HAVING
+            COUNT(*) >= 2 AND (
+              SUM(sv.vote = 1) >= 2
+              OR SUM(sv.vote = 0) >= 2
+              OR (SUM(sv.vote = 1) = 2 AND SUM(sv.vote = 0) = 1)
+            )
+        )
+      `,
+      [localeId]
+    )
+
+    return {
+      locale_id: localeId,
+      count: row?.votable_sentence_count || 0,
+    }
+  }
+
+  // Not in use yet, but ready for next phase (with caching)
+  // TODO: Needs performance measures on real data first
+  async getVotableSentenceCounts(): Promise<
+    { locale_id: number; count: number }[]
+  > {
+    const rows = await this.mysql.query(
+      `
+      SELECT s.locale_id, COUNT(*) AS votable_sentence_count
+      FROM sentences s
+      WHERE s.is_validated = FALSE
+        AND NOT (
+          s.is_validated = TRUE AND s.is_used = FALSE
+        )
+        AND NOT EXISTS (
+          SELECT 1
+          FROM sentence_votes sv
+          WHERE sv.sentence_id = s.id
+          GROUP BY sv.sentence_id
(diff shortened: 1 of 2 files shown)


server/src/lib/model/db.ts
```diff
@@ -1082,6 +1082,78 @@ export default class DB {
     }
   }
 
+  // TODO: This is possibly cacheable
+  // Can be used to speed-up findSentencesForReview in application/repository/sentences-repository.ts
```
moz-dfeller, line 1086: The result itself is already cached at a higher level under the key `get-all-language-stats`, so this is call is indirectly cached as well.

server/src/lib/model/db.ts
```diff
@@ -1082,6 +1082,78 @@ export default class DB {
...
+
+    return {
+      locale_id: localeId,
+      count: row?.votable_sentence_count || 0,
+    }
+  }
+
+  // Not in use yet, but ready for next phase (with caching)
+  // TODO: Needs performance measures on real data first
+  async getVotableSentenceCounts(): Promise<
```
moz-dfeller, line 1123: That's basically the same as `getVotableLanguageSentenceCounts` right, except that you get the result for all locales instead of one specific. Would it make sense to to only cache this result and filter by language if needed instead of calling `getVotableLanguageSentenceCounts` which should give you the same result?

common/language-stats.ts
```diff
@@ -1,6 +1,7 @@
 export type SentenceCount = {
   currentCount: number;
   targetSentenceCount: number;
+  votableCount: number;
```
moz-dfeller, line 4: Not sure about `votableCount`, I know you probably don't like `unvalidatedCount`, I was thinking maybe `inReviewCount`, `missingVoteCount`. What do you think?

moz-bozden: Yep, "votable" is not good (also bad spelling, should be `voteable` ), couldn't find a better one at that time.
I think in/un-validated distinction can not be clear, and these are not missing vote counts.
I keep using "not-yet-validated" on common channels for clarity, but it would be very long.

What about these as brainstorming (without "...Count" we can have better ones, it is also under "SentenceCount", so maybe both can be left out):
- sentencesAwaitingVotes
- pendingVoteTotal
- toBeVotedCount
- availableForVoting
- ...

server/src/lib/model/db.ts
moz-bozden: Yep, for 12 hours - it is a long duration for our purposes I think. And that stats process takes half an hour to run (it seems) - very good candidate for a spawn if we could do it.

server/src/lib/model/db.ts
moz-bozden: That was my idea as I stated in the comment, but without any measures in hand I could not decide.
My local tests with limited data shows looping through locales is costly (I had `ca` sentences in). But I don't know the hit this will make with GROUP BY on 25M+ sentences in DB.
I checked the indexes, and they seem to be fine performance-wise.

server/src/lib/model/db.ts
moz-bozden: What I mean is: If we have an hourly cache, we can directly use it in stats process, without any re-calculation. It is already a snapshot of a running process where numbers change constantly, it does not need to be exact...

The main idea behind this is: A community lead and members should know if new sentences are added and how many (roughly), 2-3 or 4000. Many communities do not check the sentence-validation regularly, because usually not much happens there, and if they look they should now the amount.

We had such a case for two languages I helped with. Somebody added 2*2k proverbs, but they all were malformed. I was aware and checking, so I alarmed the core team and 3 people worked on in-validating them for two days.

moz-bozden:
@moz-dfeller, maybe let us wait until we get a more precise view on query performances on real data.
If it will not take weeks of course...

I'll have a table for each SELECT and cache info, tested on multiple languages (small-mid range-large).
In my local tests the size of the data effects too much. Not intuitive at the first look, but in some queries languages with larger data perform much better (e.g. LIMIT 1000 can easily be fulfilled).

moz-bozden:
Closing to re-implement on next phase...
Quite a few changed and will be changing in that respect...

PR #4969, opened 2025-07-05
Update sentence bands

moz-bozden:
This migration sets sentence bands after a detailed analysis.
- Compared Pontoon locale data with MCV locales, checking population values
- Already is_contributable locales are left out
- Bands could be set to a lower level for any reason, these are also left out
- Thus this migration only sets locales (mostly new additions) to a lower band


ftyers (approved):
Lgtm

PR #4971, opened 2025-07-09
1188 create endpoint to create client credentials

moz-dfeller:
Creating endpoints to handle API client credentials creation.

### DIFF
--- a/server/src/application/profile/command-handler/deleteClientCredentialsCommandHandler.ts
+++ b/server/src/application/profile/command-handler/deleteClientCredentialsCommandHandler.ts
@@ -1,0 +1,10 @@
+import { deleteClientCredentials } from '../../../infrastructure/authentication/authentication'
+import { DeleteClientCredentialsCommand } from './command/deleteClientCredentialsCommand'
+
+export const deleteClientCredentialsCommandHandler = async (
+  cmd: DeleteClientCredentialsCommand
+) => {
+  await deleteClientCredentials({
+    ...cmd,
+  })
+}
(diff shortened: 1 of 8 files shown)


moz-bozden:
Now we have a third - outside API concept (after API's in BE & FE), maybe naming is more critical. And adding some comments what the API is about. IMO "PublicAPI" does not cut it.

moz-dfeller:
Fabulous job @akrabat, thanks! 🚀

server/src/application/profile/command-handler/deleteClientCredentialsCommandHandler.ts
```diff
@@ -0,0 +1,10 @@
+import { deleteClientCredentials } from '../../../infrastructure/authentication/authentication'
+import { DeleteClientCredentialsCommand } from './command/deleteClientCredentialsCommand'
+
+export const deleteClientCredentialsCommandHandler = async (
+  cmd: DeleteClientCredentialsCommand
+) => {
+  await deleteClientCredentials({
+    ...cmd,
+  })
+}
```
moz-dfeller, line 10: You're using async/await syntax here but you're not doing anything with the result of the `deleteClientCredentials` function call, so you might as well just return the Promise straight to the caller. That's the only thing I think might be unnecessary. Everything else is 🥇 
It's also just a comment, feel free to ignore it. The compiler is going to optimize this away anyways.
```suggestion
export const deleteClientCredentialsCommandHandler = (
  cmd: DeleteClientCredentialsCommand
) => {
  return deleteClientCredentials({
    ...cmd,
  })
}
```

PR #4982, opened 2025-07-18
Add `common-voice-network` Docker network

akrabat:
Create a second network to allow other Docker projects to connect their services to the `common-voice-network` and so communicate together.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -3,6 +3,7 @@ services:
     image: mysql:8.0.31
     networks:
       - voice-web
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
moz-dfeller, line 1: You'll need to add the network to the DB at least as well, right?

akrabat: Oh yes. I misread the file and attached to the redis container by accident. 

Fixed.

moz-dfeller (approved):

PR #4989, opened 2025-07-23
Fix for stylus causing builds to fail

moz-rotimib:
Add security override for stylus dependency

- https://github.com/stylus/stylus/issues/2938
- https://github.com/mrmckeb/typescript-plugin-css-modules/issues/287


moz-bozden:
Additional follow up here:

https://github.com/stylus/stylus/issues/2938#issuecomment-3106668235

PR #4992, opened 2025-07-23
Fix stylus dependency

moz-rotimib:


moz-rotimib:
Closing because the security issue with stylus has been fixed

https://github.com/stylus/stylus/issues/2938

PR #5003, opened 2025-07-26
1269-Add Irish (ga-IE) variants

moz-bozden:
This PR only adds Irish (`ga-IE`) variants **for sentence-variant usage ONLY**.
This is the first step of handling the FR https://github.com/common-voice/common-voice/issues/4993
Other steps need examination of existing data, data-mapping, programming, and migrations.

If in the meantime, the community members select their variants in their profiles, any 1-to-1 mapping from existing accents will be ignored and disabled.

### DIFF
--- a/server/src/lib/model/db/language-data/variants.ts
+++ b/server/src/lib/model/db/language-data/variants.ts
@@ -505,5 +505,25 @@ export const VARIANTS: Variant[] = [
     locale_name: 'gsw',
     variant_name: 'Hochàlemmànnisch (Hìniga, Pfìrt, usw.)',
     variant_token: 'gsw-FR-hochalem',
-  }
+  },
+  {
+    locale_name: 'ga-IE',
+    variant_name: 'Gaeilge Uladh - Ulster Irish',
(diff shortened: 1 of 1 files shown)


server/src/lib/model/db/language-data/variants.ts
```diff
@@ -505,5 +505,25 @@ export const VARIANTS: Variant[] = [
     locale_name: 'gsw',
     variant_name: 'Hochàlemmànnisch (Hìniga, Pfìrt, usw.)',
     variant_token: 'gsw-FR-hochalem',
-  }
+  },
+  {
+    locale_name: 'ga-IE',
+    variant_name: 'Gaeilge Uladh - Ulster Irish',
```
ftyers: I think we can leave the English translation out, people using the interface should know Irish well enough, and even if not, the word for Irish + U M C should be obvious.

ftyers (commented):
I think we can leave the English out.

moz-bozden:
Removed the English translations.

PR #5022, opened 2025-08-03
1277-Add new Irish (ga-IE) accents

moz-bozden:
Should be merged after https://github.com/common-voice/common-voice/pull/5015


ftyers (approved):
Lgtm

PR #5032, opened 2025-08-07
Fix: Preserve locale when redirecting to spontaneous speech page

NaveenKancharla28:
### Type of Pull Request

-  Bug fix(x)
-  Bulk sentence upload
-  Related to a listed issue
-  Other: UX improvement in locale persistence(x)

### Description of the Pull Request

This PR ensures that the user's selected locale is preserved when navigating to the Spontaneous Speech contribution page (`/speak` or `/listen`). Previously, the locale context was lost, causing redirection to the default language instead of the selected one (e.g., `en`, `te`, etc.).

- Introduced `useParams()` from `react-router-dom` to dynamically extract locale from the URL.
- Updated `internalHref` routes in `<LocaleLink>` to prepend the extracted locale.
- Verified navigation preserves the correct locale when switching between menu items.

This fix was independently researched and implemented by [Naveen Chaitanya](https://github.com/NaveenKancharla28), based on user reports and personal debugging. Thanks to the Common Voice community for the opportunity to contribute!

### DIFF
--- a/web/src/components/layout/nav/contribute-menu/menu-item-renderer.tsx
+++ b/web/src/components/layout/nav/contribute-menu/menu-item-renderer.tsx
@@ -1,19 +1,20 @@
-import * as React from 'react'
-import { Localized, WithLocalizationProps } from '@fluent/react'
-import classNames from 'classnames'
-import { Tooltip } from 'react-tooltip'
+import * as React from 'react';
+import { Localized, WithLocalizationProps } from '@fluent/react';
+import classNames from 'classnames';
+import { Tooltip } from 'react-tooltip';
+import { useParams } from 'react-router-dom'; //  ADDED for locale
 
-import { ContributeMenuItem } from '.'
-import { LocaleLink } from '../../../locale-helpers'
-import URLS from '../../../../urls'
+import { ContributeMenuItem } from '.';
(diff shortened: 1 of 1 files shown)


web/src/components/layout/nav/contribute-menu/menu-item-renderer.tsx
```diff
@@ -1,37 +1,41 @@
-import * as React from 'react'
-import { Localized, WithLocalizationProps } from '@fluent/react'
-import classNames from 'classnames'
-import { Tooltip } from 'react-tooltip'
+import * as React from 'react';
+import { Localized, WithLocalizationProps } from '@fluent/react';
+import classNames from 'classnames';
+import { Tooltip } from 'react-tooltip';
+import { useParams } from 'react-router-dom'; //  ADDED for locale
```
NaveenKancharla28, line 5: Added for locale

moz-bozden:
Dear @NaveenKancharla28, thank you for the PR... Although it is a plausible solution, it is a two way street, it had to be implemented in both repos - which we already did, but using localstorage, as the Spontaneous Speech webapp does not use UI language in its URL, already keeps it in localstorage.

Ref: #5043

I'm now closing this, but please keep up the good work.

PR #5034, opened 2025-08-10
1283-Fixes issues in the new guidelines pages

moz-bozden:
Fixes the following issues from staging server:

- When a header on the left navigation has no items and is clicked, it does not bring the related section
- Fix some content for clarity


moz-bozden:
Quickly merging because the keys are already in Pontoon.

PR #5038, opened 2025-08-13
1288-Some fixes to localizations

moz-bozden:
- Missing age strings in profile
- Error when a '' value is tried to be localized in select in profile
- Email privacy clause new version not applied
- Make Spontaneous Speech dataset Cypriot Greek `el-CY` translatable

### DIFF
--- a/web/locales/common-voice/en/pages/profile/info.ftl
+++ b/web/locales/common-voice/en/pages/profile/info.ftl
@@ -29,14 +29,7 @@ native-language =
 profile-form-add-accent = Add new custom accent "{ $inputValue }"
 profile-form-submit-save = Save
 profile-form-submit-saved = Saved
-male_masculine = Male/Masculine
-female_feminine = Female/Feminine
-intersex = Intersex
-transgender = Transgender
-non-binary = Non-binary
-do_not_wish_to_say = Don't wish to say
-# Gender
-other = Other
+
 why-profile-title = Why a profile?
 why-profile-text =
     By providing some information about yourself, the audio data you submit to Common Voice will be more useful to Speech
@@ -115,3 +108,23 @@ delete-api-key-success-toast-message = API credential deleted successfully
 add-api-key-name-error = Add an API credential name
 max-api-keys-reached = Exceeded API credentials limit of 10
 duplicate-api-key-name-error = Duplicate API key name
+
+# Genders
+male_masculine = Male/Masculine
+female_feminine = Female/Feminine
+intersex = Intersex
+transgender = Transgender
+non-binary = Non-binary
+do_not_wish_to_say = Don't wish to say
+other = Other
+
+# Age Values
+teens = < 19
+twenties = 19 - 29
+thirties = 30 - 39
+fourties = 40 - 49
+fifties = 50 - 59
+sixties = 60 - 69
+seventies = 70 - 79
+eighties = 80 - 89
+nineties = 90 - 99
--- a/web/locales/spontaneous-speech/en/pages/common.ftl
+++ b/web/locales/spontaneous-speech/en/pages/common.ftl
@@ -50,22 +50,22 @@ accept-privacy-and-terms = I agree to Common Voice's <termsLink>Terms</termsLink
 bsy = Sabah Bisaya
 cdo = Fuzhou
 cpx = Heng Hua
-en = English
-fr = French
 hch = Wixarika
 kzi = Kelabit
 mbf = Baba Malay
 meh = Southwestern Tlaxiaco Mixtec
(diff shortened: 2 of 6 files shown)


web/src/components/pages/profile/info/languages/languages.tsx
```diff
@@ -149,11 +149,11 @@ function ProfileInfoLanguages({
...
         </Button>
 
-        {!hasUserLanguages && (
+        {/* {!hasUserLanguages && (
           <Localized id="profile-select-language">
-            <span className="no-languages" />
+            <span className="no-languages">HERE</span>
           </Localized>
-        )}
+        )} */}
```
moz-dfeller: Looks like debug stuff?

moz-bozden: `HERE` was for debug, but the code was there and was not effective on my local tests, so I commented it out.
Let me test/check it again once more...

moz-bozden: Yes, the `profile-select-language` key is not in Pontoon, so it does not do anything except adding an empty span tag.
That part was replaced by the Button above already:
```
        <Button
          className="add-language"
          outline
          disabled={hasNewEmptyLanguage}
          onClick={handleAddNewLanguageButtonClick}>
          <Localized id="add-language">
            <span />
          </Localized>
          <span aria-hidden={true}>+</span>
        </Button>
```

I'll remove it completely.

moz-bozden: Removed...

web/locales/spontaneous-speech/en/pages/common.ftl
```diff
@@ -50,23 +50,23 @@ accept-privacy-and-terms = I agree to Common Voice's <termsLink>Terms</termsLink
 bsy = Sabah Bisaya
 cdo = Fuzhou
 cpx = Heng Hua
-en = English
```
ftyers, line 53: Why are en/fr/tr removed here?

web/locales/common-voice/en/pages/profile/info.ftl
```diff
@@ -115,3 +108,23 @@ delete-api-key-success-toast-message = API credential deleted successfully
...
+# Genders
+male_masculine = Male/Masculine
+female_feminine = Female/Feminine
+intersex = Intersex
+transgender = Transgender
+non-binary = Non-binary
+do_not_wish_to_say = Don't wish to say
+other = Other
+
+# Age Values
```
ftyers, line 121: I realise this isn't the place, but it is concievable that we have people > 99 contributing to some projects and probably it makes more sense to just be >= 90 "over 90" as a category.

web/locales/spontaneous-speech/en/pages/common.ftl
moz-bozden: These definitions are from the past, before the cv-ss Pontoon connections are in place. These were used to convert language/locale codes to language/locale names so that can be shown in pickers.
Currently they come from Pontoon => CV => SS - if they are in Pontoon.
Those NOT in Pontoon (e.g. variants like el-CY) we should define them here.
Those already in Pontoon we can remove...
I did not have time to cross-check all of them, so I only removed 3 well-known ones :/
In a near future I'll check/remove the rest (some were added recently by you, so I did not risk it)..
Multiple but different definitions can also be problematic.

web/locales/common-voice/en/pages/profile/info.ftl
moz-bozden: I concur. There is also a discrepancy in the "teens" definition. Previously it was set (max of international legal definitions):
Teens < 20
twenties 20-29

Now we have two different ones:
Teens < 19
Which makes "twenties" in the datasets a bit shaky... What if I'm 19 (I wish I were!)

moz-bozden:
@ftyers, answered your comments...

web/locales/spontaneous-speech/en/pages/common.ftl
ftyers: Ok, got it, maybe add a comment there to that effect?

web/locales/common-voice/en/pages/profile/info.ftl
ftyers: Ok, that's fine, just a comment. :)

web/locales/spontaneous-speech/en/pages/common.ftl
moz-bozden: Added to the PR description. Also adding to my to-fix list on internal 1306.

PR #5043, opened 2025-08-14
1299-V1 Make use of local storage to pass UI language info to SS

moz-bozden:
Spontaneous Speech uses browser local-storage to keep the current UI language.
This PR does two things:
- Implement local-storage based UI language passing (from SS to CV or vice-versa)
- Makes menu items adaptive to installation (they will not point to production site anymore, a prerequisite for the time when SS repo becomes public)

Notes:
- In some cases localStorage cannot be used (e.g. private tabs), so will will fall-back to English (if the URL does not contain a language). Actually this is the current behavior.
- Another PR in SS repo will compliment this

### DIFF
--- a/web/src/components/languages-provider.tsx
+++ b/web/src/components/languages-provider.tsx
@@ -21,6 +21,25 @@ import { useLocale } from './locale-helpers'
 
 const SentryRoute = Sentry.withSentryRouting(Route)
 
+const STORED_LOCALE_KEY = 'current-locale'
+
+// Safe localStorage helper functions
+function getStoredLocale() {
+  try {
+    return localStorage.getItem(STORED_LOCALE_KEY)
+  } catch (e) {
+    return null
+  }
+}
+
+function setStoredLocale(locale: string) {
+  try {
+    localStorage.setItem(STORED_LOCALE_KEY, locale)
+  } catch (e) {
+    // noop
+  }
+}
+
 interface LanguageRoutesProps {
   userLocales: string[]
   setUserLocales: (newUserLocales: string[]) => void
@@ -44,6 +63,7 @@ const LanguageRoutes = ({
   useEffect(() => {
     if (primaryUserLocale !== prevLocaleRef.current) {
       prevLocaleRef.current = primaryUserLocale
+      setStoredLocale(primaryUserLocale) // <== Save to localStorage when locale changes
     }
   }, [primaryUserLocale])
 
@@ -146,13 +166,11 @@ const LanguagesProvider = ({ children }: LanguagesProviderProps) => {
 
   async function updateLocalization() {
     const localizationUserLocales = [...userLocales]
-
     const pathname = history.location.pathname
 
     if (!languages.translatedLocales.includes(userLocales[0])) {
       localizationUserLocales[0] = DEFAULT_LOCALE
       setUserLocales(localizationUserLocales)
-
       setLocale(DEFAULT_LOCALE)
       history.replace(replacePathLocale(pathname, DEFAULT_LOCALE))
     } else {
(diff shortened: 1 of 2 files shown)


ftyers:
- Should it fallback to English or to the user's language preferences? (if the language preferences are available in private browsing mode)
- "Makes menu items adaptive to installation" - what does this mean, could you explain a bit more? 

Also a quick note that this PR is aimed at fixing: https://github.com/common-voice/common-voice/issues/5011

moz-bozden:
@ftyers ...

### Makes menu items adaptive to installation

I could/cannot find a better wording for this, but it means "keeps the links in the current domain". The menu items were hard-coded to point to the production site. E.g. when on CV on staging, and select a SS related menu item it would go to production site. That was making development and testing painful and error-prone. And for this feature it was a must to fix it (just using current domain). As you know cookies, local-storage etc. are per-domain.

So instead of:
```
const SPONTANEOUS_SPEECH_PROD_ROOT_URL = 'https://commonvoice.mozilla.org/spontaneous-speech/beta'
```
I just used:
```
const SPONTANEOUS_SPEECH_ROOT_URL = `${window.location.origin}${SPONTANEOUS_SPEECH}`
```

### Language default selection

You are correct on this. In CV and SS, both Fluent's language negotiations (`@fluent/langneg`) and browser preferences (`navigator.languages`) are in play. And they are checked against the available languages on that particular installation.
As you know it is a rather complex decision logic, and hard to test for every instance. Some notes:
- The `en` is the **ultimate** fallback language in such cases
- The `@fluent/langneg` also uses `en` as default-language = (fallback)
- The user languages in browsers are very indicative for **[fingerprinting](https://developer.mozilla.org/en-US/docs/Glossary/Fingerprinting)** and browsers started to hide them from the servers (or only provide the first one or a random selection). So they are not reliable anymore, like they were 10 years before. And some just hide them in private browsing (needs research for 2025).
- SS will be a part of CV. In normal operation:
  - Users will come to CV, that would set the default language (using `@fluent/langneg` and `navigator.languages`, with a total fallback to `en` ). That would also set the local-storage value to a default.
  - User makes a language selection from the pull-down. This will be recorded in local-storage.
  - User moves to a SS menu item, the value in the local storage will be used to continue in SS.
- The reverse case (where people go to a SS item through a link) will behave the same.

I can check/test them more detailed in some free time with multiple browsers/settings, but not today :(

Personal note: SS is designed NOT to have the UI language in the URL, It would be best if we had similar URL structures, but it would be a huge undertaking for now. Like:
```
https://commonvoice.mozilla.org/en/speak
https://commonvoice.mozilla.org/en/ss/answer
https://commonvoice.mozilla.org/en/cs/answer
```

web/src/components/languages-provider.tsx
```diff
@@ -21,6 +21,25 @@ import { useLocale } from './locale-helpers'
...
+const STORED_LOCALE_KEY = 'current-locale'
+
+// Safe localStorage helper functions
+function getStoredLocale() {
+  try {
+    return localStorage.getItem(STORED_LOCALE_KEY)
+  } catch (e) {
+    return null
+  }
+}
```
moz-dfeller, line 33: Why is that needed? Calling `localStorage.getItem('doesNotexist')` returns null by default and does not throw an exception. If you're concerned whether the localStorage is even set on the window object you should check [this](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API)

moz-bozden: I'm not sure that we can assume its existence on any browser, today and in the future. I usually use this to decide (along with MDN of course):
https://caniuse.com/?search=localstorage

Also, please check the code on MDN page:

> Here is a function that detects whether localStorage is both supported and available:
```
function storageAvailable(type) {
  let storage;
  try {
    storage = window[type];
    const x = "__storage_test__";
    storage.setItem(x, x);
    storage.removeItem(x);
    return true;
  } catch (e) {
    return (
      e instanceof DOMException &&
      e.name === "QuotaExceededError" &&
      // acknowledge QuotaExceededError only if there's something already stored
      storage &&
      storage.length !== 0
    );
  }
}
```

moz-dfeller: I wouldn't optimize for an edge case where the whole site is basically unusable, i.e. if localStorage doesn't work, most probably the user can't use the MediaAPI as well to listen to and record audios. I was trying to make sure that you yourself are clear about what you actually want to protect against.

moz-bozden: IMO, It is a necessity for WebViews and non-standard implementation.

PR #5067, opened 2025-09-04
1337-Feat: Announcement component

moz-bozden:
Adds an announcement component to which will appear the top. It can be closed and the preference will be kept in a cookie (5 days). It supports start/end days.
This is the first iteration, but in the future it can be extended to put an announcement anywhere.

- [X] Design implementation (on CV)
- [X] Mobile compatibility
- [X] RTL support
- [x] Accessibility
- [X] Font support
- [x] Functionality and cookie implementation
- [x] l10n support

=> Replicate in SS


web/locales/common-voice/en/pages/common.ftl
```diff
@@ -466,3 +466,14 @@ languages-donate-banner-cta = <mark>Using CV for</mark><br/><mark>your research?
...
+## Announcements above Header
+
+## MDC Announcement
+
+# Main text for MDC announcement. Text wrapped in <strong></strong> will be rendered with bold font
+announcement_mdc_text = <strong>New Common Voice datasets</strong> are now available to download exclusively through our sister platform, Mozilla Data Collective. Join for access to 300+ global datasets, built by and for the community.
+# Button text for MDC announcement
+announcement_mdc_button_text = Join Mozilla Data Collective
+# Aria text for button which opens MDC on a new page
+announcement_mdc_button_aria_text = Opens in a new tab
```
moz-rotimib: Fix: The convention is to use `kebab-case` for Fluent keys not `snake_case`

moz-bozden: Oh, thanks for finding that! Muscle memory :(

moz-bozden: Fixed...

PR #5068, opened 2025-09-06
1344-Provide support for multiple Matrix clients

moz-bozden:
Fix and extension to handle https://github.com/common-voice/common-voice/issues/5045

How it looks:

**Wide**

image

**Narrow**

image


ftyers:
This looks fine to me!

PR #5071, opened 2025-09-07
1343-Add validation schemas for language objects and apply to patch anonymous_user endpoint

moz-bozden:


server/src/lib/validation/languages.ts
```diff
@@ -0,0 +1,225 @@
...
+            type: 'string',
+          },
+          token: {
+            type: 'string',
+          },
+          clientId: {
+            type: 'string',
+          },
+        },
+      },
```
moz-dfeller, line 160: You are re-using accents schema exactly the same way twice. Why don't you extract it instead?

server/src/lib/validation/languages.ts
```diff
@@ -0,0 +1,225 @@
...
+    },
+  },
+}
+
+export const anonUserMetadataSchema: AllowedSchema = {
+  type: 'object',
+  required: [],
```
moz-dfeller, line 167: languages should be required as it makes no sense to call `saveAnonymousAccountLanguages` function without languages set, because saving languages is the only thing that this function is doing.

server/src/lib/api.ts
```diff
@@ -377,16 +387,15 @@ export default class API {
...
-    } = request
+    const client_id = request?.session?.user?.client_id
     if (!client_id) {
-      throw new ClientParameterError()
+      return response.sendStatus(StatusCodes.BAD_REQUEST)
     }
-    response.json(
+    const languages = request?.body?.languages
+    if (!languages) {
+      return response.json('no data')
```
moz-dfeller: `languages` should be required. It should return bad request as well.

server/src/lib/validation/languages.ts
moz-dfeller: Probably with a minimum of 1 item as well

server/src/lib/validation/languages.ts
moz-bozden: Of course my initial implementation was nested schemata, but that failed due to JSON Schema version incompatibilities. Although I tried a couple of methods I could not pass it. It is caused by fixed dependencies, and I could not risk at this stage to upgrade dependencies. Some dependencies use JSON Schema v4, some 7... The error was:
```
Type '{ id: { type: "number"; }; name: { type: "string"; }; sentenceCount: AllowedSchema; target_sentence_count: { type: "number"; }; is_contributable: { type: "boolean"; }; is_translated: { ...; }; native_name: { ...; }; text_direction: { ...; }; }' is not assignable to type '{ [k: string]: JSONSchema4; } | { [k: string]: JSONSchema6Definition; } | { [key: string]: JSONSchema7Definition; }'.
```

The initial implementation was like this:
```
properties: {
  accents: {
    type: "array",
    items: accentSchema
  }
}
```

moz-dfeller: So you can't do that:
```
const AccentsSchema = {
      type: 'array',
      items: {
        type: 'object',
        required: ['id', 'name'],
        properties: {
          id: {
            type: 'number',
          },
          name: {
            type: 'string',
          },
          token: {
            type: 'string',
          },
          clientId: {
            type: 'string',
          },
        },
      }
```
and then 
```
[..]
accents: AccentsSchema,
[...]
```
?

server/src/lib/validation/languages.ts
moz-bozden: There is an information flow issue there between FE/BE, where the data cannot contain a language, thus I removed it. I control it in code.

server/src/lib/api.ts
moz-bozden: Yes, better...

server/src/lib/validation/languages.ts
moz-bozden: I added a rework task for it...

server/src/lib/validation/languages.ts
moz-bozden: I added a task for it, should be done re-checked in the next dependency upgrade...

server/src/lib/api.ts
moz-bozden: Fixed, also added 0 check...

server/src/lib/validation/languages.ts
moz-bozden: Late reply: No I could not nest them with current dependencies...

PR #5073, opened 2025-09-09
1347-Rework endpoint validations

moz-bozden:
Old API endpoints reworked to check and validate every object and parameter for existence.
Converted different parameter retrieval methods into a single one for better readability and future rework.


sassela (approved):
Looks good to me! Will leave to your judgement but if this feels like a risky change, consider adding tests here.

(You mentioned in Slack it needs careful review which made me wonder if there are untested/undertested failure modes)

moz-bozden:
Thank you for the review @sassela...
The changes are simple, thus code-wise fool-proof, they will work, and tests cannot cut it.
I was wondering if these changes can break some workflows in the FE interface.
Although I checked, a second eye is always better...
Because we don't have full UI based e2e tests, it is impossible to test every combination (logged in/out) on every browser (with privacy extensions for example).
But these changes are mainly against session loss and/or malformed packets (e.g. NW errors) which we encounter in bulk.
I'll review it once more and merge and be vigilant on logs in the next release.

sassela:
Gotcha, thank you for the context.

>  I'll review it once more and merge and be vigilant on logs in the next release.

👍

PR #5095, opened 2025-10-01
Update COMMUNITIES.md: Add HormozganCV

farooqkz:
I don't need it needs explaining :)

### DIFF
--- a/docs/COMMUNITIES.md
+++ b/docs/COMMUNITIES.md
@@ -21,6 +21,8 @@ Many language communities are self organising and have their contact channels on
 - Bengali (`bn`):
   - [Matrix](https://app.element.io/#/room/#bn-common-voice:mozilla.org)
   - [Discord](https://discord.gg/uacmNh6KQj)
+- Bandari, Achomi and other languages of Hormozan and its neighborhood:
+  - [Telegram](https://t.me/hormozganCV)
 - Cantonese (`yue`):
   - [Telegram](https://t.me/commonvoiceyue)
   - [Instagram](https://www.instagram.com/commonvoice.yue/)


docs/COMMUNITIES.md
```diff
@@ -21,6 +21,8 @@ Many language communities are self organising and have their contact channels on
 - Bengali (`bn`):
   - [Matrix](https://app.element.io/#/room/#bn-common-voice:mozilla.org)
   - [Discord](https://discord.gg/uacmNh6KQj)
+- Bandari, Achomi and other languages of Hormozan and its neighborhood:
```
moz-bozden, line 24: Could you please also add the ISO codes as they appear in Pontoon?

farooqkz: We haven't decided on one yet :))

moz-bozden: Hey @farooqkz, do these languages not already have ISO codes?

moz-bozden: Achomi is `lrl` - already defined, even it has no counter-part in Pontoon yet...
Bandari (Garmsiri) does not have one, and in many material it is referred to as a dialect. Are you planning to apply for it?

Very interested in others also :)

farooqkz: The CV linguistic expert should tell us if they are variants of each other or what. I'm a Bandari. And it seems Bandari is a language. Achomi hasn't got an ISO code. `lrl` is the ISO code for Lari, another language in the same region.

Out of like 2 dozens of languages in this region, only Lari and Bashgardi have got ISO codes. The rest haven't got one.

We are applying for the entire languages of Hormozgan and neighborhood. However, currently I'm sick and therefore the process has been delayed, unfortunately...

moz-bozden: Take care of yourself @farooqkz...
And thank you for the information.
With this information, I'll merge these changes, for being amended later when things roll...

Whenever you feel well, it would be best to open an issue on all those languages you have been working on so that your team and CV team can work on their correct inclusion?

farooqkz: Thanks

PR #5097, opened 2025-10-02
docker-compose: Expose redis port to localhost

polyfloyd:
This exposes the Redis port of the docker-compose setup. I have this set to run the occasional `flushdb`.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -20,6 +20,8 @@ services:
     restart: unless-stopped
     networks:
       - voice-web
+    ports:
+      - 6379:6379
   storage:
     image: fsouza/fake-gcs-server
     container_name: storage


akrabat:
FWiW, I think that we should be careful about adding additional default ports as it makes it just a little bit harder to run the project with any other Docker containers also running on the same computer.

The easiest way to handle this is to create a `docker-compose-override.yaml` file and put this in it:

```
services:
  redis:
    ports:
      - 6379:6379
```

Another option could be to add a `flush-redis` script to the `package.json`.

moz-bozden:
I think the same, these are the defaults and for a specific implementation one can use a fork and work on that - adding a `docker-compose.override.yaml` file to fulfill specific needs...

polyfloyd:
Adding an override sounds good then, will do that!

PR #5100, opened 2025-10-06
Fix IMPORT_LANGUAGES not working as intended when unset

polyfloyd:
### Type of Pull Request

- [x] Bugfix

According to the comment on `importSentences()`: `If no IMPORT_LANGUAGES specified, all from the file system, else the intersection.`
This condition is implemented by checking the length of the array from splitting IMPORT_LANGUAGES string.
However, this array is produced by `trim().split(',')`. In Javascript, splitting a string which does not contain the separator yields an array with the original string as only element. This causes `[""]` to be used as set of languages, which fails to import anything.

### DIFF
--- a/server/src/lib/model/db/import-sentences.ts
+++ b/server/src/lib/model/db/import-sentences.ts
@@ -145,7 +145,7 @@ async function importLocaleSentences(
   });
 }
 
-export async function importSentences(pool: any, import_languages: string) {
+export async function importSentences(pool: any, import_languages: string[]) {
   // Decide on which locales to import.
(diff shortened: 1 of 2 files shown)


moz-bozden:
Nice find @polyfloyd, thank you. Never thought of that while implementing... And never encountered because it was there all along...

It is actually caused by this line:

https://github.com/common-voice/common-voice/blob/ff278254d632bd7d96c5b0f9b06d67b138112a5a/server/src/config-helper.ts#L122

moz-bozden (approved):
Good solution, thank you again...

polyfloyd:
Thanks!

PR #5115, opened 2025-10-12
1408-Add WebView detection and warning

moz-bozden:
Tested on sandbox server using mobile Facebook in-app browser on two Android phones..

---

Many campaigns are carried out through social media which provide links. Those links open in the in-app-browser of the related app (e.g. Facebook) which does not allow recording due their security measured and/or because they are stripped from such features.

We have been showing an error, but a novice user usually does not understand why it is happening and what to do.

With this PR we try to detect most used in-app-browsers/WebViews and show a more appropriate message:

image


moz-bozden:
Merging this to get to the bottom of ongoing errors.

PR #5129, opened 2025-10-19
1421-Refactor API endpoints and add some validation - Step-1

moz-bozden:
Regroup endpoints and pre-validate incoming requests and objects for existence, which would hopefully catch some of the incoming malformed data caused by proxies and communication errors.

These steps will also ease our move to Public API...

### DIFF
--- a/server/src/lib/api.ts
+++ b/server/src/lib/api.ts
@@ -20,7 +20,7 @@ import * as Basket from './basket'
 import Bucket from './bucket'
 import Clip from './clip'
 import Model from './model'
-import { APIError, ClientParameterError } from './utility'
+import { APIError } from './utility'
 import Email from './email'
 import Challenge from './challenge'
 import Takeout from './takeout'
@@ -95,6 +95,11 @@ export default class API {
     })
 
     router.get('/job/:jobId', validate({ params: jobSchema }), this.getJob)
+
+    //
+    // User & Profile
+    //
+
     router.get('/user_clients', this.getUserClients)
     router.post('/user_clients/:client_id/claim', this.claimUserClient)
     router.get('/user_client', this.getAccount)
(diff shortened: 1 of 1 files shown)


server/src/lib/api.ts
```diff
@@ -654,72 +760,75 @@ export default class API {
...
-        user: { client_id },
-      },
-      params: { jobId },
-    }: Request,
-    response: Response,
-    next: NextFunction
-  ) => {
+  getJob = async (request: Request, response: Response, next: NextFunction) => {
+    const client_id = request?.session?.user?.client_id
+    const job_id = request?.params?.job_id
```
sassela, line 826: is the casing correct & working as expected here? we seem to have switched from `params.jobId` to `params.job_id` in the handler.

server/src/lib/api.ts
```diff
@@ -190,14 +214,108 @@ export default class API {
...
+  }
+
+  sendLanguageRequest = async (
+    request: Request,
+    response: Response,
+    next: NextFunction
+  ) => {
+    const body = request?.body
+    const { email, languageInfo, languageLocale, platforms } = body
+    if (!body || !email || !languageInfo || languageLocale || !platforms) {
```
sassela, line 276: Should this check for a missing `languageLocale`?
```diff
- if (!body || !email || !languageInfo || languageLocale || !platforms) {
+ if (!body || !email || !languageInfo || !languageLocale || !platforms) { 
```

server/src/lib/api.ts
moz-bozden: It should not matter as it is a mere number, but it is always good to be consistent in naming...
Fixed 👍

server/src/lib/api.ts
moz-bozden: Great find 💯 
It seems my KB needs some cleaning, left shift is a bit hard to press :(
Fixed 👍

PR #5130, opened 2025-10-19
1422-Add better logging for lazycache/redis to monitor

moz-bozden:
We are getting more than expected Redis server connection issues, here we extend our monitoring to pinpoint the cause.
Details in internal ticket [1422](https://github.com/common-voice/Internal-Common-Voice/issues/1422).


moz-bozden:
@sassela, I have to merge & release this and check the results immediately - it is only reporting changes.
But please follow it, it might be infra related, which is out of my reach.

PR #5142, opened 2025-10-26
1438-Redo clip selection logic and cache durations

moz-bozden:
The selection process for clips-to-validate was taking too long, and worse, it was very different between locales, thus cache lock duration could not be set efficiently (needed 5 min). It was also very frequent causing DB hit, lock-release issues, which in turn result in errors on the FE.

We started to just adjust cache-lock timing, but finally found ourselves analyzing the bottlenecks in the query itself.
The `RAND()` in the query was causing too many rows to be scanned. So, we did these:
- Remove RAND(), instead get the new clips in reverse creation order, which will get us the newest/fresh items first. With some more optimizations the query time dropped to ~1 sec on production
- We now handle the randomization in the code, AFTER we filter out user-specific business-logic limitations (cannot vote own, cannot vote twice etc).
- We increased the selected clip limit and cache duration and could drop the lock-duration. We set it to 30 secs for now, but after getting some real world statistics, we hope to drop it more.

These changes will prevent cache lock-release errors, users will not wait more (or get timeout and/or 5xx errors), and get recently added clip contributions faster into the database.

### DIFF
--- a/server/src/lib/model/db.ts
+++ b/server/src/lib/model/db.ts
@@ -541,10 +541,13 @@ export default class DB {
   ): Promise<DBClip[]> {
     // get cached clips for given language
     /*
+    In the previous implementation where RAND() was in the SQL:
     Real world measures over 7 days (for 10k records, 1 min cache dur, 3 min lock dur - which was causing trouble):
     - Avg. query time: 2.98 secs - Worst case: 2 min 56 secs (a later measure was 4.3 min)
     - Avg rows scanned: 434,741 - Avg rows returned: 9,243
     => Worst case is caused by large number of records which are not used at all (happens in prominent languages)...
+    Most of the time use of RAND() in SQL result in too many row scanns, thus IO delays - if the LIMIT is large.
+    Thus, in this implementation we moved randomization to code.
     */
     const cachedClips: DBClip[] = await lazyCache(
       `new-clips-per-language-${locale_id}`,
@@ -591,7 +594,7 @@ export default class DB {
     )
 
     // Return random selection from eligible clips
-    if (eligibleClips.size <= count) {
+    if (eligibleClips.size > 0 && eligibleClips.size <= count) {
       // less than requested, so shuffle and return
       return this.shuffleArray(Array.from(eligibleClips))
     } else if (eligibleClips.size > count) {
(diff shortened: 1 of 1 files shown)


server/src/lib/model/db.ts
```diff
@@ -198,6 +198,33 @@ export default class DB {
     }
   }
 
+  /**
+   * Get random selection from array
+   */
+  private getRandomSelection<T>(array: T[], count: number): T[] {
+    const shuffled = [...array]
+
+    // Fisher-Yates shuffle
```
sassela, line 207: TIL!

moz-bozden: [The Art of Computer Programming](https://en.wikipedia.org/wiki/The_Art_of_Computer_Programming) by [Knuth](https://en.wikipedia.org/wiki/Donald_E._Knuth) was our text book back in the days...

I have this in all my toolsets, if the array is not huge it perfoms... RAND() is also not portable (MySQL only).
We will probably move this to a more generic location to be reused in many other hogging SQL with RAND()...

server/src/lib/model/db.ts
```diff
@@ -560,18 +581,25 @@ export default class DB {
...
       })
     )
 
-    if (validClips.size > count) {
-      return Array.from(validClips)
+    // Return random selection from eligible clips
+    if (eligibleClips.size <= count) {
+      // less than requested, so shuffle and return
+      return this.shuffleArray(Array.from(eligibleClips))
+    } else if (eligibleClips.size > count) {
```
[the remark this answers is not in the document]
moz-bozden: Well, it was not that, but it made me find an error... It had to be:
```ts
    // Return random selection from eligible clips
    if (eligibleClips.size > 0 && eligibleClips.size <= count) {
      // less than requested, so shuffle and return
      return this.shuffleArray(Array.from(eligibleClips))
    } else if (eligibleClips.size > count) {
      // more than requested, so return requested count from randomized array
      return this.getRandomSelection(Array.from(eligibleClips), count)
    }
```
I was not reaching the zero case :/

moz-bozden:
Thank you!
I'm merging this. Unfortunately, the real effect can only be seen on live database. I hope with the latest 3 PRs (inc. this) our error-out rate will drop considerably...

PR #5158, opened 2025-11-06
1456-Add migration to fix predefined French South accent cannot be added

moz-bozden:
It had a colliding "name" field with a user defined one (only in production)...
This resulted in a rather complex transaction.
Tested OK on local development environment with simulated data...

### DIFF
--- a/server/src/lib/model/db/migrations/20251106160000-fix-fr-south-accent.ts
+++ b/server/src/lib/model/db/migrations/20251106160000-fix-fr-south-accent.ts
@@ -1,6 +1,22 @@
 // This is only production specific migration
 // Fixes missing French South accent
 // It was caused because there is a user-defined accent with the same name
+//
+// The accent table has UNIQUE KEY `accent_key` (`locale_id`,`accent_name`)
+// In 2024, somebody added a user-defined accent "Français du sud de la France"
+// which blocked the creation of the predefined accent "fr-metro-south" with the same name
+// during our 20251010200000-migrate-accents-variants-fr.ts migration
+//
+// We cannot directly delete the old user-defined one, because there may be users assigned to it (there are actually two of them in production)
+//
+// So the steps are:
+// - Rename the old user-defined accent to "[DEPRECATED] Français du sud de la France"
+// - Create the predefined accent "fr-metro-south"
+// - Re-assign all users from the deprecated accent to the new predefined accent
+// - Delete the deprecated accent
+//
+// We had to wrap everything in a transaction for safety
+//
 const LOCALE_CODE = 'fr'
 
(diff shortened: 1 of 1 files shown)


server/src/lib/model/db/migrations/20251106160000-fix-fr-south-accent.ts
```diff
@@ -0,0 +1,113 @@
...
+      return null
+    }
+
+    const oldAccentRows = await db.runSql(
+      `SELECT id FROM accents WHERE locale_id = ? AND accent_token = ?`,
+      [localeId, OLD_USER_ACCENT_CODE]
+    )
+
+    if (!oldAccentRows || oldAccentRows.length === 0) {
+      console.warn(
```
akrabat, line 64: Is this an acceptable situation? i.e. "we're all good here"?

If so, making this is `info`, not `warn`. On the flip side, if it's "bad", should we use `error`?

server/src/lib/model/db/migrations/20251106160000-fix-fr-south-accent.ts
```diff
@@ -0,0 +1,113 @@
...
+    // INSERT the new accent - use regular INSERT (not IGNORE) since we checked existence
+    const insertResult = await db.runSql(
+      `INSERT INTO accents(locale_id, accent_name, accent_token) VALUES (?,?,?)`,
+      [localeId, ACCENT_NAME, NEW_ACCENT_CODE]
+    )
+
+    const newAccentId = insertResult.insertId
+
+    if (!newAccentId || newAccentId <= 0) {
+      console.warn(`New accent not created, insertId: [${newAccentId}]`)
```
akrabat, line 94: This is `error` situation? Feels "bad" if an insert fails.

server/src/lib/model/db/migrations/20251106160000-fix-fr-south-accent.ts
moz-bozden: For this specific one, we know the data, they should be there - but on production ONLY - so ERROR.
When you run it on local or in sandbox/staging, it should just give some reason and skip the transaction - so INFO.
As it is not connected to other facilities like Sentry, warn would be enough I think...
I test these though UI, if problem, check DB, still problem check the code/log on local - this is where they appear for me...

server/src/lib/model/db/migrations/20251106160000-fix-fr-south-accent.ts
moz-bozden: Again, here it depends. If for some reason it is already there, we should skip it, not throw => no error I think...

moz-bozden:
Added comments to the top, but warn is good for me because they are data/environment dependent.

PR #5183, opened 2025-11-24
Multipli fixes after  branch 1484 1450 tests

moz-bozden:
Fixes several issues after testing in:
- 1450 - https://github.com/common-voice/common-voice/pull/5175 (one endpoint was designed for both protected and unprotected request handling)
- 1484 - https://github.com/common-voice/common-voice/pull/5180 (too much cold-start logs)
- Optimization in statistics calculation
- In Review and Listen pages, there were no guard for queued sentences, thus each time an action is taken, a request was being sent to BE to refill.


moz-bozden:
Completed in https://github.com/common-voice/common-voice/pull/5184

PR #5252, opened 2026-01-15
fix: align skip transition with card animation

### LINKED ISSUE #1042: 'Skip' transition needs alignment with general card transition animation
As spoken about today, preferred transition for sentence skip is to align with full sentence card transition animation (rather than just clip text refreshing).

VedantMadane:
## Summary

This PR fixes the skip transition to properly animate the sentence card, matching the general card transition animation.

## Problem

When clicking the 'Skip' button, the sentence card's text would abruptly change without any transition animation. This was inconsistent with other card transitions (like recording completion) which animate smoothly.

## Root Cause

The issue was in the \key\ prop of the card div:

\\\	sx
key={sentence ? sentence.text : i}
\\\

When a sentence is skipped:
1. The current sentence is removed and replaced with a new one
2. The \key\ changes from the old sentence's text to the new sentence's text
3. React unmounts the old card element and mounts a new one
4. This bypasses the CSS transition defined on the \.card\ class

## Solution

Changed the key to use the array index:

\\\	sx
key={i}
\\\

This preserves the DOM element when the sentence content changes, allowing the CSS transitions (\	ransform\, \opacity\) to animate smoothly.

## Testing

1. Navigate to the Speak or Listen contribution page
2. Click the Skip button
3. Observe that the card now animates when transitioning to the next sentence

## Fixes

Fixes #1042

### DIFF
--- a/web/src/components/pages/contribution/contribution.tsx
+++ b/web/src/components/pages/contribution/contribution.tsx
@@ -413,7 +413,7 @@ class ContributionPage extends React.Component<ContributionPageProps, State> {
                       // don't let Chrome auto-translate
                       // https://html.spec.whatwg.org/multipage/dom.html#the-translate-attribute
                       translate="no"
-                      key={sentence?.id ?? `empty-${i}`}
+                      key={i}
                       className={
                         'card card-dimensions ' + (isActive ? '' : 'inactive')
                       }


moz-bozden:
Hi @VedantMadane, thank you for finding the issue and PR.

On the other hand, use of index as key is an anti-pattern and causes many more problems.
In the last few months I fixed a line of these by updating such keys, even sometimes had to generate UUID because there was no unique identifier.

An explanation is here: https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/
But you can find many more in search...

VedantMadane:
Would you be open to exploring an alternative approach that preserves the animation fix while avoiding the index-as-key anti-pattern?

moz-bozden:
Of course :)
We get very happy when we see PRs.
I also was not aware of the animation issue, thank you for finding it.

moz-bozden:
I think Copilot and you are right on the index solution to get the animation working, although anti-pattern.
I looked at the code again and, with the currently implemented state management, and mutations that remove items, the only way is use a simple index - if we want to get the animations, which I think is not breaking issue.

I need to more deeply analyze and test it to be sure we are not breaking any functionality...
I have a feeling that it is not fixed before because it does break.

PR #5258, opened 2026-01-19
1514-Add malay (ms) standard and other dialects as variants

moz-bozden:
This PR was necessary to move the Spontaneous Speech dataset ms-MY (Bahasa Malay) under Malay language as a variant.

Malay has many dialects, which also have their own ISO 639-3 language codes ([see](https://en.wikipedia.org/wiki/Malay_language)).
Those ISO 639-3 language codes should be preferred and opened in Pontoon as separate language, forming a separate dataset.

We did not define any predefined-accents, PRs are welcome.


moz-bozden:
Merging this for release.

PR #5294, opened 2026-02-14
1537-Add new SPS report keys and strings

moz-bozden:
**IMPORTANT: This should be merged/released with: https://github.com/common-voice/spontaneous-speech/pull/516**

This changes a report item at the presentation level by mapping the Pontoon keys.
Generalizes SPS report item to cover more common cases.

On SS/CS we had this case:
```
different_language = Different language
different_language-detail = The content contains a language different than I was expecting.
```

But users get many other problems, e.g. very short recordings caused by double clicks, audio without speech, corrupt audio, etc. in addition to "different language" case.
These can not be transcribed, and there is only "other" option to report these.
But the "other" option will not take them out of circulation, they will continue to appear to other users.

Thus we will change it to:
```
empty_or_unintelligible = Empty or Unintelligible
empty_or_unintelligible-detail = The content is empty, corrupt, contains a different language, etc.
```

image

We will not change the DB for now...


web/locales/spontaneous-speech/en/modals/report.ftl
```diff
@@ -6,6 +6,8 @@ offensive_speech = Offensive Content
 offensive_speech-detail = The content has disrespectful or offensive language.
 different_language = Different language
 different_language-detail = The content contains a language different than I was expecting.
+not_good_for_transcription = Not possible to transcribe
+not_good_for_transcription-detail = The content is empty, contains a different language, or not fit for transcription in any way.
```
ftyers: I don't like the fact that this contains 'different language' and the previous one does too. My proposal:

> "The content is empty or unintelligible"

web/locales/spontaneous-speech/en/modals/report.ftl
```diff
@@ -6,6 +6,8 @@ offensive_speech = Offensive Content
 offensive_speech-detail = The content has disrespectful or offensive language.
 different_language = Different language
 different_language-detail = The content contains a language different than I was expecting.
+not_good_for_transcription = Not possible to transcribe
```
ftyers: Call it "Empty or unintelligible"

web/locales/spontaneous-speech/en/modals/report.ftl
moz-bozden: The different_language keys were planned to retire.
I kept them here first not to break SPS report modal.
Now that that part is also implemented, I could remove them.
I changed it as you requested, and added an extra string for audio corruption etc.
Updating the initial topic.

PR #5323, opened 2026-03-08
Optimize flow in SCS bundler

moz-bozden:
- Pre-check and filter out existing release files so that we get correct measures
- Add `--force` CLI flag to re-execute all or (more like) selected locales (e.g. if the dataset is corrupt somehow)
- Rework misbehaving BullMQ locking mechanism, add a timer based lock-duration extension


moz-bozden:
Tested on local with seed data, merging...

PR #5374, opened 2026-04-10
1606-Make communities md reachable from front-end

moz-bozden:
Adds GitHub link invites to:
- Languages page
- About page - playbook / how to grow section

Handles: https://github.com/common-voice/common-voice/issues/5280


ftyers:
Lgtm!

PR #5396, opened 2026-05-11
Add Chichewa variants

moz-acfort:
Adding Chichewa variants per https://github.com/common-voice/common-voice/issues/5391


moz-bozden (approved):
Thank you :)

PR #5397, opened 2026-05-11
Add Sundanese Variants

moz-acfort:
Adding based on issue: https://github.com/common-voice/common-voice/issues/5250


moz-bozden (approved):
💯

PR #5414, opened 2026-06-01
1699-Add specific error handling guards at multiple locations and drop Sentry noise

moz-bozden:
This PR tightens client-side error handling and reduces Sentry noise. It preserves typed `SystemError`/`ClientHandledError` instances through the API layer, extends the global unhandled-rejection handler to cover client-handled errors (with localized 429 messaging including `Retry-After`), suppresses background refill rejections, and adds defensive guards/catches in dashboard cards, the listen page, and the contribution shortcut handler. It also configures `Sentry.init` with `ignoreErrors` and a `beforeSend` that rewrites `<unknown>` rejection values.

- API/error plumbing: don't rewrap `SystemError`/`ClientHandledError` as network errors; handle both in `useSystemErrorHandler` with `retryAfter` localization and add `error-title-429` Fluent message.
- UI hardening: add defensive `.catch`/guards in `progress-card`, `leaderboard-card` pagination, `listen` play/audio key, `contribution` keyboard shortcut, and sentence refill background calls; tolerate null vote response in `clips` store.
- Sentry config: add `ignoreErrors` patterns and a `beforeSend` that replaces `<unknown>` exception values with stringified reason.

### DIFF
--- a/web/locales/common-voice/en/pages/error.ftl
+++ b/web/locales/common-voice/en/pages/error.ftl
@@ -10,6 +10,14 @@ error-clip-upload-too-large = Your recording file is too large to upload. Please
 error-clip-upload-server-error = Server error processing your clip. Please reload the page or try again later.
 error-title-404 = We couldn’t find that page for you
 error-content-404 = Maybe our <homepageLink>homepage</homepageLink> will help? To ask a question, please join the <matrixLink>Matrix community chat</matrixLink>, monitor site issues via <githubLink>GitHub</githubLink> or visit <discourseLink>our Discourse forums</discourseLink>.
+error-title-429 =
+  { $retryAfter ->
+    [0] You’re going too fast. Please slow down and try again in a moment.
+   *[other] You’re going too fast. Please try again in { NUMBER($retryAfter) } { $retryAfter ->
+      [one] second
+     *[other] seconds
+    }.
+  }
 error-title-500 = Sorry, something went wrong
 error-content-500 = An unexpected error occurred. Please try again later. For help, please join the <matrixLink>Matrix community chat</matrixLink>, monitor site issues via <githubLink>GitHub</githubLink> or visit <discourseLink>our Discourse forums</discourseLink>.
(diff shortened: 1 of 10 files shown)


web/locales/common-voice/en/pages/error.ftl
```diff
@@ -10,6 +10,14 @@ error-clip-upload-too-large = Your recording file is too large to upload. Please
 error-clip-upload-server-error = Server error processing your clip. Please reload the page or try again later.
 error-title-404 = We couldn’t find that page for you
 error-content-404 = Maybe our <homepageLink>homepage</homepageLink> will help? To ask a question, please join the <matrixLink>Matrix community chat</matrixLink>, monitor site issues via <githubLink>GitHub</githubLink> or visit <discourseLink>our Discourse forums</discourseLink>.
+error-title-429 =
```
flodolo, line 13: Looks like GitHub lost my comment :-( 

This is broken in Pontoon, e.g. https://pontoon.mozilla.org/fy-NL/common-voice/all-resources/?string=334714

The 0 case is not guaranteed to exist across locales, so you probably want a separate string.

In any case, this is how the string should look like (no variant logic in the middle of a string).

```
error-title-429 =
  { $retryAfter ->
    [0] You’re going too fast. Please slow down and try again in a moment.
    [one] You’re going too fast. Please try again in { $retryAfter } second.
   *[other] You’re going too fast. Please try again in { $retryAfter } seconds.
  }
```

Note that you will need a new ID to fix it.

Correct approach


```
error-title-429-no-time = You’re going too fast. Please slow down and try again in a moment.
error-title-429-with-time =
  { $retryAfter ->
    [one] You’re going too fast. Please try again in { $retryAfter } second.
   *[other] You’re going too fast. Please try again in { $retryAfter } seconds.
  }
```

moz-bozden: Thank you for the feedback... I didn't know that, fixing.

PR #5445, opened 2026-07-06
Add Kabyle variants

moz-acfort:
Adding Kabyle variants as proposed in https://github.com/common-voice/common-voice/issues/5436


moz-bozden:
LGTM, I'll test it tonight...

moz-bozden (approved):

PR #5451, opened 2026-07-18
Add Hausa variants to variants.ts

moz-acfort:
Adding Hausa variants per this issue https://github.com/common-voice/common-voice/issues/5450


moz-bozden (approved):
lgtm

PR #5473, opened 2026-08-24
Add Levantine Arabic variants

moz-acfort:
Adding Levantine variants based on this GitHub issue https://github.com/common-voice/common-voice/issues/5472


moz-bozden (approved):
Tested on local OK, thank you!

PR #5517, opened 2026-09-18
Adding Brazilian Portuguese tag here to support adding it to SPS

moz-acfort:
Adding pt-BR to common.ftl and language.py files to support having Brazilian Portuguese on Common Voice Spontaneous Speech, following issue https://github.com/common-voice/common-voice/issues/5515 and pull request https://github.com/common-voice/spontaneous-speech/pull/559 in SPS repo


moz-bozden:
This PR adds Pontoon and bundler necessities because `pt-BR` had been added as the third variant-based dataset in SPS. These will be specially handled during monorepo migration where we will merge the two systems and give the SPS datasets varint feature.
@moz-acfort, I added some nits for completeness...

